PR #30, opened 2020-05-20
Use custom compiled SQLite binary

evansd:
Due to its use in tiny embedded systems, SQLite has quite a low default
parameter limit (999). Most OS distributions bump this signficantly
(e.g. Ubuntu to 250,000 and OS X to 500,000) but the pysqlite3-binary
package leaves it at the default. Until we get a change merged upstream
we use a custom-compiled package.


evansd:
Closed in favour of #32 because Charles Leifer is very fast.

PR #31, opened 2020-05-20
Add Sentry for alerting and debugging support

evansd:
The appropriate DSN has already been added to the production config so
this should start working as soon as it's deployed.

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -14,6 +14,9 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 import os
 import sys
 
+import sentry_sdk
+from sentry_sdk.integrations.django import DjangoIntegration
+
 # Patch sqlite3 to ensure recent version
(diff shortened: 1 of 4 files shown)


opencodelists/settings.py
```diff
@@ -175,6 +179,15 @@
 }
 # fmt: on
 
+SENTRY_DSN = os.environ.get("SENTRY_DSN")
+if SENTRY_DSN:
+    sentry_sdk.init(
+        dsn=SENTRY_DSN,
+        integrations=[DjangoIntegration()],
+        # If you wish to associate users to errors (assuming you are using
+        # django.contrib.auth) you may enable sending PII data.
```
inglesp, line 188: I'm not quite sure who this comment's for.

evansd: Ah sorry, the whole thing was copied verbatim from Sentry's docs.

PR #56, opened 2020-07-31
Implement create_[super]user methods

ghickman:
This adds a custom Manger to our `User` model, implementing `create_user` and `create_superuser` methods so we can use Django's `createsuperuser` command to create a user when working locally.


opencodelists/models.py
```diff
@@ -3,6 +3,43 @@
 from django.utils.text import slugify
 
 
+class UserManager(BaseUserManager):
+    """
+    Custom user model manager where email is the unique identifiers
```
inglesp, line 8: Typo?  s/identifiers/identifier/

opencodelists/models.py
```diff
@@ -3,6 +3,43 @@
...
+
+    def create_user(self, username, email, password, **extra_fields):
+        """Create and save a User with the given email and password."""
+        if not email:
+            raise ValueError("The Email must be set")
+
+        org = Organisation.objects.first()
+        if not org:
+            org = Organisation.objects.create(
+                name="Data Lab", slug="datalab", url="https://ebmdatalab.net/"
```
inglesp, line 20: Nit (but I think it should be fixed) -- it's "DataLab" not "Data Lab".

opencodelists/models.py
```diff
@@ -3,6 +3,43 @@
...
+    Custom user model manager where email is the unique identifiers
+    for authentication instead of usernames.
+    """
+
+    def create_user(self, username, email, password, **extra_fields):
+        """Create and save a User with the given email and password."""
+        if not email:
+            raise ValueError("The Email must be set")
+
+        org = Organisation.objects.first()
```
inglesp, line 17: Maybe for later: could the organisation slug be passed in as a required argument?

opencodelists/models.py
ghickman: That'll teach me to copy and paste!

ghickman: Fixed and rebased

opencodelists/models.py
ghickman: Fixed and rebased

opencodelists/models.py
ghickman: I had a quick look for this and couldn't find an easy way to make it happen with `createsuperuser`.

Adding a custom command might be quicker to get that.

PR #59, opened 2020-08-03
Tooling Changes

ghickman:
This makes a variety of changes to the tooling with the primary goals of explicitness, maintainability, and a bit of documentation [if you squint!].

The Makefile wraps up common commands such that someone wanting to run them doesn't need to know how to run them (do I put `--check` on `black` or `isort`?) or the specifics of any tooling choice (is this project using `flake8` or `pylint`?).  The Makefile happens to act as a very basic form of documentation in that it shows how we expect a given tool to run (but it isn't expected to replace real docs!).

It also has the benefit of allowing us to be less prescriptive on how those tools are run by an individual by "configuring" a tool via CLI args.  For instance maybe I'd like to run pytest in a very specific way while I fix a single test, instead of having to edit/override pytest's config I can construct the CLI args as I see fit for that particular fix.

It's out of scope for this PR but I'm also planning to call the Makefile commands when running tooling in CI.

Separately, I've also combined the various tooling configs into `pyproject.toml` since they have all recently added support for this format (except flake8 which looks to be a long way off), and it seems nice to have things all in one place!

---

The potentially controversial change here is moving to explicitly configured targets for the `black`, `flake8`, and `isort`.  I've been bitten many times in the past by 1 or more of those tools traversing directories I didn't expect.  I set up my virtualenvs in the project repo (with either direnv or poetry), which is the more obvious failure mode but I've worked on projects with large directories I didn't want to be checked (isort once tried to run on 50Gb of .txt files for some reason!).  I consider this kind of change similar to defensively programming in that it's really planning for the worse case scenario.

There are still downsides to this pattern (adding a new Django app and not updating the Makefile, for instance) so it's not a clear win in that respect.  However, having been burnt by the alternative a few times in the past I thought I'd at least float the change 🙂


ghickman:
Between

> I'm not totally convinced by the need to be explicit about the paths covered

and

> I think I'm -1 on the bit you flagged as controversial

I think non-explicit paths is a winner then!  I'll remove that commit and add a few extra known ignores to the appropriate lists.

---

> I'd also be very keen to have a single command

@evansd – I was planning on introducing [pre-commit](https://pre-commit.com/) for this later.  I dislike that we have to manually keep the tool versions in sync between it and the repo but I find tooling versioning doesn't change all that often so it hasn't been a hardship for me in the past.  But it covers the one command you'd like (you can run `pre-commit` or do `pre-commit install` to have it set up as a pre-commit hook in git) and it comes with a huge library of hooks for various tooling.  Unrelated but a really useful one is the isort seed tool which looks at code being committed and makes sure third party libs are explicitly listed in the isort config.  I've found that adds an extra level of stability to isort in the past.

However, in the interim, `make fix` will "just fix everything" by design.  I even format and import-sort my migrations files normally because `make fix` sorts them out trivially for me 🙂

evansd:
Sounds great :+1:

ghickman:
Removed the explict paths commit and told flake8 to ignore direnv and local venv directories.  I think this is good to go now?

PR #60, opened 2020-08-05
Add Sign-Offs and References

ghickman:
This adds the ability to add Sign-Offs and References to Codelists when creating them without having to use the Admin.

It uses two formsets to deal with the multiple Signoff/Reference models related to a Codelist and uses a small amount of Javascript to handle adding more forms.

I've switched the create view over to a Class-based View mainly to separate out the code.  It's based on TemplateView because there's no GCBV for dealing with a form and two formsets but I've mirrored how the various methods are used in [CreateView](http://ccbv.co.uk/projects/Django/3.0/django.views.generic.edit/CreateView/) to maintain consistency with the community (eg conditional creation of Form[Set] instances in `get_context_data`).  I also wanted to showcase a strength of CBVs - in more complicated views a class lets you split things up into methods which I find can aid readability.

As part of testing this I needed to have unique email and usernames for each of the Users created so I've added Factory Boy (in a separate commit for ease of review) and combined it with our actions.

---

Testing Discussion!

## Building Data Structures
When we write tests we often need some data structures to run the code we're testing against, factories are great for this and we have some of those already.  However, I like to partition those into two things: factories (building blocks) and fixtures ("things you care about most of the time").

Factories are building blocks because they're a function/class/$thing which produces one or more working model instances.  They're already useful at that point because you often need a single model in a test, eg a `User` for authenticating with a View.

Fixtures are where things start to shine though.  Those larger data structures I talked about in the stand up often need multiple models in various states (eg a fire engine with 3 hoses but one has a leak or something equally appropriate for a codelist 😉).  I like to define fixtures as "things I would like in my tests that I don't want to care [too much] about the internals of".

As a concrete example - we need to pass in a CSV file to create Codelists and create/modify Versions.  In our view tests we have a function to turn a string into a BytesIO object.  The contents of this "file" is nearly identical in every test.  This would be perfect as a fixture, where I could ask for `dummy_csv` in my test and not have to care what's in the CSV, except that it's valid!  The code to create that CSV still exists in conftests as a fixture and we only have to care about it when the CSV isn't correct for a tests we're writing.

## RequestFactorys
I only got introduced to these a few months ago, but I love them.  They let you bypass all of Django's URL routing code and get you closer to the view code you're trying to test.  Relatedly if you like CBVs the `.setup()` method used in conjunction with a RequestFactory means you can test specific methods of a view which I really like for larger views.

## Tests Layout
I tend to put tests in their own directory called `tests` at the top level of the repo.  Testing related files go in there, ie  `factories.py` and `conftests.py`.  Then I mirror the project structure inside that directory (eg `tests/codelists/test_views.py`).  And finally I run them with `pytest [options] tests`.  The biggest pros I've seen for this are:
* slight optimisation of test collection (few files to look at but a bit dubious)
* easier to ignore when building a deployment asset, eg a Docker image

Personally I like the grouping of tests from an organisation standpoint.

Thoughts and options welcome!

### DIFF
--- a/codelists/forms.py
+++ b/codelists/forms.py
@@ -52,7 +52,6 @@ class CodelistForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
         self.helper = FormHelper()
         self.helper.form_tag = False
-        self.helper.disable_csrf = True
         super().__init__(*args, **kwargs)
 
 
--- a/codelists/forms.py
+++ b/codelists/forms.py
@@ -2,7 +2,44 @@ from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Submit
 from django import forms
 
-from .models import Codelist, CodelistVersion
+from .models import Codelist, CodelistVersion, Reference, SignOff
+
+
+class FormSetHelper(FormHelper):
+    """
+    FormHelper for use with Crispy Forms and FormSets
+
+    When using Crispy Forms helpers with FormSet their layout is applied to the
+    Forms but attributes to the FormSet, so we need to pass the FormHelper
+    instance into the crispy_forms templatetag in the view.
+
+    https://django-crispy-forms.readthedocs.io/en/latest/crispy_tag_formsets.html#formsets
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.disable_csrf = True
+        self.form_tag = False
+
+
+class ReferenceForm(forms.ModelForm):
+    class Meta:
+        model = Reference
+        fields = [
+            "text",
+            "url",
+        ]
+
+
+class SignOffForm(forms.ModelForm):
+    date = forms.DateField(widget=forms.TextInput(attrs={"type": "date"}))
+
+    class Meta:
+        model = SignOff
+        fields = [
+            "user",
+            "date",
+        ]
 
 
 class CodelistForm(forms.ModelForm):
@@ -14,7 +51,8 @@ class CodelistForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         self.helper = FormHelper()
-        self.helper.add_input(Submit("submit", "Submit"))
+        self.helper.form_tag = False
+        self.helper.disable_csrf = True
         super().__init__(*args, **kwargs)
 
 
--- a/codelists/tests/test_views.py
+++ b/codelists/tests/test_views.py
@@ -1,9 +1,11 @@
 import csv
 from io import BytesIO, StringIO
+from unittest.mock import patch
 
 import pytest
 from pytest_django.asserts import assertContains, assertRedirects
 
+from codelists.views import CreateCodelist
 from opencodelists.tests.factories import ProjectFactory, UserFactory
 
 from . import factories
@@ -18,8 +20,12 @@ def logged_in_client(client, django_user_model):
     return client
 
 
-def test_create_codelist(logged_in_client):
-    p = ProjectFactory()
+def test_createcodelist_success(rf):
+    project = ProjectFactory()
+    signoff_user = UserFactory()
+
+    assert project.codelists.count() == 0
+
     csv_data = "code,description\n1067731000000107,Injury whilst swimming (disorder)"
     data = {
         "name": "Test Codelist",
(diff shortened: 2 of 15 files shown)


codelists/tests/test_views.py
```diff
@@ -18,18 +20,97 @@ def logged_in_client(client, django_user_model):
...
+    with patch("codelists.views.CreateCodelist.all_valid") as mock_all_valid, patch(
+        "codelists.views.CreateCodelist.some_invalid"
+    ) as mock_some_invalid:
+        view.post(request)
+
+    # was the error handler, some_invalid, called?
+    # we're not doing any custom error handling in the form or formsets so no
+    # need to test what Django is already testing
+    mock_all_valid.assert_not_called()
+    mock_some_invalid.assert_called_once()
```
inglesp, line 113: I'm not keen on this style of testing -- it couples the tests to the implementation (making implementation changes harder) and doesn't test what the user experiences (so that it's less useful as a specification of behaviour).  If the form is invalid, does the user get a message?  If so, I think that'd be a better thing to test.

codelists/views.py
```diff
@@ -26,30 +38,87 @@ def index(request):
...
-        "form": form,
-    }
+        # get changed forms so we ignore empty form instances
+        reference_forms = (f for f in reference_formset if f.has_changed())
+        signoff_forms = (f for f in signoff_formset if f.has_changed())
+
+        for form in itertools.chain(reference_forms, signoff_forms):
+            instance = form.save(commit=False)
+            instance.codelist = codelist
+            instance.save()
```
inglesp, line 65: Would it be possible to add `references` and `signoffs` as arguments to  `actions.create_codelist()`?

We've now got transaction logic in two places, and if we wanted a factory to create a codelist with references, we'd have to duplicate what we have here.

codelists/forms.py
```diff
@@ -14,7 +51,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         self.helper = FormHelper()
-        self.helper.add_input(Submit("submit", "Submit"))
+        self.helper.form_tag = False
+        self.helper.disable_csrf = True
```
inglesp, line 55: Why do we need to do this?

codelists/tests/test_views.py
```diff
@@ -14,27 +16,105 @@
...
+    response = CreateCodelist.as_view()(request, project_slug=project.slug)
+
+    assert response.status_code == 302
+    assert response.url == f"/codelist/{project.slug}/test-codelist/"
+
+    assert project.codelists.count() == 1
+    codelist = project.codelists.first()
+    assert codelist.name == "Test Codelist"
+
+    # we should have one reference to example.com
```
inglesp, line 61: +1 for comments in tests, I should do more of this

codelists/views.py
```diff
@@ -26,30 +38,87 @@ def index(request):
...
-                coding_system_id=form.cleaned_data["coding_system_id"],
-                description=form.cleaned_data["description"],
-                methodology=form.cleaned_data["methodology"],
-                csv_data=form.cleaned_data["csv_data"].read().decode("utf8"),
-            )
-            return redirect(codelist)
-    else:
-        form = CodelistForm()
+@method_decorator(login_required, name="dispatch")
+class CreateCodelist(TemplateView):
```
inglesp, line 42: I'm not used to CBVs, so I don't find this particularly easy to read.  One change that might help would be to reorder the methods like this:

* dispatch
* post
* all_valid/some_valid
* get_context_data
* get_form, get_X_formset

You can then more-or-less read the code top to bottom.

codelists/forms.py
ghickman: We want to keep the benefit of the Crispy `FormHelper` instance but also to end up with an HTML form comprised of 1 Form and 2 FormSet instances.

* `form_tag = False` removes the wrapping `<form>` element which is step 1.
* `disable_csrf = True` removes the `{% csrf_token %}` and I've just realised that's a mistake on my part from trying to add formsets initially!  Fixed this in e1a9fe1

codelists/views.py
ghickman: Yep, that makes sense, fixed in 0e203a1

codelists/tests/test_views.py
ghickman: That's a fair point, I think I'll remove the mocking and not check that outcome other than the response code.  I like your point about "less useful as a specification of behaviour"!  Since Django (and to some degree Crispy) handle rendering error messages for us I don't think we need to check that?

ghickman:
> The ordering of methods in the CBV could be changed to make reading it easier.

> I would reorder the methods on CreateCodeList to be in something like execution order

But my alphabetisation! 😜

I'm happy to re-order them to aid learning here.  Generally though I really would espouse alphabetisation since some methods can be called from different places (eg `get_context_data` is also called inside [TemplateView.get](http://ccbv.co.uk/projects/Django/3.0/django.views.generic.base/TemplateView/#get)).

---
> terminological distinction between factories and fixtures

I should have been clearer, this is _my_ distinction, but it's born from seeing how well they can layer together in practice!

---
>  have seen factory setup which is totally unintelligible

Same!  And same for pytest fixtures too, I think they're both incredibly powerful tools which means they can get out of hand without supervision.  My rule of thumb for factories are:
* there should be one factory for a model
* calling `SomeFactory()` should provide me with a basic, working model instance (eg unique fields need a Sequence)

There are a variety of ways to call factories with configuration at runtime, eg `UserFactory(email="george@example.com")` to get the desired state in a fixture/test.

---
> -0 on RequestFactory. I'm not quite sure what it offers

More targetted tests.  Using Client (which is built on RequestFactory!) sends requests through the entire Django request/response stack, so that's at minimum URL routing, middleware, etc.  RequestFactory is a way to isolate a view for testing.  An added bonus is they decouple your tests from your URLs - having centralised URLs and `reverse[_lazy]()` means not having to edit URLs in multiple places, using hardcoded URLs couples those tests to a particular URL implementation.

I'm by no means saying we should rewrite all our tests to use them 🙂  There are definite downsides to them (eg the lack of middleware means you have to construct the ones you need manually if your view relies on them) so it's a balance.  However I do think they're clearer as a default option (but I'm unlikely to push to have a Client-based test rewritten in a review!).

---
> having a separate directory means that you have to use absolute imports

Yep, I've had this layout catch odd import bugs before.  I find it akin to the [folks](https://hynek.me/articles/testing-packaging/) who say libraries should live inside a src dir.  I like the separation of "that is the app" and "this tests the app" too.

codelists/views.py
ghickman: Done in 2aff811

evansd:
> Generally though I really would espouse alphabetisation since some methods can be called from different places

Interesting. I don't think I've ever encountered this as an approach before. Would you order functions in a module similarly as well?

ghickman:
> Would you order functions in a module similarly

As a default, yes.  I don't treat it as a hard and fast rule by any means, but rather a starting point based on previous experiences.

Consider that all software grows and mutates over time.  A good ordering for functions|methods|classes now will change over time as the software evolves, maybe a new method is needed in a CBV and that arrives part way in the flow, and then it's position in the flow is changed later on.  Moving the position of the method adds churn to the diffs for reviewers/history spelunking.

Personally this was born out of working with lots of other folks' code at various jobs/clients over the years and coming to the conclusion that most stable default ordering was alphabetical.  In existing code I try to copy whatever ordering exists (or add to the bottom if there isn't an obvious one!) though 🙂

sebbacon:
FWIW I don't have opinions about ordering.  On the navigation argument I use my text editor's search/jump features to navigate a file; on the diff argument, reordering changes are infrequent and should usually be easy enough to follow.

In summary, +0 on continuing to be unopinionated :)

codelists/tests/test_views.py
inglesp: > Since Django (and to some degree Crispy) handle rendering error messages for us I don't think we need to check that?

No need to check that Django's doing the right thing, but it's often worth checking that we're using Django correctly.  In this case, checking for the error message is a good way of validating the behaviour of the view.

Let's not hold this PR up any longer though.  I may do a drive-by refactor of the test in the future, or you can if you like.

codelists/tests/test_views.py
```diff
@@ -18,18 +20,97 @@ def logged_in_client(client, django_user_model):
...
+        "reference-INITIAL_FORMS": "0",
+        "reference-MIN_NUM_FORMS": "0",
+        "reference-MAX_NUM_FORMS": "1000",
+        "reference-0-text": "foo",
+        "reference-0-url": "http://example.com",
+        "signoff-TOTAL_FORMS": "1",
+        "signoff-INITIAL_FORMS": "0",
+        "signoff-MIN_NUM_FORMS": "0",
+        "signoff-MAX_NUM_FORMS": "1000",
+        "signoff-0-user": signoff_user.username,
```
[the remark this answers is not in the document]
ghickman: I've just tested this and both the missing key and an empty value produce the same error in the formset.

codelists/tests/test_views.py
ghickman: Fixed in 8a48704

Note that I tested the vieww context contained the expected formset instance with errors rather than looking at the HTML output since Crisipy should handle that.

Issue #49, opened 2020-07-26
Deploy automatically when main is updated


inglesp:
This can be done via [ebmbot](https://github.com/ebmdatalab/ebmbot).

iaindillingham:
I'm going to close this issue, as embot is deploying automatically.

inglesp:
> I'm going to close this issue, as embot is deploying automatically.

Afraid it's not, sorry.

iaindillingham:
😢

rebkwok:
Closing as github actions now deploys automatically

PR #69, opened 2020-08-05
Configure CI

### LINKED ISSUE #54: Set up CI

ghickman:
This sets up GitHub Actions to mirror my preferred CircleCI workflow, specifically a build step, followed by "trivial" checks, finished up with the [potentially] heavier tests step (we could add a deploy step after tests, I'll save that for #49!).

This is my first time really using GitHub Actions so I've used various open source repos (including our own) alongside the docs to build this config.

One thing to take note of - I've used `venv` from the stdlib to allow us to cache the virtualenv in the build step.  However activating that before each run seems a little clunky.

Fixes #54

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,106 @@
+---
+name: CI
+
+on:
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.6"
+      - name: "Install dependencies"
+        run: |
+          set -xe
+          python -m pip install --upgrade pip
+          python -m venv venv
+          venv/bin/pip install --progress-bar=off --requirement requirements.txt
+      - name: Cache the venv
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,106 @@
+---
+name: CI
+
+on:
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
```
sebbacon, line 8: Is this necessary? Any meaningful PR creation must either follow or precede some kind of push... I think?

ghickman: If I understand [this](https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/2) this block will run actions on a push to `master` OR a PR created with a target of `master`.  So feature branches won't have actions run until they turn into a PR.

On rethinking this, I think we want this to trigger on any push (and thus any PR)?  Otherwise you could be working on a feature branch and only find out after creating your PR that something is broken according to CI.

ghickman: Based on [this](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) I've removed the push filter and all of the PRs section in f18cb85.

sebbacon: :+1:

PR #71, opened 2020-08-05
Add Pre-Commit

ghickman:
[Pre-Commit](https://pre-commit.com/) is a framework for running tools before committing via Git's pre-commit hook.

I find it really useful as a smoke test before committing since it runs the configured tools on the diff being committed, and for many of them will do the Right Thing™ (eg black will modify the files which need changes).  It also isn't a requirement, so if this flow doesn't work for someone they can simply not install the hook.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = true
 skip_glob = [".direnv", "*migrations*", "tmp", ".venv"]
+known_third_party = ["crispy_forms", "django", "fabric", "hypothesis", "pytest", "pytest_django", "sentry_sdk"]
 
 [tool.pytest.ini_options]
 env = [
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -17,6 +17,7 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = true
 skip_glob = [".direnv", "*migrations*", "tmp", ".venv"]
+known_third_party = ["crispy_forms", "django", "fabric", "hypothesis", "pytest", "pytest_django", "sentry_sdk"]
```
inglesp, line 20: What's this for?

ghickman: It's seeded by [one of configured pre-commit tools](https://github.com/asottile/seed-isort-config) - when it finds an import of a module not in the project or stdlib it seeds that isort option.  It's an extra level of stability for isort which I've seen trip up on some files from time to time.

PR #77, opened 2020-08-06
Bump pytest to 6.x

ghickman:
This bumps Pytest to 6.x.  I must have forgotten to do it as part of #59, sorry about that!


ghickman:
I think it just hasn't reach this point in the updates yet - it seems to be a handful per day, it's set up to run daily but it only seems to do 5-7 per day so far.  I didn't want to wait for it to get to pytest in case it broke the tests!

PR #78, opened 2020-08-06
Switch CharField for TextField

ghickman:
SQLite doesn't [enforce limits on varchars](https://www.sqlite.org/faq.html#q9) so our max_lengths were being ignored.  Under the hood SQLite stores varchars as TEXT, the same as Postgres.  Since neither database is using a special storage type for varchars this converts all our CharField instances to TextField.

This fixes the snomed fixture not loading when using Postgres because the Description.effective_time field was set to 8 characters but needed 10.

@inglesp – do we need to enforce these max lengths for any reason I'm missing?  I realise this is a bit of a sweeping change!


inglesp:
Blimey.  What implication does this have for `ModelForm`s?

ghickman:
> Blimey.

Sorry!

>  What implication does this have for ModelForms?

Good point!  They'll render as `<textarea>`s by default, so we'll need to change their widgets to `TextInput`.  Are you happy with making that part of this change?

inglesp:
Rather than changing `CharField` to `TextField` everywhere (and requiring us to mess about with widgets, losing validation, etc) think it'd be less invasive to change the lengths on the problematic fields.  It'd be even better to store dates in a `DateField` and I've no idea why `Description.effective_time` and `Relationship.effective_time` are not `DateField`s.

As you've not got a working SQLite, I'll work out how to migrate the problem fields locally, and will push a new PR.

PR #80, opened 2020-08-06
Set up Postgres

ghickman:
This switches the project to support Postgres by default by installing `psycopg2` instead of `pysqlite3[-binary]`.

Locally I have sqlite3 `3.24.0` from my OS (macOS 10.14 Mojave) which has the `OMIT_LOAD_EXTENSION` compile option set.  This is problematic for [pysqlite3](https://github.com/coleifer/pysqlite3/issues/1) because it needs to load extension(s).  I've managed to work around this by cloning pysqlite3 and building locally once… but can't seem to replicate the it.  However the issue still remains that I can't do `pip install -r requirements.txt` because `pysqlite-binary` isn't compatible with macOS and `pysqlite3` will install but won't run because of the load extension issue.

I've set up settings.py such that we're still using SQLite by default (so there should be no change locally for you both) but one can set DATABASE_URL to change that to Postgres.

However, I'm potentially missing something with "downgrading" the SQLite driver to the stdlib one?

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -12,14 +12,10 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 
 
 import os
-import sys
 
+import dj_database_url
 import sentry_sdk
 from sentry_sdk.integrations.django import DjangoIntegration
-
-# Patch sqlite3 to ensure recent version
-__import__("pysqlite3")
-sys.modules["sqlite3"] = sys.modules.pop("pysqlite3")
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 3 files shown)


evansd:
This may well be fine for now, but I think the right long term solution is to use docker for local dev so we avoid these kind of problems entirely (and admittedly generate a new set of problems, but hopefully more tractable ones, and ones with more generalisable solutions)

ghickman:
I've also found the SQL in `coding_systems/snomedct/coding_system.py` isn't valid in Postgres because you can't use SELECT aliases in a WHERE clause.

> as I mentioned on the call the other day there are some reasons to use it

I _knew_ we'd discussed it somewhere!

Given we have SQL tied to SQLite and I don't really want to go and work out how to translate between the two dialects I'm gong to close this PR and see if I can work out how to install pysqlite3 on macOS and document that instead.

---

> use docker for local dev

That's one solution, it's not amazing on macOS since it's still techncially a VM under the hood, but it could be workable.

PR #84, opened 2020-08-07
Validate CSVs

### LINKED ISSUE #48: Validate CSV structure
In the first instance, we should validate that every row of an uploaded CSV file contains the same number of columns.  If a CSV file fails validation, the user should get a helpful message.

This applies to creating a codelist, creating a version, and updating a version.

ghickman:
This validates CSVs uploaded when creating a new CodeList or Version, and when updating an existing Version.  It ensures each row of the CSV has the same number of colmuns as there are headers/columns in the first row.

I chose to split out the Create/Update Version forms into separate views to simplify showing the user errors (I suspect I could have made the modals open on load but I wasn't confident so I went with what I knew!).  So I've cut down the `version` view to no longer handle forms and added `VersionCreate` and `VersionUpdate` to handle them instead.

**Create Version**

**Update Version**

Fixes #48

### DIFF
--- a/codelists/views.py
+++ b/codelists/views.py
@@ -73,7 +73,7 @@ class CreateCodelist(TemplateView):
             coding_system_id=form.cleaned_data["coding_system_id"],
             description=form.cleaned_data["description"],
             methodology=form.cleaned_data["methodology"],
-            csv_data=form.cleaned_data["csv_data"].read().decode("utf8"),
+            csv_data=form.cleaned_data["csv_data"],
             references=references,
             signoffs=signoffs,
         )
--- a/codelists/views.py
+++ b/codelists/views.py
@@ -1,11 +1,10 @@
-from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.db.models import Q
 from django.forms import formset_factory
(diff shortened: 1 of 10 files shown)


codelists/views.py
```diff
@@ -73,7 +73,7 @@ def all_valid(self, form, reference_formset, signoff_formset):
             coding_system_id=form.cleaned_data["coding_system_id"],
             description=form.cleaned_data["description"],
             methodology=form.cleaned_data["methodology"],
-            csv_data=form.cleaned_data["csv_data"].read().decode("utf8"),
+            csv_data=form.cleaned_data["csv_data"],
```
inglesp, line 75: ?

ghickman: In [`CodelistVersionForm.clean_csv_data` (inherited from `CSVValidationMixin`)](https://github.com/opensafely/opencodelists/pull/84/files#diff-dce4bf0e7b1fc6b64bd36f4b7f656403R50) we need to read the contents to validate it.  Since we'd already read the contents into a variable, it seemed natural/in-keeping with how Forms often work to return the contents of that variable.

PR #87, opened 2020-08-10
Publish Draft Versions

### LINKED ISSUE #50: Add publish button
On the page for a draft version, there should be a "Publish version" button underneath "Update version" button, visible only to users who are signed in.

When clicked, this should open an "are you sure?" modal, and if "yes" is clicked, this should submit a POST which calls `codelists.actions.publish_version`.  If "no" is clicked, the modal should be dismissed.

ghickman:
This adds a new button to the LHN controls:

Pops up a modal when clicking this button:

And calls the appropriate action to publish the given Version.

The final commit returns the mutated Version to make redirecting to it cleaner.

Note: the first commit removes some noise from the tests output but is expected to not affect deprecations raised by our code calling deprecated APIs.

Fixes #50

### DIFF
--- a/codelists/views.py
+++ b/codelists/views.py
@@ -215,7 +215,7 @@ def version_publish(request, project_slug, codelist_slug, qualified_version_str)
     if expect_draft != version.is_draft:
         return redirect(version)
 
-    actions.publish_version(version=version)
+    version = actions.publish_version(version=version)
 
     return redirect(version)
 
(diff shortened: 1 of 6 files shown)


codelists/views.py
```diff
@@ -215,7 +215,7 @@ def version_publish(request, project_slug, codelist_slug, qualified_version_str)
     if expect_draft != version.is_draft:
         return redirect(version)
 
-    actions.publish_version(version=version)
+    version = actions.publish_version(version=version)
```
inglesp, line 218: This is a good spot, and it's not obvious why it's necessary.  Perhaps the commit message could be included as a comment?

PR #89, opened 2020-08-11
Editing Codelists

### LINKED ISSUE #47: Allow codelists to be edited
There should be a page where signed in users can edit the following fields:

* `name`
* `project`
* `coding_system_id`
* `description`
* `methodology`
* `signoffs`
* `references`

ghickman:
This adds the ability for a logged in User to edit an existing Codelist.

I've dropped Crispy Forms for the forms in the two formsets.  Instead they each have a template so the `empty_form` template can be generated easily too.

The new related forms look like this:

It also combines the templates and Javascript for the Create and Update views since they're so similar now, particularly the related forms parts.

Fixes #47


codelists/actions.py
```diff
@@ -67,6 +67,31 @@ def create_version(*, codelist, csv_data):
...
+    """Update a Codelist."""
+
+    codelist.project = project
+    codelist.name = name
+    codelist.coding_system_id = coding_system_id
+    codelist.description = description
+    codelist.methodology = methodology
+
+    codelist.save(
+        update_fields=[
```
inglesp, line 83: TIL.

static/js/codelist.js
```diff
@@ -0,0 +1,43 @@
...
+
+    event.preventDefault();
+}
+
+$(document).ready(() => {
+    $('#add-reference').click((event) => addForm('reference', event));
+    $('#add-signoff').click((event) => addForm('signoff', event));
+
+    $('.remove-reference').bind('click', () => removeRow('reference', event));
+    $('.remove-signoff').bind('click', () => removeRow('signoff', event));
```
inglesp, line 42: What's `event` here?

templates/codelists/version.html
```diff
@@ -38,6 +38,17 @@ <h3>{{ codelist.name }}</h3>
 
     {% if user.is_authenticated %}
     <hr />
+
+    <div class="btn-group-vertical btn-block" role="group">
+      <a
+        class="btn btn-outline-info btn-block"
+        href="{% url 'codelists:codelist-edit' project_slug=codelist.project.slug codelist_slug=codelist.slug %}">
+        Edit
```
inglesp, line 46: Could this be "Edit codelist metadata"?

codelists/actions.py
inglesp: Not sure it's necessary though, and means extra work if we add another field.

codelists/views.py
```diff
@@ -113,6 +112,109 @@ def get_signoff_formset(self, data=None, files=None):
         return self.SignOffFormSet(data, files, prefix="signoff")
 
 
+@method_decorator(login_required, name="dispatch")
+class CodelistUpdate(TemplateView):
+    ReferenceFormSet = modelformset_factory(
+        Reference, form=ReferenceForm, can_delete=True
+    )
+    SignOffFormSet = modelformset_factory(SignOff, form=SignOffForm, can_delete=True)
```
inglesp, line 120: Do these definitions belong in `forms.py`?

codelists/actions.py
ghickman: That's true, this is more a habit of mine after working in some databases where tables had large fields.  I'll remove this as I don't think it adds anything performance or readability wise!

codelists/views.py
ghickman: I hadn't even considered putting them in there, good call!

static/js/codelist.js
ghickman: `undefined` because they're missing from the function declaration 😬

codelists/actions.py
ghickman: Fixed in 7abe19b

static/js/codelist.js
ghickman: Fixed in 13bac65

templates/codelists/version.html
ghickman: I went with `Edit metadata` since the longer form overlapped the button boundary:



Fixed in 1912ea1

codelists/views.py
ghickman: Fixed in 84b5596

PR #92, opened 2020-08-11
List Versions

ghickman:
This lists all versions of a Codelist in the LHN as links.  It marks drafts with a badge and doesn't link the Version currently being viewed.

### DIFF
--- a/templates/codelists/version.html
+++ b/templates/codelists/version.html
@@ -55,13 +55,13 @@
         href="{% url 'codelists:version-create' project_slug=codelist.project.slug codelist_slug=codelist.slug %}">
         Create new version
       </a>
+
       {% if clv.is_draft %}
       <a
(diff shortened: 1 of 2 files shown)


templates/codelists/version.html
```diff
@@ -70,8 +70,29 @@ <h3>{{ codelist.name }}</h3>
...
       </button>
       {% endif %}
-      {% endif %}
     </div>
+
+    <hr />
+
+    <h6>Versions</h6>
+    <ul class="pl-3">
+      {% for version in versions %}
```
inglesp, line 79: How are these ordered?

ghickman: The view is doing `order_by("-version_str")`

inglesp: Totally missed that.

PR #93, opened 2020-08-11
Add Test Coverage

ghickman:
This adds test coverage via pytest-cov/coverage.py.

I've opted to skip some modules which I don't think are currently of interest to us, and removed some which coverage pointed out weren't used (autogenerated `apps.py`s).

I've turned on skipping modules with 100% coverage in reports because seeing the list get smaller is a nice positive achievement and gives you fewer things to read!

I've configured it such that `make test` runs coverage (via the `--cov` flags) so we can have coverage by default (and in CI) but a developer can still run pytest without coverage, which is good for tight feedback loops.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,4 +48,9 @@ sort:
 
 .PHONY: test
 test:
-	python manage.py test
+	pytest \
+		--cov=builder \
+		--cov=codelists \
+		--cov=coding_systems \
+		--cov=mappings \
+		--cov=opencodelists \
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,6 @@ environment
 tmp
 venv
 __pycache__/
+.coverage
 .hypothesis
+htmlcov
--- a/builder/apps.py
+++ b/builder/apps.py
@@ -1,5 +1,0 @@
-from django.apps import AppConfig
-
-
-class BuilderConfig(AppConfig):
-    name = "builder"
--- a/codelists/apps.py
+++ b/codelists/apps.py
@@ -1,5 +1,0 @@
-from django.apps import AppConfig
-
-
-class CodelistsConfig(AppConfig):
-    name = "codelists"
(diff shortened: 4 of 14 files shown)


Makefile
```diff
@@ -48,4 +48,9 @@ sort:
 
 .PHONY: test
 test:
-	python manage.py test
+	pytest \
+		--cov=builder \
+		--cov=codelists \
+		--cov=coding_systems \
+		--cov=mappings \
+		--cov=opencodelists \
```
inglesp, line 56: 1) Is there a good reason this isn't in the `pyproject.toml`?
2) Could we have `--source .` instead?  This pick up the right files, unless you've got a virtual environment in the same directory, in which case that could be omitted explicitly.

ghickman:
> Why do we both omit stuff and also include stuff with --cov?

> I'd like to avoid having to add new apps to the invocation in the makefile though.

There's a few reasons here!

I dislike configuring too much of pytest or its plugins in config since I often don't want all the features when I'm running 1 or a small subset of tests (eg `some/path/test_views.py::test_foo` or `some/path/test_views.py -k myview`).  If I'm debugging a test/code problem I often run a tight feedback loop with a file watcher.  So extra output or slower tests are a noticeable annoyance!

Coverage.py does come with a [performance penalty](https://coverage.readthedocs.io/en/coverage-5.2.1/source.html#specifying-source-files) and my interpretation of the page linked is they expect you to specify paths (granted it's not clear if `.` is expected!).  However I suspect that's negligible for a project of this size.

I've tried configuring pytest with `--cov` and `--cov=.` which sets source=None and `['.'] respectively in pytest-cov but we lose 0% covered modules.  I haven't managed to track down _why_ yet (which is somewhat frustrating!).

---

If everyone is happy to do so I think we leave it configured as is so it shows those 0% modules and I'll take another bash at this next week with a fresh brain.

sebbacon:
:shipit:

PR #111, opened 2020-08-20
Improve the codelist builder

inglesp:
@ghickman as discussed, I'm not sure how reviewable this is.  In particular, the frontend code is not tested or optimised and I know I need to do further work on it to make it "done".

Perhaps you could review the changes to `tree_utils`?


builder/actions.py
```diff
@@ -1,63 +1,44 @@
...
+def delete_search(*, search):
+    search.delete()
+    search.codelist.codes.annotate(num_results=Count("results")).filter(
+        num_results=0
+    ).delete()
+
+
+@transaction.atomic
+def update_code_statuses(*, codelist, updates):
+    print(updates)
```
ghickman, line 34: Stray print 🙂

builder/actions.py
```diff
@@ -1,63 +1,44 @@
...
-            code=code,
-            matches_term=(code in results["matching_codes"]),
-            is_ancestor=(code in results["ancestor_codes"]),
-        )
-        for code in results["all_codes"]
-    )
+def create_search(*, codelist, term, codes):
+    search = codelist.searches.create(term=term, slug=slugify(term))
+    for code in codes:
+        search.results.create(code=codelist.codes.get_or_create(code=code)[0])
```
ghickman, line 20: One for your next pass:

Could the generation of Codelist objects be pulled out of this for loop?

Maybe an inner generator function doing `codelist, _ = …` so you can yield codelist?

It's a slight efficiency improvement because you can then do a bulk_create of Result instances, but I think mainly it would improve readability to show that you're `get_or_create`ing Codelists first.

builder/views.py
```diff
@@ -47,143 +47,58 @@ def user(request, username):
...
+        codes_for_display = {c for c in codes_for_display if "+" in code_to_status[c]}
+    elif filter == "excluded":
+        codes_for_display = {c for c in codes_for_display if "-" in code_to_status[c]}
+    elif filter == "unresolved":
+        codes_for_display = {c for c in codes_for_display if code_to_status[c] == "?"}
+    elif filter == "in-conflict":
+        codes_for_display = {c for c in codes_for_display if code_to_status[c] == "!"}
+        filter = "in conflict"
+    else:
+        codes_for_display = set(codes_for_display)
```
ghickman, line 77: It looks like the only thing you're changing is the conditional in the comprehension, so could probably use the dispatch pattern?

codelists/tree_utils.py
```diff
@@ -100,19 +166,203 @@ def helper(tree):
...
+
+        # some ancestors are included and some are excluded, and neither set of
+        # ancestors overrides the other
+        return "!"
+
+    return {node: render_node(node) for node in ancestors_map}
+
+
+def find_ancestors(subtree, nodes):
+    """Given set of nodes, return subset which are ancestors of other nodes.  An
```
ghickman, line 363: s/other/those/ ?

ghickman (approved):
A couple of possible improvements but don't gate this PR on them.

Loving the tree_utils and testing comments 👌

codelists/tree_utils.py
inglesp: I've rewritten the docstring and have provided an example.

builder/views.py
inglesp: This whole function is going to be rewritten, so I won't do that now.

PR #112, opened 2020-08-20
Refactor builder.js

inglesp:
This PR refactors builder.js.  It breaks a big component up into lots of smaller ones, and uses props correctly.

### DIFF
--- a/static/js/builder.js
+++ b/static/js/builder.js
@@ -43,6 +43,9 @@ class CodelistBuilder extends React.Component {
 
     this.updateStatus = this.updateStatus.bind(this);
     this.toggleVisibility = this.toggleVisibility.bind(this);
+    this.getStatus = this.getStatus.bind(this);
+    this.getIsExpanded = this.getIsExpanded.bind(this);
+    this.getIsVisible = this.getIsVisible.bind(this);
   }
 
   updateStatus(code, status) {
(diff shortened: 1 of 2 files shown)


static/js/builder.js
```diff
@@ -1,56 +1,34 @@
...
+      this.state["status-" + code] = props.codeToStatus[code];
       this.state["expanded-" + code] = true;
     });
 
     this.updateStatus = this.updateStatus.bind(this);
     this.toggleVisibility = this.toggleVisibility.bind(this);
+    this.getStatus = this.getStatus.bind(this);
+    this.getHasDescendants = this.getHasDescendants.bind(this);
+    this.getIsExpanded = this.getIsExpanded.bind(this);
+    this.getIsVisible = this.getIsVisible.bind(this);
```
ghickman, line 20: You can avoid this step by doing `someProp={() => this.someMethod()}` at the call site.

inglesp: Doesn't this create a new function on every render?

static/js/builder.js
```diff
@@ -392,7 +461,21 @@ function getCookie(name) {
   return cookieValue;
 }
 
+function readValueFromPage(id) {
+  return JSON.parse(document.getElementById(id).textContent);
+}
```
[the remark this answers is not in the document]
inglesp: That's quite neat.

static/js/builder.js
ghickman: [Apparently so](https://reactjs.org/docs/faq-functions.html#arrow-function-in-render), TIL!

PR #132, opened 2020-08-27
Upsert CTV3 -> SNOMED Mappings

ghickman:
This changes how we import CTV3 -> SNOMED mapping records.  The mapping file includes deactivated maps so instead of only ingesting active maps we need to consider those as well.  Rows in this file with the same Map ID are disambiguated using the Effective Date field (eg we only need to care about the latest version of any Map ID).  This uses an UPSERT query to handle this.

As a nice unintended bonus ingesting is now ~2x fast.

### DIFF
--- a/mappings/ctv3sctmap2/import_data.py
+++ b/mappings/ctv3sctmap2/import_data.py
@@ -1,45 +1,83 @@
 import csv
-import datetime
 import glob
 import os
+import sqlite3
 
-from django.db import transaction
-
-from .models import Mapping
+from django.db import connection as django_connection
 
 
 def import_data(release_dir):
-    def load_records():
-        paths = glob.glob(
-            os.path.join(
-                release_dir,
-                "Mapping Tables",
-                "Updated",
-                "Clinically Assured",
-                "ctv3sctmap2*.txt",
-            )
+    """
+    Import NHSD CTV3 -> SNOMED concept maps
+    """
+    paths = glob.glob(
+        os.path.join(
+            release_dir,
+            "Mapping Tables",
+            "Updated",
+            "Clinically Assured",
+            "ctv3sctmap2*.txt",
         )
-        assert len(paths) == 1
+    )
+    assert len(paths) == 1
 
-        with open(paths[0]) as f:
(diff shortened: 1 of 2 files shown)


mappings/ctv3sctmap2/import_data.py
```diff
@@ -1,45 +1,83 @@
...
-
-def parse_date(datestr):
-    return datetime.date(int(datestr[:4]), int(datestr[4:6]), int(datestr[6:]))
+    )
+    assert len(paths) == 1
+
+    with open(paths[0]) as f:
+        reader = csv.DictReader(f, delimiter="\t")
+
+        values = list(iter_values(reader))
```
inglesp, line 27: Don't need to turn this into a list.

mappings/ctv3sctmap2/import_data.py
```diff
@@ -1,45 +1,83 @@
...
+    connection.close()
+
+
+def iter_values(rows):
+    """
+    Convert the given rows into an iterable of values for upserting to the DB.
+    """
+    for r in rows:
+        # ignore SNOMED Concept and Description IDs when the Concept ID is _DRUG
+        sct_concept = r["SCT_CONCEPTID"] if r["SCT_CONCEPTID"] != "_DRUG" else ""
```
inglesp, line 64: For drugs, this has changed from `None` to `""` -- is this deliberate?

inglesp (approved):
:rocket:

mappings/ctv3sctmap2/import_data.py
ghickman: Nope, good catch!

ghickman: Fixed in fe08eb4

mappings/ctv3sctmap2/import_data.py
ghickman: The cast to list means we can unindent the rest of the function and close the file handle here.

inglesp: Fair enough.

PR #136, opened 2020-08-27
Show an Error for Duplicate Codelist Name

### LINKED ISSUE #129: Users sees 500 on duplicate codelist name
They should get a nice error message explaining the problem.

ghickman:
This validates the uniqueness of a Codelist name when submitting the Create Codelist form.  It handles the form not having a Project field by passing in a Codelist instance with the project attached so the form can build a full instance before running validation.

Fixes #129


codelists/models.py
```diff
@@ -26,7 +26,7 @@ class Codelist(models.Model):
     methodology = models.TextField()
 
     class Meta:
-        unique_together = ("project", "slug")
+        unique_together = ("project", "name")
```
inglesp, line 29: We need to keep `unique_together = ("project", "slug")`.  If two codelists are created in the same project called "George's list" (your list) and "Georges list" (a list of people called George -- hey, it could happen) then they'll both get the same slug, and so the `get_object_or_404` (which uses the slug) will find more than one object.

codelists/tests/test_views.py
```diff
@@ -131,6 +132,43 @@ def test_codelistcreate_when_not_logged_in(client):
...
+    request.user = UserFactory()
+    response = CodelistCreate.as_view()(request, project_slug=project.slug)
+
+    assert project.codelists.count() == 1
+
+    # we're returning an HTML response when there are errors so check we don't
+    # receive a redirect code
+    assert response.status_code == 200
+
+    # confirm we have errors from the signoff formset
```
inglesp, line 168: Copy/paste error in the comment?

codelists/views.py
```diff
@@ -105,8 +105,28 @@ def get_context_data(self, **kwargs):
...
+
+        Since the `project` field isn't included in the CodelistCreateForm
+        fields we pass our passed-in Project to the form using the instance
+        keyword argument so it can be used at validation time.
+        """
+        kwargs = {}
+        if project:
+            kwargs["instance"] = Codelist(project=project)
+
+        return CodelistCreateForm(data, files, **kwargs)
```
inglesp, line 129: What happens if you have `return CodelistCreateForm(data, files, instance=Codelist(project=self.project))`?

codelists/forms.py
```diff
@@ -68,6 +68,27 @@ def __init__(self, *args, **kwargs):
...
+        # Call super()'s full_clean to populate the Codelist instance with
+        # values from the form.
+        super().full_clean()
+
+        # Validate uniqueness of the instance now it's been populated with
+        # values from the form and the Project instance.
+        try:
+            self.instance.validate_unique()
+        except forms.ValidationError as e:
+            self._update_errors(e)
```
inglesp, line 90: Do we need this stuff on the `CodelistUpdateForm` too?

inglesp (changes requested):
Looks good (comments especially) but we need to handle updating a codelist, and there's a very edge edgecase I think we should avoid if we can too.

codelists/views.py
ghickman: It's identical, but I try not to rely on self outside of dispatch/the verb methods in CBVs to make testing them with View.setup() easier.

codelists/models.py
ghickman: Good point!  Fixed in db3cdfd

> a list of people called George

There's at least two of us!

ghickman: I've hit upon the same problem as `project` with `slug` - it's not included in the form so we have to add it so we can validate uniqueness.  But that brings up a UX issue - say `George's list` already exists and someone else wants to add `Georges list`.  They'll get an error on the page which says something like "Codelist must have a unique Project, Name, and Slug."  However, they can't change the slug we generate and are left with modifying the Title.

Is that the flow we want to push for?  If not should we look at updating the slug to avoid a known collision?

inglesp: This is rare enough that I think it's acceptable to tell the user that there's a duplicate slug, and to leave them to work out the rest.

codelists/views.py
ghickman: Based on testing chat in slack, this makes use of self.project in d565d39

codelists/models.py
ghickman: Fixed in 62bc490

codelists/forms.py
ghickman: Yes, extracted to a mixin (after much testing pain 😫) in ffcc49f

ghickman:
@inglesp – I think this is good for another review now, particularly because I wrapped the `create_codelist` factory in a Factory Boy factory (factory factory factory 🤪) to make configuring it at call sites easier, or so I'm hoping!  I'd love to hear how you felt about the API that produces 🙂

Also, I extracted the validation of Codelist uniqueness out of the two Codelist forms and into a Mixin as they were identical.  However, that still leaves two very similar ModelForms for the same Model… should we make a CodelistForm and have the CodelistCreateForm be a subclass which adds the CSV field?

codelists/forms.py
```diff
@@ -56,7 +89,7 @@ def clean_csv_data(self):
         return data
 
 
-class CodelistCreateForm(forms.ModelForm, CSVValidationMixin):
+class CodelistCreateForm(CodelistUniquenessMixin, forms.ModelForm, CSVValidationMixin):
```
inglesp, line 92: Interesting choice of sorting mixins alphabetically, by lower-casing the class names and stripping the first letter...

codelists/forms.py
```diff
@@ -36,6 +37,38 @@ class Meta:
...
+        # populate the instance's slug so the validate_unique() check below can
+        # use it to check uniqueness
+        self.instance.slug = slugify(self.instance.name)
+
+        # Validate uniqueness of the instance now it's been populated with
+        # values from the form and the Project instance.
+        try:
+            self.instance.validate_unique()
+        except forms.ValidationError as e:
+            self._update_errors(e)
```
inglesp, line 69: Nit: you could call call `self.validate_unique` directly here, which avoids prodding at parts of the private API.

inglesp (approved):
Looks good.

The Factory Boy stuff is fine, and looks like it gives all the flexibility we need, while keeping the simple case simple.

codelists/forms.py
ghickman: Hah!  Not quite though…

Superclass/mixin order is important here unfortunately.  CodelistUniquenessMixin's `full_clean()` needs to be the direct parent of `CodelistCreateForm` so _it's_ call to `super()` calls `ModelForm` to populate the instance values from the Form values so it can then handle the uniqueness check for us.

codelists/forms.py
ghickman: `self` here is a ModelForm instance and `validate_unique` looks like it's only exposed as a [Model method](https://docs.djangoproject.com/en/3.1/ref/models/instances/#django.db.models.Model.validate_unique), am I missing something?

Good spot on the private API though, that should be a call to `add_error(None, e)`!

inglesp: It's also a [method on `BaseModelForm`](https://github.com/django/django/blob/master/django/forms/models.py#L413-L422), and does almost exactly what you're doing here.

codelists/forms.py
inglesp: Oof.

codelists/forms.py
ghickman: TIL 🎉  Thanks!

ghickman: We can't use `BaseModelForm.validate_unique` because it drops fields not defined on the form (I've added more information in a comment which hopefully gives enough context in 15e56e4), bit frustrating!

I've switch to `add_error(None, e)` to avoid the private API in cb95f37.

inglesp: Thanks.  Don't think it's worth spending any more time on this!

Issue #100, opened 2020-08-13
Adapt tree view for SNOMED CT codelists


inglesp:
The tree view of a codelist is intended all the codelist's concepts in the context of the coding system's hierarchy.  [Here's an example of a CTV3 codelist](https://codelists.opensafely.org/codelist/opensafely/aplastic-anaemia/2020-04-24/#tree).  However the current approach does not work for SNOMED CT codelists.

Most coding system hierarchies allow a concept to have multiple parents, and so there are often multiple paths to a concept from the root of the hierarchy.  That is, the hierarchies are directed acyclic graphs, not tree.  These can be difficult to display meaningfully.

We cope with this by showing each concept in the codelist multiple times, once per path from the root.  In other words, we convert the DAG to a tree.  In the example above, Adult Fanconi Syndrome appears 3 times.

This is practical for CTV3 codelists, because relatively few CTV3 concepts have multiple parents.  However, the SNOMED CT hierarchy is much richer and most SNOMED CT concepts have multiple parents.  This means that there can be hundred of paths from the root to any given concept.  For instance, there are 507 paths from the root to "Acute severe exacerbation of mild persistent allergic asthma co-occurrent with allergic rhinitis (disorder)".

This makes viewing the tree difficult, as a concept can also appear hundreds of times.  Additionally, it makes it expensive to compute, transfer, and render the tree.  (I don't have numbers, but when I tried this out, I recall it taking tens of seconds to compute the tree structure; tens of seconds to render the template in Django; and minutes to render the HTML in the browser.  All of these could be optimised, but there seemed little point if the end result is unusable.)

So for now, we have disabled the tree view for SNOMED CT codelists, which limits the usefulness of the page.

A proposed solution is that instead of showing a single tree whose root is the root of the hierarchy, we show several trees -- one for each concept in the codelist that has no ancestors in the codelist.  These should be grouped together by top-level concept.  This would mean that "Acute severe exacerbation of mild persistent allergic asthma co-occurrent with allergic rhinitis (disorder)" appears only 15 times in the asthma codelist tree.

This is the approach I've used in the prototype codelist builder, where it seems to work:

![image](https://user-images.githubusercontent.com/28734/90155989-dc9dba00-dd83-11ea-9b4b-f653bc428eda.png)

The structure of the tree, and any HTML, could be precomputed/cached for performance if necessary

(The page could be made a bit nicer visually while we're at it!)

PR #138, opened 2020-08-28
Use hierarchies in more places

inglesp:
Notes:

* there's some duplication in setting up database fixtures -- I'll come back to working out the pytesty way to do this another time
* the test for `build_html_tree_highlighting_codes` is only interesting for showing that there are no regressions -- when we tackle #100 we can add proper tests

### DIFF
--- a/codelists/presenters.py
+++ b/codelists/presenters.py
@@ -1,0 +1,43 @@
+from django.urls import reverse
+
+from .tree_utils import walk_tree_depth_first
+
+
+def build_html_tree_highlighting_codes(coding_system, hierarchy, subtree, definition):
+    codes = definition.codes(hierarchy)
+    codes_in_definition = {r.code for r in definition.including_rules()}
+    codes_in_tree = set(walk_tree_depth_first(subtree))
+    code_to_name = coding_system.lookup_names(codes_in_tree)
+    sort_key = code_to_name.__getitem__
+
+    lines = []
+    last_direction = 1
+    for code, direction in walk_tree_depth_first(subtree, sort_key):
+        if direction == last_direction == 1:
+            lines.append("<ul>")
+        elif direction == last_direction == -1:
+            lines.append("</ul>")
+
+        last_direction = direction
+
+        if direction == -1:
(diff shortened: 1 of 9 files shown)


ghickman (approved):
Looks great, the hierarchy stuff definitely seems like a good fit as we use it more.

I particularly like the idea of pulling out presentation rendering into a module 👍

Not for this PR but given the test:
```
def test_build_html_tree_highlighting_codes():
    fixtures_path = Path(settings.BASE_DIR, "coding_systems", "snomedct", "fixtures")
    call_command("loaddata", fixtures_path / "core-model-components.json")
    call_command("loaddata", fixtures_path / "tennis-elbow.json")

    with open(fixtures_path / "disorder-of-elbow-excl-arthritis.csv") as f:
        cl = create_codelist(csv_data=f.read())

    coding_system = cl.coding_system
    clv = cl.versions.get()
    hierarchy = Hierarchy.from_codes(coding_system, clv.codes)
    definition = Definition.from_codes(set(clv.codes), hierarchy)

    html = presenters.build_html_tree_highlighting_codes(
        coding_system, hierarchy, definition
    )

    with open(
        Path(settings.BASE_DIR, "codelists", "tests", "expected_html_tree.html")
    ) as f:
        assert html.strip() == f.read().strip()
```

I would consider pulling out up to 5 pytest fixtures:

1. fixtures_path
1. core_model_components
1. tennis_elbow
1. disorder_of_elbow_excluding_arthritis
1. expected_html_tree

The first two look very useful based on other tests we've got.  The third, potentially, but I'm not sure it's used elsewhere?  However I'm very dubious of the last two as pytest fixtures, they look very specific to this test and I have a feeling they'd make the test harder to grok by hiding useful details.

inglesp:
> The first two look very useful based on other tests we've got. The third, potentially, but I'm not sure it's used elsewhere? However I'm very dubious of the last two as pytest fixtures, they look very specific to this test and I have a feeling they'd make the test harder to grok by hiding useful details.

Makes sense.

PR #139, opened 2020-08-28
Add Logging

ghickman:
This adds [more] logging to the project via [Structlog](https://www.structlog.org/en/stable/) and it's Django companion package [django-structlog](https://django-structlog.readthedocs.io/en/latest/index.html).

django-structlog binds request and user IDs to the log context allowing us to emit logs from inside our actions and get this kind of output for submitting the Codelist creation form:

```
2020-08-28T07:41:27.826096Z [info     ] request_started                [django_structlog.middlewares.request] ip=127.0.0.1 request=<WSGIRequest: POST '/codelist/test/'> request_id=d3804098-c2c5-4cd3-8e80-b1e3ee2b94a0 user_agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:80.0) Gecko/20100101 Firefox/80.0 user_id=ghickman
2020-08-28T07:41:27.891123Z [info     ] Codelist created               [codelists.actions] codelist_pk=8 ip=127.0.0.1 request_id=d3804098-c2c5-4cd3-8e80-b1e3ee2b94a0 user_id=ghickman
2020-08-28T07:41:27.891862Z [info     ] request_finished               [django_structlog.middlewares.request] code=302 ip=127.0.0.1 request=<WSGIRequest: POST '/codelist/test/'> request_id=d3804098-c2c5-4cd3-8e80-b1e3ee2b94a0 user_id=ghickman
```

I've configured both Django and Gunicorn to use the new logging config.  This will normalise our logs in production which currently emit different formats for Django vs Gunicorn (with the latter emitting two timestamps).

---

#### Services Package
I've placed the logging configuration into `services/logging.py` to separate it from Django's settings.  I don't typically create a services package until I have something which is _actually_ a service (in the [12 factor app](https://12factor.net/backing-services) sense), but a core reason I abstract those from Django settings is to avoid poluting its global namespace with config for those services.  In this instance it provides two benefits: 
1. I want to configure Gunicorns logs so the config can't be inside Django anyway.
1.  I don't want to add `pre_chain` and `timestamper` as things one can import from `django.conf.settings`.

Of course this is Python and there's nothing to stop someone doing `from services.logging import pre_chain` but the layout aims to keep settings.py simpler without removing access to a user who wants those lower level parts.

I wrote up this pattern as an [Architecture Pattern for Mozilla's Treeherder](https://github.com/mozilla/treeherder/blob/5c849ef3b20f4bb29a9d5a0ceba02460c946f696/docs/arch/adr-001.rst) (the doc didn't get merged for unrelated reasons but the pattern did).

### DIFF
--- a/builder/actions.py
+++ b/builder/actions.py
@@ -1,16 +1,23 @@
 from collections import defaultdict
 
+import structlog
 from django.db import transaction
 from django.db.models import Count
 from django.utils.text import slugify
 
 from codelists.hierarchy import Hierarchy
 
+logger = structlog.get_logger()
+
 
 def create_codelist(*, owner, name, coding_system_id):
-    return owner.draft_codelists.create(
+    codelist = owner.draft_codelists.create(
         name=name, slug=slugify(name), coding_system_id=coding_system_id
     )
+
+    logger.info("Create Codelist", codelist_pk=codelist.pk)
+
+    return codelist
 
 
 @transaction.atomic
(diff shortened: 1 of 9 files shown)


inglesp (approved):
Thanks for setting this up.

PR #140, opened 2020-08-31
Start to use hierarchies in builder

inglesp:


codelists/tests/test_presenters.py
```diff
@@ -38,3 +39,91 @@ def test_build_html_tree_highlighting_codes():
...
         assert html.strip() == f.read().strip()
+
+
+def test_tree_tables():
+    fixtures_path = Path(settings.BASE_DIR, "coding_systems", "snomedct", "fixtures")
+    call_command("loaddata", fixtures_path / "core-model-components.json")
+    call_command("loaddata", fixtures_path / "tennis-elbow.json")
+
+    with open(fixtures_path / "disorder-of-elbow.csv") as f:
+        cl = create_codelist(csv_data=f.read())
```
ghickman, line 50: We've done these steps in a few tests now, and while I don't thinking DRYing out tests is a good thing, it might be time to think about turning these steps into fixtures?  (not in this PR though!)

ghickman (approved):
Great use of comments to explain the hierarchy -> presentation transforms again 👌

codelists/tests/test_presenters.py
inglesp: Agree.

PR #149, opened 2020-09-01
Add Definition Presenter

ghickman:
This adds the Definition presentation function.  It converts a given Definition instance into a structure (list of dictionaries) ready for display in a template.  The template is now in charge of styling rows of a Definition.

This is a rebuild of #110 on top of @inglesp's new `Hierarchy` object.

### DIFF
--- a/codelists/presenters.py
+++ b/codelists/presenters.py
@@ -29,7 +29,7 @@ class DefinitionRow:
     all_descendants: bool = attr.ib(default=True)
 
 
-def _iter_rules(hierarchy, rules, name_for_rule, excluded=None):
+def _iter_rules(hierarchy, rules, name_for_rule, excluded):
     for rule in rules:
         row = DefinitionRow(
             name=name_for_rule(rule),
@@ -48,9 +48,14 @@ def _iter_rules(hierarchy, rules, name_for_rule, excluded=None):
         ]
 
         # generate excluded children by recursing into this function
-        row.excluded_descendants = list(
-            _iter_rules(hierarchy, excluding_rules, name_for_rule)
-        )
+        row.excluded_descendants = [
+            DefinitionRow(
+                name=name_for_rule(rule),
+                code=rule.code,
+                all_descendants=rule.applies_to_descendants,
+            )
+            for rule in excluding_rules
+        ]
 
         yield attr.asdict(row)
 
--- a/codelists/presenters.py
+++ b/codelists/presenters.py
@@ -5,33 +5,33 @@ from django.urls import reverse
 
 
 @attr.s
-class Row:
+class DefinitionRow:
     """
     Data structure for a Definition to prepare for display
 
     name: name of the definition
     code: code of the definition
-    excluded_children: list of excluded children (created with this class too)
+    excluded_descendants: list of excluded children (created with this class too)
     all_descendants: are all of this Definitions descendants included?
 
-    Between all_descendants and excluded_children we can cover the three state
-    situation a Definition can be in:
+    Between all_descendants and excluded_descendants we can cover the three state
+    situation a DefinitionRow can be in:
 
-        * all descendants are included (all_descendants = True, excluded_children = [])
-        * all descendants except N (all_descendants = True, excluded_children = [...])
-        * no descendants (all_descendants = False, excluded_children ignored)
+        * all descendants are included (all_descendants = True, excluded_descendants = [])
+        * all descendants except N (all_descendants = True, excluded_descendants = [...])
+        * no descendants (all_descendants = False, excluded_descendants ignored)
 
     """
 
     name: str = attr.ib()
     code: str = attr.ib()
-    excluded_children: list = attr.ib(default=list())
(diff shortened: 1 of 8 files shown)


codelists/presenters.py
```diff
@@ -1,8 +1,72 @@
 from itertools import groupby
 
+import attr
 from django.urls import reverse
 
 
+@attr.s
+class Row:
```
inglesp, line 8: Could we call this `DefinitionRow`?

codelists/presenters.py
```diff
@@ -1,8 +1,72 @@
...
+
+        * all descendants are included (all_descendants = True, excluded_children = [])
+        * all descendants except N (all_descendants = True, excluded_children = [...])
+        * no descendants (all_descendants = False, excluded_children ignored)
+
+    """
+
+    name: str = attr.ib()
+    code: str = attr.ib()
+    excluded_children: list = attr.ib(default=list())
```
inglesp, line 28: Can you change this to `excluded_descendants`?

codelists/views.py
```diff
@@ -311,7 +311,7 @@ def version(request, project_slug, codelist_slug, qualified_version_str):
         "headers": headers,
         "rows": rows,
         "html_tree": html_tree,
-        "html_definition": html_definition,
+        "definition": definition_rows,
```
inglesp, line 314: Could the key be `"definition_rows"`?

codelists/presenters.py
```diff
@@ -1,8 +1,72 @@
...
+    all_descendants: bool = attr.ib(default=True)
+
+
+def _iter_rules(hierarchy, rules, name_for_rule, excluded=None):
```
inglesp, line 32: This works while a `DefinitionRule` does not have both `applies_to_descendants=True` and `code_is_excluded=True`.  That is currently the case because of how `Definition.from_codes` works.

If this wasn't the case, we'd end up recursing a second time (ie calling the function three times) which would mean that `r in excluded` raises a `TypeError` because `excluded` is `None`.

I wonder whether it's worth adding a check (an `assert` would do) to `DefinitionRule.__init__` to ensure that the the invariant is maintained.

codelists/presenters.py
```diff
@@ -1,8 +1,72 @@
...
+    """
+    Data structure for a Definition to prepare for display
+
+    name: name of the definition
+    code: code of the definition
+    excluded_children: list of excluded children (created with this class too)
+    all_descendants: are all of this Definitions descendants included?
+
+    Between all_descendants and excluded_children we can cover the three state
+    situation a Definition can be in:
```
inglesp, line 18: You're using `Definition` but I think you mean `DefinitionRule`.

codelists/templatetags/definitions.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+from django.urls import reverse
+
+register = template.Library()
+
+
+@register.simple_tag
+def concept_url(coding_system_id, code):
+    return reverse(f"{coding_system_id}:concept", args=[code])
```
inglesp, line 9: I like this.

codelists/presenters.py
```diff
@@ -1,8 +1,72 @@
...
+
+        # filter the excluded rules down to children of the current Rule
+        excluding_rules = [
+            r for r in excluded if r.code in hierarchy.descendants(rule.code)
+        ]
+
+        # generate excluded children by recursing into this function
+        row.excluded_children = list(
+            _iter_rules(hierarchy, excluding_rules, name_for_rule)
+        )
```
inglesp, line 53: I found the recursion a bit hard to follow -- in part because on the first call, `rules` are included rules, and on the second call (ie the first recursion) they are excluded rules.

Given that on the first recursion, we only yield `Row(name=name_for_rule(rule), code=rule.code, all_descendants=False)`, I'd be tempted to avoid recursing altogether, and build up `row.excluded_children` directly.

inglesp (approved):
Good stuff.  There are some small naming changes, and a thought about how you could avoid some of the recursion.

codelists/presenters.py
ghickman: Great shout, this was very much a case of not seeing the wood for the trees.  Fixed in dad666a and much easier to read I think!

codelists/presenters.py
ghickman: Fixed 8491893

codelists/presenters.py
ghickman: Fixed in 8491893

codelists/presenters.py
ghickman: Fixed in 8491893

codelists/views.py
ghickman: Fixed in 8491893

codelists/presenters.py
ghickman: Makes sense!  I've added it in c44bbcb as an exception so I can express why it happened and for what code (which should help traceability when we hit this $later).

inglesp: Great.

PR #150, opened 2020-09-01
Build JS with Webpack

inglesp:
This is the noddiest thing that works.


package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "dependencies": {
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ebmdatalab/opencodelists.git"
+  },
+  "license": "MIT"
```
evansd, line 23: I'm pretty sure this is just an npm default, but I don't know if we've decided on the license yet so should probably leave this out.

ghickman (approved):

evansd (approved):
Looks sensible to me.

PR #153, opened 2020-09-02
Ignore Request Logging in Sentry

ghickman:
Sentry [defaults](https://docs.sentry.io/platforms/python/guides/logging/) to picking up logs of `INFO` and above.  django-structlog [logs exceptions](https://github.com/jrobichaud/django-structlog/blob/master/django_structlog/middlewares/request.py#L77-L98) which duplicates the events we already receive.  The events are also ungrouped since the event message contains too many unique parts, giving us a lot of noise.

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -14,6 +14,7 @@ import sys
 
 import sentry_sdk
 from sentry_sdk.integrations.django import DjangoIntegration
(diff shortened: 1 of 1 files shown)


evansd (approved):
:+1:

inglesp (approved):

PR #157, opened 2020-09-02
Convert coding system ID for URL reversing

### LINKED ISSUE #155: Definition URL Reversing Uses Incorrect Namespace
[Sentry](https://sentry.io/organizations/ebm-datalab/issues/1870830983)

Reversing URLs in the template needs to convert `ctv3tpp` -> `ctv3` to match the namespaces we're using for URLs.

ghickman:
Fixes #155


inglesp (approved):
This is a perfectly good sticking plaster.  There's similar code dotted about the codebase, so at some point we'll want to fix properly.  But :shipit:

PR #161, opened 2020-09-03
Add Creating Users

### LINKED ISSUE #90: Allow new users to be added without going via admin

ghickman:
This adds pages to allow the creation of a User without using the Django admin.

Flow TLDR:
1. Admin User creates new User
1. Admin User is shown an Activation URL
1. New User visits Activation URL
1. New User sets password, is logged in

---

There are 3 pages to this flow:
* User creation form
* User Activation URL page
* Set password form

### User Creation Form

This mirrors the admin form but drops a few fields we don't need (eg date joined).  It doesn't add a password so the new User set it at the end of the flow.  New Users are added as in-active so they're unable to login until they set a password.

### User Activation URL Page

This displays some basic User details (name & email) and their activation URL with the expectation that the a Admin who created the User will pass this link on to the New User.  This page is accessible as long as the User has `is_active = False` set.  This aims to help with lost activation links, but avoids abuse after the User has set a password.

### Set Password Form

This page lets a User set their password and flips their `is_active` bit to True.  The view checks both the URL is valid (using the signed username string) and the Username is for a valid User.  On a successful submission the User is logged in and redirected to the homepage ready to use the site.

The form driving this page mirrors Django's own `contrib.auth.forms.SetPasswordForm` but without the dependency on a User object.

Note: programmatically logging a User in with Django requires setting a backend on the User which has been hard-coded here since we're using the default backend.

---

### Some Notes on Testing
I've opted for `RequestFactory` where I think they make sense, which is vaguely defined as "produces less test set up".  However, you'll see quite a few uses of `Client`.  This is because Django's messages framework requires a Session (configured via middleware) and while one can set that up with a `RequestFactory` I didn't feel it added anything to the relevant tests.

Fixes #90

### DIFF
--- a/templates/opencodelists/user_activation_url.html
+++ b/templates/opencodelists/user_activation_url.html
@@ -12,7 +12,9 @@
 
     <p>Send the link below to the new user so they can set their password:</p>
 
-    <pre>{{ url }}</pre>
+    <pre class="mb-5">{{ url }}</pre>
+
+    <a class="btn btn-primary" href="{% url 'user-create' %}">Add another User</a>
 
   </div>
 </div>
--- a/templates/opencodelists/user_activation_url.html
+++ b/templates/opencodelists/user_activation_url.html
@@ -1,0 +1,21 @@
+{% extends 'base.html' %}
+
+{% block content %}
+
+<div class="row mt-5">
+  <div class="col-8 offset-1">
+
+    <div class="d-flex align-items-baseline">
+      <h2 class="mb-5 mr-3">{{ user.name }}</h2>
+      <small><i>{{ user.email }}</i></small>
+    </div>
+
+    <p>Send the link below to the new user so they can set their password:</p>
+
+    <pre>{{ url }}</pre>
+
+  </div>
+</div>
+<br />
+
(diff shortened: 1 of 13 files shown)


templates/opencodelists/user_activation_url.html
```diff
@@ -0,0 +1,21 @@
...
+
+    <div class="d-flex align-items-baseline">
+      <h2 class="mb-5 mr-3">{{ user.name }}</h2>
+      <small><i>{{ user.email }}</i></small>
+    </div>
+
+    <p>Send the link below to the new user so they can set their password:</p>
+
+    <pre>{{ url }}</pre>
+
```
inglesp, line 16: Maybe add a link for signed-in admins to add another user?

inglesp (approved):
The workflow is very clear, and all the code is exactly where I'd expect to find it, with good tests.

One tiny suggestion, but that doesn't need to hold this PR up.

templates/opencodelists/user_activation_url.html
ghickman: Ooh, I like that!

PR #163, opened 2020-09-03
Stop using History Substitution Table to generate SNOMED CT codelists

inglesp:
The mega commit message is mostly for future documentation -- if anything there doesn't make sense, let me know.


ghickman (approved):
Great commit message!

PR #168, opened 2020-09-04
Fix convert_codelist

ghickman:
This is a grab bag of updates to fix `convert_codelist`.

Codelists started tracking versions with the `CodelistVersion` model since we last generated SNOMED codelists.  This updates how we create a Codelist to work with the new model.  It doesn't make use of the `create_codelist` action so we can copy over the `version_str` attribute from the CTV3 Codelist instance.

### DIFF
--- a/codelists/management/commands/convert_codelist.py
+++ b/codelists/management/commands/convert_codelist.py
@@ -21,10 +21,12 @@ def convert_codelist(project, slug):
     codelist = Codelist.objects.get(project_id=project, slug=slug)
     assert codelist.coding_system_id in ["ctv3", "ctv3tpp"]
 
+    version = codelist.versions.first()
+
     buf = StringIO()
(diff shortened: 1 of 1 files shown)


inglesp (changes requested):
Can you do this via `codelist.actions.create_codelist()`?

ghickman:
I can, but as I mentioned in the description, that won't copy the `version_str` attribute over.  It's probably a low likelihood of them diverging, but given we're generating one codelist from the other I wanted to make sure they matched, what do you think?  Is it worth it for that or are we preparing to shoot ourselves in the foot by diverging from the actions pattern?

inglesp:
Sorry, missed that.  This is fine then (or you could make `version_str` an optional argument to `create_codelists()`, but that might be overkill.)

inglesp (approved):

PR #169, opened 2020-09-07
Split up Active and Inactive SNOMED Concepts

### LINKED ISSUE #164: Handle inactive SNOMED CT concepts in Definition tab
In SNOMED CT, concepts are removed from the hierarchy when they become inactive.  (There is an `active` field on the `Concept` model.)

Before 783bfa920187c300705381f74b476cee8232361c, concepts which are outside of a coding system's hierarchy were not included in a codelist's definition.  This was not intentional, but it led to some nice fairly small definitions.

Since 783bfa920187c300705381f74b476cee8232361c, concepts which are outside the hierarchy are included, but because they're outside the hierarchy, there is one rule for each inactive concept and this leads to a much larger definition.

So before, we had:

image

And now, we have:
image

So for SNOMED CT codelists, the definition should be built from only active concepts.  In the Definition tab, there should be a second section for inactive concepts.  My gut feeling is that a SNOMED CT definition shouldn't need to think about the inactive concepts.

ghickman:
This splits up the inactive and active SNOMED Concepts in the Definition tab of a Codelist:

For non-SNOMED definitions it keeps the Definition unchanged:

Fixes #164

### DIFF
--- a/codelists/views.py
+++ b/codelists/views.py
@@ -8,6 +8,7 @@ from django.utils.decorators import method_decorator
 from django.views.decorators.http import require_POST
 from django.views.generic import FormView, TemplateView
 
+from coding_systems.snomedct.models import Concept as SnomedConcept
 from opencodelists.models import Project
 
 from . import actions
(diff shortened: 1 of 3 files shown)


inglesp (approved):
Looks good.

PR #170, opened 2020-09-07
Lint & Format JavaScript

### LINKED ISSUE #162: Lint and test JS in CI

ghickman:
This sets up a linter (eslint) and a formatter (prettier) for our JavaScript code.  Both are configured in the Makefile and run as part of CI.

Note: eslint has hardcoded paths because the jQuery plugin creates false-negatives for our React modules.

Note: Prettier will reformat lots of things but it has trouble with Django templates (among  others) so I've disabled it for that.

Fixes #162


.github/workflows/main.yml
```diff
@@ -100,4 +100,44 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
```
inglesp, line 115: What's the relevance of this comment?

.github/workflows/main.yml
```diff
@@ -100,4 +100,44 @@ jobs:
...
+      - name: Install dependencies
+        run: npm install
+
+  test-js:
+    needs: [build-js]
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
```
inglesp, line 151: micronit: Stray blank line.

static/src/js/builder.jsx
```diff
@@ -8,13 +8,21 @@ class CodelistBuilder extends React.Component {
...
+        // build up an array of 2-item arrays to convert to an Object with
+        // fromEntries
+        const status = [`status-${code}`, props.codeToStatus[code]];
+        const expanded = [`expanded-${code}`, true];
+        return [status, expanded];
+      })
+    );
+
+    // Create initial state
+    this.state = { updateQueue: [], updating: false, ...codeStates };
```
inglesp, line 25: What was wrong with the old code?

static/src/js/builder.jsx
```diff
@@ -258,11 +258,12 @@ function Row(props) {
     toggleVisibility,
   } = props;
 
-  const style = { display: isVisible ? "flex" : "none" };
-  const className = row.pipes.length === 0 ? "mt-2" : "mt-0";
+  const visibility = isVisible ? "d-flex" : "d-none";
```
inglesp, line 261: TIL!

static/src/js/hierarchy.js
```diff
@@ -8,7 +8,7 @@ class Hierarchy {
   }
 
   getAncestors(node) {
-    if (!this.ancestorMap.hasOwnProperty(node)) {
+    if (!Object.prototype.hasOwnProperty.call(this.ancestorMap, node)) {
```
inglesp, line 11: Yuck!

From the docs:

> You may want to turn this rule off if your code only touches objects with hardcoded keys, and you will never use an object that shadows an `Object.prototype` method or which does not inherit from `Object.prototype`.

Can we turn it off?

inglesp (approved):
Great, thanks.  A few small notes.

.github/workflows/main.yml
ghickman: Copy/pasta from GitHub's docs… which is now wrong isn't it 😬

static/src/js/builder.jsx
ghickman: You should only set an initial State once per React component.  Previously we had 1 line creating State and two mutating it, eslint rightly called this out.  I tried to convert the lines to `setState()` calls but couldn't get them to work as expected (maybe because they're being called from the constructor??) so opted to switch to one initial creation of state.

Digging into setting state further, [this SO answer](https://stackoverflow.com/a/37760774/158304) (and the `setState()` docs link it contains) lay out the problem quite well I think.  I suspect using `setState()` in the constructor might be related to that method triggering a re-render of the component so it was possibly ignored/batched for later.

static/src/js/hierarchy.js
ghickman: Good find, strongly agree on the "yuck"!

ghickman: Removed in ff1a143

static/src/js/builder.jsx
ghickman: Relatedly (but not useful for this situation I think) there's the [classnames package](https://www.npmjs.com/package/classnames) which can be useful for building up `className` values dynamically.

static/src/js/builder.jsx
inglesp: I understand why fiddling with `state` directly outside of a constructor is a bad thing.  But I don't understand why doing so in the constructor is a problem.

What's the actual eslint error?

I'm all for public immutability, but code is often clearer with some local mutability, including in this case.

Could we do this instead?

```
    let state = { updateQueue: [], updating: false };

    this.codes.forEach((code) => {
      state["status-" + code] = props.codeToStatus[code];
      state["expanded-" + code] = true;
    });

    this.state = state;
```

ghickman: > What's the actual eslint error?

[react/no-direct-mutation-state](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-direct-mutation-state.md)

There's a discussion on the rule [here](https://github.com/yannickcr/eslint-plugin-react/issues/832) which might be of interest, it boils down to "mutation is bad", which I generally agree with (and appears to be the direction the JS community is taking to try and squash the languages various unexpected behaviours).

Given the example above though, that's a very clear block, so lets use that 🙂

ghickman: Fixed in ac6a40a

inglesp: :+1:

PR #179, opened 2020-09-09
Quieter Deploys

ghickman:
This sets both`git-fetch` and `npm install` to quiet mode so deploys don't output progress line-by-line.


inglesp (approved):
-0 As ebmbot's responsible for the deploys, we'll never see the output, and it might be useful... but have no strong opinions.

PR #184, opened 2020-09-10
Use fixture in js test

inglesp:


ghickman (approved):
I like that you're using the view to generate the fixture!

inglesp:
> I like that you're using the view to generate the fixture!

Yeah, I've got a bit of a thing about having realistic test data!

Issue #193, opened 2020-09-16
Single Code Hierarchies


ghickman:
TLDR: they don't currently work, but should they?

When we instantiate a Hierarchy with a single code we get an instance no edges and thus no nodes.  This only appears to affect two tests currently and I can't see a reason to build a single code codelist in normal usage.

Either way we need to fix the affected tests (which will be part of my work on #100), but should we also add a conditional to `Hierarchy.from_codes()` to guard against codes being <2?

inglesp:
Do you have an example?

ghickman:
I hit this with `1067731000000107` via the tests which use `CodelistFactory`.

You can recreate in a terminal with:
```
from codelists.hierarchy import Hierarchy
from coding_systems.snomedct import coding_system

h = Hierarchy.from_codes(coding_system, "1067731000000107")

h.nodes == set()  # True
```

inglesp:
The second argument to `Hierarchy.from_codes` should be a list of codes, not a single code.

ghickman:
Ah, that makes sense.

We should probably guard against the string version then?

inglesp:
Could do -- but where are we calling it with a string?

inglesp:
Ah right.  Strings being iterable is a nice convenience in Python, but it does cause this kind of footgunnery sometimes.  Maybe an `assert not isinstance(codes, str)` at the top of the function?

PR #194, opened 2020-09-16
Tree Tab 2.0

### LINKED ISSUE #100: Adapt tree view for SNOMED CT codelists

### LINKED ISSUE #165: Add collapse/expand to Tree tab
Should match behaviour of builder interface.  Depends on #100.

ghickman:
This rebuilds the Codelist version page's Tree tab with a modified version of the CodelistBuilder component.  It turned out that only `TermAndCode` was needed from CodelistBuilder so I've left the rest of the sub components where they are.

This also pulls a couple of `tree_tables`-related structures into the coding system specific modules (so we can support CTV3 and SNOMED).  Most notbably, concept codes grouped by type are now generated in the coding system (since CTV3 codes can have more than one group) and passed into `tree_tables`.

These changes brought up the single code Hierarchy edge case (#193), which in turn brought up the three view tests (checking redirects work as expected) that broke due to this work.  I've reduced their scope to only check they are redirecting to the correct location so they're less brittle in the future.

**Aplastic Anaemia (SNOMED)**

**Aplastic Anaemia**

Fixes #100 
Fixes #165

### DIFF
--- a/coding_systems/ctv3/coding_system.py
+++ b/coding_systems/ctv3/coding_system.py
@@ -1,3 +1,5 @@
+import collections
+
 from opencodelists.db_utils import query
 
 from .models import Concept, ConceptTermMapping
@@ -72,23 +74,32 @@ def code_to_term(codes, hierarchy):
     return {c.read_code: c.preferred_term() for c in concepts}
 
 
-def code_to_type(codes, hierarchy):
+def codes_by_type(codes, hierarchy):
     """
-    Create a mapping between a code and its Concept types
+    Group codes by their Concept "types"
 
     We treat the children of CTV3's root Concept as types.  However, CTV3
-    Concepts can be descended from more than one of these "types", so we use
-    the first type.
+    Concepts can be descended from more than one of these "types", so each
+    grouping of codes can have an overlap with other groupings.
     """
+
+    # Treat children of CTV3 root as types
     types = hierarchy.child_map[root]
     concepts = Concept.objects.filter(read_code__in=types).prefetch_related("terms")
     terms_by_type = {c.read_code: c.preferred_term() for c in concepts}
 
+    lookup = collections.defaultdict(list)
     for code in codes:
-        type_nodes = hierarchy.ancestors(code) & types
+        # get groups for this
+        groups = hierarchy.ancestors(code) & types
 
-        # TODO: can we pass in codes bucketed by type to tree_tables?
-        first = list(type_nodes)[0]
-        term = terms_by_type[first]
+        # Remove Concept "Additional values" when there is at least one other
+        # type to use.
+        if len(groups) > 1 and "X78tJ" in groups:
+            groups.remove("X78tJ")
 
-        yield code, term
+        for group_code in groups:
+            group_term = terms_by_type[group_code]
+            lookup[group_term].append(code)
+
+    return dict(lookup)
--- a/coding_systems/ctv3/coding_system.py
+++ b/coding_systems/ctv3/coding_system.py
@@ -1,6 +1,6 @@
 from opencodelists.db_utils import query
 
-from .models import ConceptTermMapping
+from .models import Concept, ConceptTermMapping
 
 name = "CTV3 (Read V3)"
 short_name = "CTV3"
@@ -63,3 +63,32 @@ def descendant_relationships(codes):
     """
 
     return query(sql, codes)
+
+
+def code_to_term(codes, hierarchy):
+    concepts = Concept.objects.filter(read_code__in=hierarchy.nodes).prefetch_related(
+        "terms"
+    )
+    return {c.read_code: c.preferred_term() for c in concepts}
+
+
+def code_to_type(codes, hierarchy):
+    """
+    Create a mapping between a code and its Concept types
+
+    We treat the children of CTV3's root Concept as types.  However, CTV3
+    Concepts can be descended from more than one of these "types", so we use
+    the first type.
+    """
+    types = hierarchy.child_map[root]
+    concepts = Concept.objects.filter(read_code__in=types).prefetch_related("terms")
+    terms_by_type = {c.read_code: c.preferred_term() for c in concepts}
+
+    for code in codes:
+        type_nodes = hierarchy.ancestors(code) & types
+
+        # TODO: can we pass in codes bucketed by type to tree_tables?
+        first = list(type_nodes)[0]
+        term = terms_by_type[first]
+
+        yield code, term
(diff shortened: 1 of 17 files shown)


inglesp (changes requested):
Lots of good stuff here, but the tree view is no longer showing whether a code is in the codelist or not!

coding_systems/ctv3/coding_system.py
```diff
@@ -63,3 +63,32 @@ def descendant_relationships(codes):
...
+    """
+    types = hierarchy.child_map[root]
+    concepts = Concept.objects.filter(read_code__in=types).prefetch_related("terms")
+    terms_by_type = {c.read_code: c.preferred_term() for c in concepts}
+
+    for code in codes:
+        type_nodes = hierarchy.ancestors(code) & types
+
+        # TODO: can we pass in codes bucketed by type to tree_tables?
+        first = list(type_nodes)[0]
```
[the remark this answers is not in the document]
ghickman: Quite possibly (sets are unordered IIRC?), but this went away in a future commit 🙂

static/src/js/tree/index.jsx
```diff
@@ -0,0 +1,20 @@
+"use strict";
+
+import ReactDOM from "react-dom";
+// Although React is not used in this module, if this line is removed, there is
+// a runtime error in codelistbuilder.jsx.
```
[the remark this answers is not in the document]
ghickman: In the `ReactDOM.render()` call we pass a JSX element (`<Tree>` here), that's syntactic sugar for `React.createElement()` so we have to have the `React` import.

ghickman:
Black, blue, and underlined codes reinstated.  I've also fixed a bug with alignment of the descendant toggles and pipes.

I removed the `TermAndCode` component in favour of separate components for each of its constituent parts.  Both the builder and tree want to display the items like this:

```
[ N pipes ] [ +/- /? ] [ Term ] [ Code ]
```

where the `+/-/?` section is Builder specific and the `Term` needs to be optionally linked.  So by breaking out these parts both the `CodelistBuilder` and `Tree` components can compose them as necesary.

### Codelist Builder


### Tree (CTV3)


### Tree (SNOMED)

templates/codelists/version.html
```diff
@@ -344,6 +344,8 @@ <h5 class="modal-title" id="publish-version-label">Publish Version</h5>
 </script>
 
 {{ child_map|json_script:"child-map" }}
+{{ code_to_url|json_script:"codeToURL" }}
+{{ codes_in_definition|json_script:"definitionCodes" }}
```
inglesp, line 348: Can this be `codesInDefinition`?

static/src/js/tree/Tree.jsx
```diff
@@ -30,6 +30,28 @@ class Tree extends React.Component {
...
   }
 
+  renderTerm(code, term) {
+    const element = <Term term={term} />;
+
+    const inDefinition = this.props.definitionCodes.includes(code);
+
+    // conditionally wrap Term in an <href> element if there's a URL for it. We
+    // only want to link to codes included in the codelist so some codes will
+    // stay unlinked.
```
inglesp, line 40: > We only want to link to codes included in the codelist so some codes will stay unlinked.

Why do we want this?

Looks like this might be a result of building up `code_to_url` in the view.  I don't think we need to do this, since the URL can be determined from the coding system ID (which the JS doesn't yet have) and the concept code (which it does have).  As such, I think it's worth hard coding the URL structure in the JS.

inglesp (approved):
Looks good.  A minor tweak suggested.

templates/codelists/version.html
ghickman: The JS variable?

static/src/js/tree/Tree.jsx
ghickman: > Why do we want this?

I might have misunderstood something here, I'd assumed we only wanted the included codes linked.  Should I have linked all the codes?

inglesp: That matches the existing behaviour.

templates/codelists/version.html
inglesp: Yeah -- that would make it match the pattern in all the other `json_script`s.

static/src/js/tree/Tree.jsx
ghickman: Fixed in 67abb8a

templates/codelists/version.html
ghickman: Fixed in 59f3365

PR #195, opened 2020-09-16
Tennis elbow fixture

ghickman:
⚠️ Blocked by #194 ⚠️

This extracts the use of the Tennis Elbow codelist into two fixtures:
* importing the data required for the codelist.
*  constructing a `Codelist` instance from that data.


inglesp (approved):
A nice net negative, thanks!

PR #198, opened 2020-09-17
Deploy as Ebmbot… again

ghickman:
This is take 3 at deploying as `ebmbot`. 😬

I've removed the `setup_sudo()` function as I completely misunderstood what it was doing, and it's not useful to opencodelists.

Instead, I've configured `ebmbot` to have password-less sudo access to `ln` and `systemctl` via `/etc/sudoers`, with this line:

    ebmbot	ALL=NOPASSWD:/bin/ln,/bin/systemctl

(TIL: sudo takes the last matching line when checking what permissions a user can have)

The deploy directory has been chown'd to `ebmbot:www-data` (from `www-data:www-data`).

The relevant lines in the fabfile have had `sudo ` prepended instead of calling `sudo()` because that throws up a password prompt regardless of sudoers config.

I've tested this setup locally (my SSH key is authorized for `ebmbot` currently, but I'll remove that once this is merged) to confirm it works.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -38,7 +38,7 @@ def install_python_requirements():
 
 
 def chown_everything():
-    run(f"chown -R www-data:www-data {env.path}")
+    run(f"chown -R {env.user}:www-data {env.path}")
 
 
 def run_migrations():
--- a/fabfile.py
+++ b/fabfile.py
@@ -1,4 +1,4 @@
-from fabric.api import abort, env, prefix, run, sudo, task
+from fabric.api import abort, env, prefix, run, task
 from fabric.context_managers import cd, shell_env
 from fabric.contrib.files import exists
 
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -63,7 +38,7 @@ def install_python_requirements():
 
 
 def chown_everything():
-    run(f"chown -R www-data:www-data {env.path}")
+    run(f"chown -R {env.user}:www-data {env.path}")
```
inglesp, line 41: Is this chowning necessary now?  It was before, because files were being created by the root user.

inglesp (approved):
Probably OK?  I'm not very good at this stuff.

sebbacon:
Basically if it works, I'm happy: my top priority for Simon & Tom is probably picking one way of deploying stuff and implementing it, so we'll almost certainly be chucking all this away anyway.

fabfile.py
ghickman: Yes, something in a previous step was chowning as `ebmbot:ebmbot` and I figured I'd change as few things as possible!

PR #203, opened 2020-09-18
Download Codelist from Builder

### LINKED ISSUE #200: Add download button to builder
As a stopgap, the builder should have a "Download as CSV" button.  It should be greyed out (with a suitable tooltip) if any concepts are unresolved or in conflict.

I'd suggest that it's a link to `https://codelists.opensafely.org/builder/<username>/<slug>/download.csv`, which is an endpoint handled by a Django view.

The downloaded CSV should have two columns: `id` and `term`.  The filename should be something like `<username>-<slug>-<timestamp>.csv`.

ghickman:
This adds a download button to the Codelist builder page which builds a CSV of the codelist (in code,term format).

The button is disabled on codelists with unresolved or conflicted concepts.

Fixes #200

### DIFF
--- a/builder/views.py
+++ b/builder/views.py
@@ -33,7 +33,7 @@ def download(request, username, codelist_slug):
         coding_system = CODING_SYSTEMS["snomedct"]
 
     # get terms for codes
-    code_to_term = coding_system.get_code_to_term(codes)
+    code_to_term = coding_system.lookup_names(codes)
 
     timestamp = timezone.now().strftime("%Y-%m-%dT%H-%M-%S")
     filename = f"{username}-{codelist_slug}-{timestamp}.csv"
--- a/builder/views.py
+++ b/builder/views.py
@@ -24,7 +24,9 @@ def download(request, username, codelist_slug):
     codelist = get_object_or_404(DraftCodelist, owner=username, slug=codelist_slug)
 
     # get codes
-    codes = list(codelist.codes.values_list("code", flat=True))
+    codes = list(
+        codelist.codes.filter(status__contains="+").values_list("code", flat=True)
+    )
 
     # get coding_system module (and prepare for CTV3 integration)
     if codelist.coding_system_id in ["ctv3", "ctv3tpp"]:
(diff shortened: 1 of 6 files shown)


builder/views.py
```diff
@@ -17,6 +20,35 @@
 from .models import DraftCodelist
 
 
+def download(request, username, codelist_slug):
+    codelist = get_object_or_404(DraftCodelist, owner=username, slug=codelist_slug)
+
+    # get codes
+    codes = list(codelist.codes.values_list("code", flat=True))
```
inglesp, line 27: This will pick up all codes, whether or not they're included or excluded.

coding_systems/snomedct/coding_system.py
```diff
@@ -85,3 +85,11 @@ def descendant_relationships(codes):
     """
 
     return query(sql, codes)
+
+
+def get_code_to_term(codes):
```
inglesp, line 90: What's the difference between this and `lookup_names()` above?

inglesp (changes requested):

builder/views.py
ghickman: Can I rely on the status strings to filter these?  eg `.filter(status__contains="+")`?  or do I need to build a hierarchy from the codes?

coding_systems/snomedct/coding_system.py
ghickman: "programmer error" 😬  fixed in c5c1140

builder/views.py
inglesp: > Can I rely on the status strings to filter these? eg .filter(status__contains="+")?

Yes.

ghickman: Brilliant, fixed in 6012368 then 🙂

ghickman:
@inglesp – do you want to check this one locally to confirm it exports the expected codelist? (I'm assuming you have more test data than I do!)

inglesp:
> @inglesp – do you want to check this one locally to confirm it exports the expected codelist?

Have done this.

How much work would it be to add a test?

PR #204, opened 2020-09-18
Convert SNOMED codes -> CTV3 codes

### LINKED ISSUE #201: Implement mappings.ctv3sctmap2.mappers.snomedct_to_ctv3()
This should be a function that takes a list of SNOMED CT ids, and returns a list of corresponding CTV3 ids, using the active and assured rows from `mappings.ctv3sctmap2.models.Mapping`.  (These come from `Mapping Tables/Updated/Clinically Assured/ctv3sctmap2_uk_XXX.txt` in [this TRUD subpack](https://isd.digital.nhs.uk/trud3/user/authenticated/group/0/pack/8/subpack/9/releases).)

We should check with Chris/Jonny from TPP whether they do anything fancier.

ghickman:
This adds a function to convert SNOMED CT concept codes to CTV3 concept codes using the assured mappings.

Fixes #201

### DIFF
--- a/mappings/ctv3sctmap2/mappers.py
+++ b/mappings/ctv3sctmap2/mappers.py
@@ -88,3 +88,14 @@ def ctv3_to_snomedct(ctv3_ids):
             )
 
     return records
+
+
+def snomedct_to_ctv3(snomedct_ids):
+    """Convert SNOMED CT Concept codes to CTV3 Concepts IDs."""
+    return set(
+        CTV3Concept.objects.filter(
+            snomedct_mappings__sct_concept_id__in=snomedct_ids,
+            snomedct_mappings__map_status=True,
+            snomedct_mappings__is_assured=True,
+        ).values_list("pk", flat=True)
+    )
--- a/mappings/ctv3sctmap2/mappers.py
+++ b/mappings/ctv3sctmap2/mappers.py
@@ -1,7 +1,8 @@
 from collections import defaultdict
 
+from coding_systems.ctv3 import coding_system as ctv3
 from coding_systems.ctv3.models import Concept as CTV3Concept
-from coding_systems.snomedct.coding_system import descendant_relationships, lookup_names
+from coding_systems.snomedct import coding_system as snomedct
 from coding_systems.snomedct.models import Concept as SCTConcept
 from coding_systems.snomedct.models import QueryTableRecord
 
@@ -39,7 +40,7 @@ def ctv3_to_snomedct(ctv3_ids):
     # mappings from leaf CTV3 concepts.
     descendant_ids = {
         child_id
-        for (parent_id, child_id) in descendant_relationships(
+        for (parent_id, child_id) in snomedct.descendant_relationships(
             list(ctv3_leaf_snomedct_ids)
         )
     } - snomedct_ids
@@ -64,7 +65,7 @@ def ctv3_to_snomedct(ctv3_ids):
     ):
         snomedct_id_to_ctv3_id[mapping.sct_concept_id].append(mapping.ctv3_concept_id)
 
-    snomedct_id_to_name = lookup_names(all_ids)
+    snomedct_id_to_name = snomedct.lookup_names(all_ids)
     snomedct_id_to_active = {
         concept.id: concept.active
         for concept in SCTConcept.objects.filter(id__in=all_ids)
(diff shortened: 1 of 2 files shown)


inglesp (changes requested):
This logic looks sensible.  Could we make it so that the interface matches `ctv3_to_snomedct` -- ie so that it returns a list of dicts?

ghickman:
Sure thing.

Should we be mirroring (reversing) other parts of the CTV3 -> Snomed function too? eg:
* leaf IDs
* descendant IDs
* inactive IDs

inglesp:
> Should we be mirroring (reversing) other parts of the CTV3 -> Snomed function too?

Don't think so.

leaves/descendants: Almost every CTV3 concept is in SNOMED CT (that's what the "CT" refers to), so I don't expect we'll pick up any extra concepts by looking at the descendants of CTV3 concepts which are mapped to from the leaf SNOMED CT concepts.

inactives: There isn't as far as I know the equivalent of a query table for CTV3, so I don't think there's anything we can do.

mappings/ctv3sctmap2/mappers.py
```diff
@@ -1,7 +1,8 @@
 from collections import defaultdict
 
+from coding_systems.ctv3 import coding_system as ctv3
 from coding_systems.ctv3.models import Concept as CTV3Concept
-from coding_systems.snomedct.coding_system import descendant_relationships, lookup_names
+from coding_systems.snomedct import coding_system as snomed
```
inglesp, line 5: For consistency, can we import this as `snomedct`?

mappings/ctv3sctmap2/tests/test_mappers.py
```diff
@@ -0,0 +1,77 @@
...
+        module=snomed_concept,
+        type=snomed_concept,
+    )
+
+    ctv3_concept = CTV3Concept(
+        read_code="J650.",
+    )
+    ctv3_concept.another_concept = ctv3_concept
+    ctv3_concept.save()
+    # another_concept_id = factory.SelfAttribute("read_code")
```
inglesp, line 46: Stray comment?

mappings/ctv3sctmap2/mappers.py
```diff
@@ -88,3 +89,39 @@ def ctv3_to_snomedct(ctv3_ids):
             )
 
     return records
+
+
+def snomedct_to_ctv3(snomedct_ids):
+    """Convert SNOMED CT Concept codes to CTV3 Concepts IDs."""
+    ctv3_ids = set(
+        CTV3Concept.objects.filter(
+            snomedct_mappings__sct_concept__active=True,
```
inglesp, line 98: Why are we filtering out inactive SNOMED concepts here?

mappings/ctv3sctmap2/mappers.py
```diff
@@ -88,3 +89,39 @@ def ctv3_to_snomedct(ctv3_ids):
...
+
+    code_to_name = ctv3.lookup_names(ctv3_ids)
+
+    records = []
+    for ctv3_id in ctv3_ids:
+        records.append(
+            {
+                "id": ctv3_id,
+                "name": code_to_name[ctv3_id],
+                "active": True,
```
inglesp, line 121: CTV3 concepts can be in one of [four statuses](https://github.com/opensafely/opencodelists/blob/master/coding_systems/ctv3/models.py#L10-L13).  Suggest dropping the "active" key for now.

inglesp (changes requested):

mappings/ctv3sctmap2/mappers.py
ghickman: Shoot, sorry, yes - thought I'd fixed all of these!

mappings/ctv3sctmap2/mappers.py
ghickman: The reverse of this function only looks for active Snomed concepts, should this function ignore that state?

mappings/ctv3sctmap2/mappers.py
ghickman: Fixed in 8df5628

mappings/ctv3sctmap2/tests/test_mappers.py
ghickman: Fixed in 8df5628

mappings/ctv3sctmap2/mappers.py
ghickman: Fixed in 8df5628

mappings/ctv3sctmap2/mappers.py
inglesp: When going from SNOMED to CTV3, it makes sense not to exclude inactive SNOMED concepts, since the SNOMED concepts are provided by the caller of the function.

When going from CTV3 to SNOMED, I'm not sure why we've previously ignored inactive SNOMED concepts that are in the mapping.  I'll do some digging through emails/Slack, and make sure this is written up somewhere.

ghickman: Ah, of course, good point!  Fixed in 3833abc

inglesp: Turns out this is moot, because no inactive SNOMED concepts are included in the mappings that are both assured and have `map_status=True`

```
db.sqlite3> select active, count(*) from snomedct_concept where id in (select distinct sct_concept_id from ctv3sctmap2_mapping where is_assured and map_status) group by active;                           
+--------+----------+
| active | count(*) |
+--------+----------+
| 1      | 150725   |
+--------+----------+
```

Further thought required (outside the scope of this PR) about what, if anything, we can do about mapping from inactive SNOMED concepts.

inglesp (approved):

inglesp:
Merging now because I want to start using it

PR #212, opened 2020-09-24
Add page for converting codelists between coding systems 

inglesp:
This is about the simplest thing that's useful, and only allows for conversions from SNOMED CT to CTV3, which is most urgent.

There are no tests yet -- I'll add some once we've worked out whether this is something people actually need.

### DIFF
--- a/conversions/views.py
+++ b/conversions/views.py
@@ -1,0 +1,83 @@
+import csv
+import os
+from io import StringIO
+
+from django.http import HttpResponse
+from django.shortcuts import render
+
+from codelists.coding_systems import CODING_SYSTEMS
+from mappings.ctv3sctmap2.mappers import get_mappings
+
+from .forms import ConvertForm
+
+
+def convert(request):
+    if request.POST:
+        form = ConvertForm(request.POST, request.FILES)
+        if form.is_valid():
+            from_coding_system_id = form.cleaned_data["from_coding_system_id"]
+            to_coding_system_id = form.cleaned_data["to_coding_system_id"]
+            assert from_coding_system_id == "snomedct"
+            assert to_coding_system_id == "ctv3"
+            from_coding_system = CODING_SYSTEMS[from_coding_system_id]
+            to_coding_system = CODING_SYSTEMS[to_coding_system_id]
+
+            base_filename, _ = os.path.splitext(form.cleaned_data["csv_data"].name)
+            csv_data = form.cleaned_data["csv_data"].read().decode("utf-8-sig")
+            rows = list(csv.reader(StringIO(csv_data)))
+            snomedct_ids = [row[0] for row in rows[1:]]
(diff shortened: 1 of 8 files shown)


inglesp:
Merging now because site is broken without 5a86e8db3cad1e88c59e775ebbee1194d5ca0b9a.  A quick review would still be appreciated @ghickman!

conversions/views.py
```diff
@@ -0,0 +1,83 @@
...
+            response = HttpResponse(content_type="text/csv")
+            response["Content-Disposition"] = f'attachment; filename="{filename}"'
+            writer = csv.writer(response)
+            writer.writerow(headers)
+            writer.writerows(sorted(data))
+
+            return response
+
+    else:
+        form = ConvertForm()
```
ghickman, line 80: This is so far away from the opening `if` that I can't easily see them both on a large monitor!  This would be easier to follow if the else was a shortcut returning `render` et al.

conversions/views.py
```diff
@@ -0,0 +1,83 @@
...
+from codelists.coding_systems import CODING_SYSTEMS
+from mappings.ctv3sctmap2.mappers import get_mappings
+
+from .forms import ConvertForm
+
+
+def convert(request):
+    if request.POST:
+        form = ConvertForm(request.POST, request.FILES)
+        if form.is_valid():
```
ghickman, line 17: There's no `else` to this check so the invalid form isn't handled, as below that would probably be better served as a shortcut return.

conversions/views.py
```diff
@@ -0,0 +1,83 @@
...
+            writer.writerow(headers)
+            writer.writerows(sorted(data))
+
+            return response
+
+    else:
+        form = ConvertForm()
+
+    ctx = {"form": form}
+    return render(request, "conversions/convert.html", ctx)
```
ghickman, line 83: Returning a `TemplateResponse` here would provide some more context (get it 😉) for things handling the response.

Some useful context in the [Views the Right Way article](https://spookylukey.github.io/django-views-the-right-way/the-pattern.html) (search for `The issue with just using render is that you get a plain HttpResponse object back`, I can't directly link to the paragraph) which lays out why it's useful to use that.  It uses the same constructor as `render` too which is a nice bonus.

ghickman (commented):
Looks good!

Consider swapping the view out for a FormView subclass though, all your logic would go in `form_valid` and you'd gain support for the invalid form case without adding anything while keeping all the other bits the same (templated response, GET/POST form rendering, etc).

conversions/views.py
inglesp: Get a bigger screen?

conversions/views.py
inglesp: Thanks for the reminder to read this properly.

PR #217, opened 2020-09-28
Import TPP's CTV3 codes

inglesp:
TPP have an extension to CTV3 ("Y codes").  They have shared details of these codes, along with the subset of CTV3 that is actually used.

We'll use the TPP concepts in eg the builder and to validate uploaded codelists.  And we'll keep the raw data around too, to allow for:

* searching against synonyms
* showing concept statuses


inglesp:
Have converted to draft, because this will have broken `ancestor_relationships` and `descendant_relationships`.

ghickman (commented):
👍

Great to see CTV3 being made to the builder!

PR #221, opened 2020-09-30
JS housekeeping

inglesp:
@ghickman do you have any better ideas for fixing 7d98862c0dfa2e203a51c443d6148f26bcd51772?

### DIFF
--- a/static/src/js/builder/codelistbuilder.jsx
+++ b/static/src/js/builder/codelistbuilder.jsx
@@ -42,6 +42,15 @@ class CodelistBuilder extends React.Component {
     this.getIsVisible = this.getIsVisible.bind(this);
   }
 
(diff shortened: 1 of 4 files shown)


ghickman:
@inglesp – nothing immediately springs to mind.  Would mounting the componet in a single test to exercise that code path solve the warning?  Or does it fire for every test which touches it?

Tracking mounted state of the component in an instance variable feels like the wrong approach, but given I don't have a better idea, this is a pragmatic solution until we find something better 🙂

ghickman (approved):

inglesp:
> until we find something better

This'll probably be to have a HOC to only manage state changes, and using a function with no hooks for rendering.  But those are words I've only just learnt, so it'll be a while before I try to implement this!

PR #223, opened 2020-09-30
Allow codelists of any coding system to be downloaded

inglesp:


ghickman (approved):
Great to see how the coding_system pattern is working out here and bonus more deletions than additions 😀

PR #248, opened 2020-10-08
Tweak how duplicate codelist names are handled

inglesp:
This is in advance of moving the project/name/slug attributes off
Codelist.  Once this has happened, we will no longer be able to use a
ModelForm.


inglesp:
@ghickman this is part of a larger piece of work.  I wanted to check this commit with you first, to make sure that I'm not missing anything in the current implementation.

ghickman (approved):
👍

`fields_for_model` is useful!

PR #260, opened 2020-10-14
Rewrite the tree-rendering code for tree tab

inglesp:


opencodelists/settings.py
```diff
@@ -85,6 +85,11 @@
     "django_structlog.middlewares.RequestMiddleware",
 ]
 
+if not os.environ.get("DDT_ENABLED"):
+    # Django Debug Toolbar adds 77s to the load time of a large codelist!
+    INSTALLED_APPS.remove("debug_toolbar")
+    MIDDLEWARE.remove("debug_toolbar.middleware.DebugToolbarMiddleware")
+
```
ghickman, line 92: Did you test which panel was adding the render penalty? (Request is a common candidate for rendering slowdowns).

Turning off the problematic panel (but not disabling it) means you can still enable it from the UI, and the preference is kept in a cookie to persist reloads.

static/src/js/tree-tables.jsx
```diff
@@ -0,0 +1,196 @@
...
+      treeTables={treeTables}
+      codeToStatus={codeToStatus}
+      codeToTerm={codeToTerm}
+      visiblePaths={visiblePaths}
+    />,
+    document.querySelector("#codelist-tree")
+  );
+}
+
+setUpAndRender();
```
ghickman, line 196: 👍

ghickman (approved):

PR #279, opened 2020-10-19
Fix Edit Codelist Form

ghickman:
This fixes the Edit Codelist Form being blank when you edit a Codelist.  It instantiates the CodelistUpdateForm with the necessary values from the existing `Codelist`.

I've removed the `get_form` method now that get/post have diverged enough to remove the abstractions usefulness.

### DIFF
--- a/codelists/tests/test_views.py
+++ b/codelists/tests/test_views.py
@@ -226,7 +226,30 @@ def test_codelistupdate_invalid_post(rf):
     assert response.context_data["signoff_formset"].errors
 
 
-def test_codelistupdate_success(rf):
+def test_codelistupdate_success_get(rf):
+    codelist = CodelistFactory()
+    SignOffFactory(codelist=codelist)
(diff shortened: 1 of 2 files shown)


inglesp (approved):
Did this ever work, or did I break it?

It would be nice to have a test, but it's the kind of thing that won't break silently, so don't worry if testing setup is fiddly.

ghickman:
It was the move from `ModelForm` to `Form` which did it: https://github.com/opensafely/opencodelists/pull/248/files#diff-dbd9d3a4ee9a3a9300bfbf824321619f90d5623bcc004d29dad66da9fc3751c5L231-R233

Super easy mistake though, when I first saw the bug I even naively thought "oh I've just not passed in an instance", but Form fun lead to a bigger change!

ghickman:
@inglesp – test added!

inglesp:
Great, thanks.

PR #292, opened 2020-10-23
More frontend work

inglesp:
This batches up several changes, and should be reviewed commit-by-commit.

The main change is that `CodelistBuilder` now uses the new `TreeTables` component, which means that large codelists can be rendered by React, and are more tractable for users.

This change also allows a couple of other changes in behaviour, including showing slightly more information in the modal popups.

There'll be another PR to delete a bunch of unused code.


static/src/js/tree/index.jsx
```diff
@@ -1,26 +1,37 @@
...
   readValueFromPage("parent-map"),
   readValueFromPage("child-map")
 );
 
+const treeTables = readValueFromPage("tree-tables");
+const codeToStatus = readValueFromPage("code-to-status");
+const codeToTerm = readValueFromPage("code-to-term");
+
+const ancestorCodes = treeTables
+  .map(([_, ancestorCodes]) => ancestorCodes) // eslint-disable-line no-unused-vars
```
ghickman, line 20: Does JS not have a way to ignore variables when looping?? (arg!)

static/src/js/builder/index.jsx
```diff
@@ -21,7 +21,8 @@ const ancestorCodes = treeTables
   .flat();
 const visiblePaths = hierarchy.initiallyVisiblePaths(
   ancestorCodes,
-  codeToStatus
+  codeToStatus,
+  1
```
ghickman, line 25: Is there a way to make this a kwarg/named argument in JS?  (without going the full on Object route at least)

ghickman (approved):

static/src/js/tree/index.jsx
inglesp: Turns out there is!  Fixed in 2405a22f.

static/src/js/builder/index.jsx
inglesp: Not according to https://stackoverflow.com/questions/11796093/is-there-a-way-to-provide-named-parameters-in-a-function-call-in-javascript.

Agree, it's unsightly.

PR #294, opened 2020-10-23
Fix Deleting Related Forms

### LINKED ISSUE #244: Removing references bug
I've just come across a bug with references that means I cannot progress. I have added a reference to this codelist https://codelists.opensafely.org/codelist/opensafely/ named  `Diabetes Ketoacidosis - CTV3 (DKA - unspecific)` (i can't seen to generate a unique url before submission)

however I no longer want it and have removed it via the x button. It is clearly removed. Once I press submit it reappears :(

image

ghickman:
This turns on deleting related existing forms in the Reference and SignOff FormSets and modifies the template to retain the choices made by a User.

Fixes #244

### DIFF
--- a/codelists/forms.py
+++ b/codelists/forms.py
@@ -9,6 +9,20 @@ from django.forms.models import fields_for_model
 from .models import Codelist, CodelistVersion, Reference, SignOff
 
 
+def data_without_delete(cleaned_data):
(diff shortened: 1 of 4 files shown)


inglesp (approved):
WHAT A FAFF

PR #305, opened 2020-11-02
Add cronjob to back up database every night

inglesp:
We back up data from just the apps that have user-created data.

### DIFF
--- a/deploy/bin/backup.sh
+++ b/deploy/bin/backup.sh
@@ -1,13 +1,13 @@
 #!/bin/bash
 
-REPO_ROOT=$(dirname $(dirname $(dirname $0)))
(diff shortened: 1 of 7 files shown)


deploy/bin/backup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
ghickman, line 2: Would be best to add options thingy:

`set -euo pipefail`

ghickman (changes requested):
Looks good!

Couple of quick things because shell scripts are always "great" to deal with…

Could you set up shellcheck for CI and pre-commit too?  If you want to add that in another PR, totally fine!

ghickman (approved):

PR #306, opened 2020-11-02
Bonfire of unused code

inglesp:
> Could you TL;DR for posterity, and because my brain has GC'd where OC is at, why all of this can go?

* `tree_utils.py` can go because it is unused
* The tree walking code in `hierarchy.py` can go because it is done (and done better) in JS
* The deleted JS can go because instead of having a bunch of common components (eg `TermAndCode`) that are used by both the tree tab and the builder, there is now a single common component (`TreeTables`) that is used by both.


ghickman (approved):
That deleted count 👌 

Could you TL;DR for posterity, and because my brain has GC'd where OC is at, why all of this can go?

inglesp:
Thanks.  Have updated the PR description.

PR #331, opened 2020-11-10
Make builder.actions.create_search more efficient

### LINKED ISSUE #330: Gunicorn timeout after searching for term with lots of results
When a user clicks "Search" in the builder, a form is submitted via `POST` to `/builder/[username]/[codlelist_slug]/search/`.  This is handled by [builder.views.new_search](https://github.com/opensafely/opencodelists/blob/bd8bbc7db6fe34055b7f41f3eaf1c03ae6ca74e6/builder/views.py#L198-L209), which calls [builder.actions.create_search](https://github.com/opensafely/opencodelists/blob/master/builder/actions.py#L23-L31).  For search terms with lots of results, this action is slow, and gunicorn times out.

Depending on the browser, a user sees a blank page or a 405 (possibly because Cloudflare sends a redirect to `/builder/[username]/[codlelist_slug]/search/`, which only allows `POST`s).

The fix is one of:

* increase the gunicorn timeout
* make `builder.actions.create_search` faster
* not create a search if there are too many results

inglesp:
The calls to get_or_create were inefficient.  With this change, creating
a search for "tia" takes ~5s, down from >30s.

This addition highlights a poor terminology choice on my part.
builder.models.Code should be called something like
builder.models.CodeObj, to distinguish it from a concept code.

Fixes #330.

### DIFF
--- a/builder/actions.py
+++ b/builder/actions.py
@@ -7,6 +7,8 @@ from django.utils.text import slugify
 
 from codelists.hierarchy import Hierarchy
 
+from .models import Code, SearchResult
+
 logger = structlog.get_logger()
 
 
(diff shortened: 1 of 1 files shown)


ghickman (approved):
Nice!

PR #333, opened 2020-11-11
Add machinery for converting codelists to become editable

inglesp:
This is the next baby step in making codelists editable.

For a codelist to be editable, we need to store the codes and the rules in the database, instead of storing the original CSV data.  I'm calling these "new-style" codelists.

This PR:

* adds the models/migrations for this
* adds an action that converts a codelist
* amends the `.codes` and `.table` properties accordingly


codelists/models.py
```diff
@@ -99,24 +99,36 @@ def coding_system(self):
...
+                    if self.codelist.slug == "ethnicity":
+                        ix = 1
+                    else:
+                        ix = 0
+
+                return tuple(sorted({row[ix] for row in rows}))
 
-            return tuple(sorted({row[ix] for row in rows}))
+        else:
+            return tuple(sorted(self.code_objs.values_list("code", flat=True)))
```
ghickman, line 131: You could flip this to short-circuit on the negative of the conditional, which I think would be clearer, and save some indentation on the positive part of the condition.

ghickman (approved):
👍

codelists/models.py
inglesp: Thanks.  Have addressed in 0bc934f1a2b3a86e8c592f0b544792172c442c08.

PR #337, opened 2020-11-14
Simplify organisation data model (ii)

inglesp:
This PR allows users to be members of multiple organisations, through the `Membership` model.  Additionally, a user can be an admin for an organisation.

Creating/updating a codelist or version is now restricted to members of the organisation that owns the codelist.

The big implementation change here is rewriting several CBVs as FBVs.  The reason I've done this now is because I thought it would make writing the new `load_xxx` decorators easier.  However, I hadn't properly understood `django.utils.decorators.method_decorator` at the start of the process.  (If I wasn't working on a Saturday I'd have discussed this first!)

I hope I've managed to recreate some of the benefits of CBVs in my implementation -- specifically, separating the handling of GETs and POSTs, and giving things a recognisable structure.  Splitting the views into their own modules feels like a benefit, since views.py was getting quite large.


opencodelists/tests/factories.py
```diff
@@ -14,7 +14,14 @@ class Meta:
     @classmethod
     def _create(cls, model_class, *args, **kwargs):
         """Override the default ``_create`` with our actions."""
-        return actions.create_organisation(*args, **kwargs)
+        organisation = actions.create_organisation(*args, **kwargs)
+        organisation.regular_user = UserFactory()
```
ghickman, line 18: I'm not sure I understand this line.  There's no `regular_user` field on the `Organisation` model, is this to make it easy to grab an authorised User from the Organisation instances inside tests?

ghickman (approved):
I don't love the CBV/FBV hybrid but the organisation is solid, it works, and if it improves readability for _you_ as the primary developer of the project then that's a net win!

---

The abstraction of the auth and permission assertions feels like a code smell to me.  I see how you got there, but as someone who doesn't know the project very well, coming to a test which looks like this:

```
def test_something(rf):
    some_model = SomeModelFactory()
    assert_a_thing(rf, some_view_func, some_model)
```

It's very hard to reason why a test has failed without digging the dynamic parts of those abstracted assert_* functions.

All that being said, I don't want to gate this PR on it.  Can you try them out for a while and see if it trips you up?  "in anger" etc etc

---

The `load_X` decorators are great, they tie into exactly how we're getting objects in views 👌  `require_permission` fits in really nicely too, makes the views so readable!

opencodelists/tests/factories.py
```diff
@@ -23,13 +23,3 @@ class Meta:
...
-
-    @classmethod
-    def _create(cls, model_class, *args, **kwargs):
-        """Override the default ``_create`` with our actions."""
-        o = Organisation.objects.first() or OrganisationFactory()
-        u = User.objects.create(**kwargs)
-        actions.add_user_to_organisation(
-            user=u, organisation=o, date_joined="2020-11-12"
-        )
-        return u
```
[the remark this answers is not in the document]
inglesp: There's some work to do to make a nice set of fixtures.

opencodelists/migrations/0005_auto_20201112_1231.py
```diff
@@ -0,0 +1,30 @@
...
+        membership.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("opencodelists", "0004_auto_20201112_1200"),
+    ]
+
+    operations = [migrations.RunPython(add_opensafely_users_to_organisation)]
```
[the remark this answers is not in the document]
inglesp: We've got backups, so if we need to reverse the migration (unlikely) we can do so.

opencodelists/tests/factories.py
inglesp: Comment added in 76d9db284e32c47345a2cb287b6a8365f67ad9d0.

inglesp:
> It's very hard to reason why a test has failed without digging the dynamic parts of those abstracted `assert_*` functions.

I've added assert messages in 90cbf85c9f9fb88f9bf7fa9a039d375453d4ff2e to provide a little more information.

conftest.py
```diff
@@ -7,6 +7,8 @@
 from codelists import actions
 from codelists.tests.factories import CodelistFactory
 
+pytest.register_assert_rewrite("codelists.tests.views.assertions")
```
ghickman, line 10: whoa, TIL!

PR #339, opened 2020-11-16
Allow for non search codes in builder

inglesp:
This is next baby step towards allowing uploaded codelists to be edited.


ghickman (approved):
+1 for all these baby steps 👍

PR #341, opened 2020-11-17
Support user-owned codelists

inglesp:


codelists/models.py
```diff
@@ -36,6 +36,15 @@ class Codelist(models.Model):
...
         unique_together = [("organisation", "name", "slug"), ("user", "name", "slug")]
+        constraints = [
+            models.CheckConstraint(
+                name="%(app_label)s_%(class)s_organisation_xor_user",
+                check=(
+                    models.Q(organisation_id__isnull=False, user_id__isnull=True)
+                    | models.Q(user_id__isnull=False, organisation_id__isnull=True)
+                ),
+            )
+        ]
```
ghickman, line 47: 🤩

evansd: This is so nice. I'd forgotten that the ORM had grown support for specifying constraints.

ghickman:
Rather than leave a Review where I have to choose YAY/NAY I want to discussion some of the implementation:

There's logic in various places around "is this owned by an Org or a User" [or is this just fantasy 🎶] and I have a counter proposal for avoiding that:
* all codelists are owned by an Org
* create an Org for every user upon sign up using their name as the Org name (eg there's an OpenSAFELY Org and a George Hickman Org).

This has the benefit that, internally, ownership is reduced to The One Owner Model.  Any Org/User logic is coralled to the Org itself (I've done this with an extra, nullable, FK between Org and User before, to denote "User Orgs").  To a User they still see "I own this Codelist" or "OS owns this Codelist" but we simplify the inner workings of that.

inglesp:
This might be a really neat solution, I'm going to have a think though the implications.

inglesp:
> There's logic in various places around "is this owned by an Org or a User"

I share this concern, but rereading the code the situation isn't as bad as I first thought.

At the moment, the only places we have this logic relate to URLs, because the only difference between org and user codelists is that we're serving them from different places.

The relevant parts of the code are:

* In [`Codelist.url_kwargs()`](https://github.com/opensafely/opencodelists/blob/decdb1c53ca2ff9c8df70793c8347920ed5852c7/codelists/models.py#L84-L95), where we have to check whether the codelist is for an org or a user in order to decide what kwargs to pass to `reverse()`
* In [decorators.py](https://github.com/opensafely/opencodelists/blob/decdb1c53ca2ff9c8df70793c8347920ed5852c7/codelists/views/decorators.py), where we have to look at the parameters in a URL to work out how to pull out a `Codelist`/`CodelistVersion`.

We may well want to treat them differently in other places too.

If we want to treat user and org codelists differently at all, then we'll still have to have this kind of logic somewhere.  Ghost organisations are one solution (and one which I prefer to other ideas I threw away while implementing this).

What we've got works, and we can avoid proliferation of dispatching on `codelist_type` by creating `OrganisationCodelistFacade` an `UserCodelistFacade` classes, and dispatching on `codelist_type` just once in a `.facade` method.

(It's possible we could do something neater with [django-polymorphic](https://django-polymorphic.readthedocs.io/en/stable/).)

Keeping what we've got won't paint us into a corner in future, so in the interests of keeping moving, I'd like to stick with what we've got.

But thanks for making me think through this properly.

ghickman:
> the situation isn't as bad as I first thought

> won't paint us into a corner in future

> in the interests of keeping moving

Awesome to hear all that, totally agree on forward progress, and as you're still happy with the current situation after a re-read and thinking through I'd say that shows we're heading in the right direction!

ghickman (approved):

PR #347, opened 2020-11-19
Use tags and hashes for versions

inglesp:
This PR introduces hashes (eg `42abe21f`) to identify codelist versions.  These are computed from the primary key of the `CodelistVersion` object.  This is because when codelists are properly editable, we expect that they may change multiple times in a day.  As such, the old system of using the date + a single-character suffix (eg `2020-11-19-q`) is going to get a bit ugly.

Codelist versions can also be tagged.  (`CodelistVersion.tag` replaces `CodelistVersion.version_str`.)  Versions are accessible by either `/<codelist_slug>/<tag>/` or `/<codelist_slug>/<hash>/`.  Currently, a user cannot set a tag on a version, but this will follow.

This PR also removes some of the fuss around redirecting between draft/published versions.


opencodelists/hash_utils.py
```diff
@@ -0,0 +1,39 @@
+# This module contains a pair of functions for efficiently converting between integers
```
ghickman, line 1: Modules can have docstrings 🙂

opencodelists/hash_utils.py
```diff
@@ -0,0 +1,39 @@
...
+# 08d91b7b
+# 0a9e20fa
+# 0c632679
+# 0e282bf8
+# 0fed3177
+#
+# I can't remember how this works, but mumble mumble Euler's theorem mumble mumble .
+
+
+N = 2 ** 31 - 1  # Do not change this value!
```
ghickman, line 24: Could you explain what this number is used for?  It's great to say we shouldn't change but I'd like to know why too!

opencodelists/hash_utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def hash(n, key):
```
ghickman, line 27: Can you explain how the hash works in a docstring?  Or link to an explainer of the algorithm?

opencodelists/hash_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    if n >= N:
+        raise ValueError(f"{n} is too large")
+    c = int(key, 36)
+    return hex((n * c) % N)[2:].rjust(8, "0")
+
+
+def unhash(h, key):
```
ghickman, line 34: Same again here.

ghickman (changes requested):
Looking good!

The hashing module could do with more docs though, it's pretty opaque currently.

ghickman (approved):

PR #366, opened 2020-11-25
Don't try to show missing descriptions

inglesp:


inglesp:
Merging myself, because the site is a bit broken.

PR #371, opened 2020-11-26
Add new migration to fix broken data created by previous migration

inglesp:
Have tested locally, with production database.


ghickman (approved):
missing `__code`, how painful 😫

A+ explainer and comments in the migration 👍

PR #384, opened 2020-11-30
Fix fixtures

inglesp:
This introduces a new fixtures module, with a universe of fixtures that should be sufficient for testing all existing functionality.

The goal is to replace all existing fixtures and factories with fixtures in this module.

### DIFF
--- a/opencodelists/tests/fixtures.py
+++ b/opencodelists/tests/fixtures.py
@@ -54,6 +54,7 @@ AVAILABLE_FIXTURES = {
     "version_with_no_searches",
     "version_with_some_searches",
     "version_with_complete_searches",
+    "new_style_version",
     "draft_codelist",
     "draft_version",
     "user_codelist",
@@ -202,6 +203,10 @@ def build_fixtures():
         updates=[("116309007", "-")],  # exclude "Finding of elbow region"
     )
     save(draft=version_with_complete_searches)
+
+    # new_style_version
+    # - an alias for version_with_some_searches
+    new_style_version = version_with_some_searches
 
     # draft_codelist
     # - belongs to organisation
--- a/opencodelists/tests/fixtures.py
+++ b/opencodelists/tests/fixtures.py
@@ -12,9 +12,9 @@ a given state you might need to have an Organisation, a User, a Codelist, and ot
 CodelistVersions.  The approach we have taken is to build a "universe" of test objects
 once, and then to pull out members of that universe as and when we need them.
 
-This module contains some Cleverness, through use of locals() and dynamically generating
-pytest fixtures.  This is acceptable Cleverness, because it should not leak out of the
-module, and users of the module do not need to understand in detail how it works.
+Since each fixture function needs to do the same thing (find the member of the universe
+with the given name, reload it from the database, and return it) we provide
+build_fixture() to avoid excessive duplication.
 """
 
 import csv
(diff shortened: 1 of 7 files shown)


opencodelists/tests/fixtures.py
```diff
@@ -0,0 +1,289 @@
+"""This module contains fixtures to be used in all apps.
+
+One of the design goals of the project is that there should be a limited and controlled
+number of states that objects can be ing  This is achieved by requiring (through
```
ghickman, line 4: `ing` -> `in.`

opencodelists/tests/fixtures.py
```diff
@@ -0,0 +1,289 @@
...
+in one of the actions.py modules.
+
+As such, test fixtures should only be created through invoking those actions.
+
+However, this can make testing fiddly, since for instance to create a CodelistVersion in
+a given state you might need to have an Organisation, a User, a Codelist, and other
+CodelistVersions.  The approach we have taken is to build a "universe" of test objects
+once, and then to pull out members of that universe as and when we need them.
+
+This module contains some Cleverness, through use of locals() and dynamically generating
```
ghickman, line 15: `Cleverness` could benefit from some tongue-in-cheek ™ 😉

Not a requirement!

inglesp:
@ghickman feel free to wait for me to get tests passing before reviewing!

opencodelists/tests/fixtures.py
```diff
@@ -0,0 +1,289 @@
...
+        return fixture
+
+    # Update the module's locals with the fixture's name.  This lets us do:
+    #
+    #   from opencodelists.tests.fixtures import *
+    #
+    # in conftest.py.
+    locals()[fixture_name] = pytest.fixture(scope="function")(
+        build_fixture(fixture_name)
+    )
```
ghickman, line 289: Maybe I'm missing something here, but why not build them all in a `conftest.py` module to save doing this bit?

ghickman (commented):
This is definitely very clever!

But I'm not sure I see why the cleverness is needed, maybe I'm missing something??

It looks like you've created a selection of (useful!) fixtures, then dynamically loaded them into Pytest's fixture system… but could you not define them in `conftest.py` instead and avoid all the cleverness?

inglesp:
> It looks like you've created a selection of (useful!) fixtures, then dynamically loaded them into Pytest's fixture system… but could you not define them in conftest.py instead and avoid all the cleverness?

We'd end up with lots of:

```
@pytest.fixture()
def codelist(universe):
    codelist = universe["codelist"]
    codelist.refresh_from_db()
    return codelist

@pytest.fixture()
def version(universe):
    version = universe["version"]
    version.refresh_from_db()
    return version
```

which is fine, but the block would appear 14 times (so far).  To me this feels like a good reason to introduce some cleverness.

@sebbacon @evansd what do you think?

sebbacon:
Not got time to think hard about it; but if the local-magic *happens* in conftest (i.e. still avoiding copy and paste duplication) that might be nice because it's a place people might look when wondering where the fixtures are coming from?

Or a comment in there, at least?

evansd:
I very much applaud the strategy here. I think you could reduce the magic significantly with only a slight increase in boilerplate by doing something like this in `conftest.py`:
```
from opencodelists.tests.fixtures import build_fixture

organisation = build_fixture("organisation")
organisation_admin = build_fixture("organisation_admin")
organisation_user = build_fixture("organisation_user")
...
```

To avoid having to apply the `pytest.fixture` decorator to each one you could move it inside `build_fixture()` and apply it to the inner `fixture()`. (You may need to explicitly pass `name` to the pytest decorator so it knows what the fixture should be called.)

You could maybe rename `fixtures.py` to `dynamic_fixtures.py` or something to make it clear that something clever is going on here.

ghickman:
> We'd end up with lots of

Ah, that's what I was missing!  That definitely doesn't seem ideal, I don't like the idea of having to update 14+ functions.

A mixture of Dave & Seb's suggestions sounds prtty good:
* Move the fixture definitions inside a conftest.py, so it's obvious where they're coming from.
* Build each fixture using a function defined elsewhere, so one can find the fixture but conftest.py isn't cluttered with the `universe` gubbins.

opencodelists/tests/fixtures.py
inglesp: Fixed in f90032fe285ed70a9b4e821773d90046a0231876.

evansd (approved):

ghickman (approved):

PR #390, opened 2020-12-04
Fix CSV download

inglesp:
Also includes a couple of other random commits from the branch I was working on, sorry.

### DIFF
--- a/builder/actions.py
+++ b/builder/actions.py
@@ -88,6 +88,7 @@ def save(*, draft):
     All this does is unset the draft_owner attribute.
     """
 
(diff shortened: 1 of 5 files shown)


builder/actions.py
```diff
@@ -88,6 +88,7 @@ def save(*, draft):
     All this does is unset the draft_owner attribute.
     """
 
+    assert not draft.code_objs.filter(status__in=["?", "!"]).exists()
```
ghickman, line 91: A comment with this line just to sum up what it's asserting would be great, I can't remember what `?` and `!` mean.

ghickman (approved):
The comment can come whenever you have the time!

PR #403, opened 2020-12-09
Enable the service

madwort:
* this means the service will start automatically when the host machine is restarted


inglesp (approved):
Cheers.

PR #440, opened 2021-01-04
Allow searching for codes in builder

inglesp:
This is already supported for BNF codes

### DIFF
--- a/coding_systems/snomedct/coding_system.py
+++ b/coding_systems/snomedct/coding_system.py
@@ -1,6 +1,8 @@
 import collections
 import re
 
+from django.db.models import Q
+
 from opencodelists.db_utils import query
 
 from .models import FULLY_SPECIFIED_NAME, IS_A, Concept, Description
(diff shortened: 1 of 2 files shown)


coding_systems/snomedct/coding_system.py
```diff
@@ -25,9 +27,11 @@ def lookup_names(codes):
 
 def search(term):
     return set(
-        Concept.objects.filter(
-            descriptions__term__contains=term, descriptions__active=True, active=True
-        ).values_list("id", flat=True)
+        Concept.objects.filter(active=True)
+        .filter(
+            Q(descriptions__term__contains=term, descriptions__active=True) | Q(id=term)
+        )
```
ghickman, line 33: Are these in two different `.filter()` calls on purpose?

ghickman (approved):

coding_systems/snomedct/coding_system.py
inglesp: Yes, I find `.filter(q1).filter(q2 | q3)` slightly easier to parse than `.filter(q1 & (q2 | q3))`.

ghickman: +1 to that!

PR #442, opened 2021-01-04
Add <strong> and <em> to allowed markdown tags

### LINKED ISSUE #441: Some markdown being double-converted
`strong` and `em` tags in eg codelist methodologies are being double-converted.

```
# heading

[a link](https://opensafely.org/)

Some _em_ text
Some **strong** text
```

is displayed as 

image

inglesp:
Fixes #441.

@ghickman can you think of any other tags that should be allowed?

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -208,6 +208,8 @@ MARKDOWN_FILTER_WHITELIST_TAGS = [
     "ul",
     "li",
     "code",
+    "em",
+    "strong",
 ]
 
 


ghickman:
@inglesp – `` maybe?  I've used it in other markdown textfields but would it be of any use in a codelist??

inglesp:
> @inglesp – `` maybe? I've used it in other markdown textfields but would it be of any use in a codelist??

Good thought, but not sure that's something we need to support.

ghickman (approved):

PR #455, opened 2021-01-12
Don't chown everything on deploy

inglesp:
This is a hangover from when fabric ran as root

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -37,10 +37,6 @@ def install_python_requirements():
         run("pip install --quiet --requirement requirements.txt")
 
 
-def chown_everything():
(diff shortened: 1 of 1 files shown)


evansd (approved):

fabfile.py
```diff
@@ -37,10 +37,6 @@ def install_python_requirements():
         run("pip install --quiet --requirement requirements.txt")
 
 
-def chown_everything():
-    run(f"chown -R {env.user}:www-data {env.path}")
```
madwort, line 41: do we still need to set group `www-data`, though? or are all relevant files set to world-readable?

PR #464, opened 2021-01-15
Fix tree view

inglesp:
This was broken by fa47d6ca.

Users who didn't own the codelist saw:

image

They now see:

image

and nothing happens if you click the buttons.


evansd (approved):
```
updateStatus && updateStatus.bind(null, code, symbol)
```
Ah, of course

PR #469, opened 2021-01-19
Add API endpoint for creating new codelist versions

inglesp:
It's not yet clear how much of an API we'll need and using DRF is a bit of a sledgehammer for this endpoint -- but it may prove useful if we want to extend the API in future.  At the moment, we're only really using it for authentication.

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -66,6 +66,8 @@ INSTALLED_APPS = [
     "django_extensions",
     "debug_toolbar",
     "markdown_filter",
+    "rest_framework",
+    "rest_framework.authtoken",
     "django.contrib.admin",
     "django.contrib.auth",
     "django.contrib.contenttypes",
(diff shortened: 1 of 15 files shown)


opencodelists/settings.py
```diff
@@ -234,3 +234,15 @@
...
+# DRF
+
+REST_FRAMEWORK = {
+    "DEFAULT_AUTHENTICATION_CLASSES": [
+        "rest_framework.authentication.TokenAuthentication",
+    ],
+    "DEFAULT_PERMISSION_CLASSES": [
+        "rest_framework.permissions.IsAuthenticated",
+    ],
+}
```
ghickman, line 248: As there's only one endpoint currently, you could keep this with the view?

ghickman (approved):
👍

+1 for starting with an APIView view!

opencodelists/settings.py
inglesp: TIL.  Won't change that now though.

PR #479, opened 2021-01-25
Add rough and ready codelist diff view

inglesp:


templates/codelists/version_diff.html
```diff
@@ -0,0 +1,79 @@
...
+                        {% if record.descendants %}with these descendants{% endif %}
+                </li>
+                <ul>
+                {% for descendant in record.descendants %}
+                <li>{{ descendant.term }} (<code>{{ descendant.code }}</code>)</li>
+                {% endfor %}
+                </ul>
+                {% endfor %}
+                </ul>
+        </div>
```
ghickman, line 75: Looks like you've indented with tabs here, was that intentional?

ghickman (approved):

templates/codelists/version_diff.html
inglesp: No, silly vim.

PR #508, opened 2021-02-09
Support inactive SNOMED codes

inglesp:
Until now, inactive SNOMED codes haven't shown up in searches in the builder, and they have been dropped when creating a codelist either by uploading a CSV file or from an ECL expression.

This PR fixes that, but also includes a significant refactor to introduce a `Codeset` class, which encapsulates a lot of the logic to do with a collection of codes within the context of part of a coding system hierarchy.

Further work will be required to support codelists that contain concepts that were once active and are now inactive (and changes to the SNOMED hierarchy more generally).


ghickman (approved):
Really clear sequence of commits!

This looks great, loving the evolution of Definition too.

PR #535, opened 2021-02-22
Add endpoint to list an organisation's codelists

inglesp:


inglesp:
Merging myself because I need this now.  @ghickman please could you do a retrospective review when you get the chance?

ghickman (approved):
👍

You could do the data gathering with a serialiser since it guarantees various data formatting bits, but I'm not sure what a single endpoint would necessarily gain from that.

inglesp:
> You could do the data gathering with a serialiser since it guarantees various data formatting bits, but I'm not sure what a single endpoint would necessarily gain from that.

You're right, but I don't know how to do that and don't have time to learn right now!

PR #591, opened 2021-03-22
Add banner with links to videos and docs

inglesp:
image


evansd (approved):
No idea why this would cause your tests to fail!

PR #596, opened 2021-03-23
Add introduction and videos to homepage

inglesp:
image

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -111,7 +111,9 @@
       </div>
       {% endfor %}
 
-      {% block content %} {% endblock %}
+      <div class="mt-4">
+        {% block content %} {% endblock %}
+      </div>
 
       <script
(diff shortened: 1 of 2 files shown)


ghickman:
@inglesp – fixed both issues!

The menu was an easy one, it was missing the button and the collapsible wrapper (which only triggers on smaller viewports).

The videos were less fun: everyone [and their dog] suggests the CSS I've used to make them response but because we move to relative/absolute positioning we lose the ability (I think) to set spacing as we'd like, so I took the pragmatic approach and only applied them to viewports up 768px (so that's all viewports up to and including what Bootstrap classifies as Medium).  It all appears to work for me on macOS 11/Firefox 86, but I haven't used max-width in combination with Bootstrap (which uses min-width) before so there's possibly edgecases I don't know about.

**Hamburger Menu**


**iPhone X/XS iOS 12**


**iPad**


**Desktop**

PR #602, opened 2021-03-25
Support codesets for old-style versions

inglesp:
This fixes the broken "Download Definition" link for old-style versions.


ghickman (approved):
painful! 😩

PR #621, opened 2021-03-31
Fix bug in icd10.code_to_term

inglesp:
Previously, codes at the end of a block (eg M779 in block M75-M77) were
not captured correctly.

### DIFF
--- a/coding_systems/icd10/management/commands/generate_icd10_hierarchy_fixture.py
+++ b/coding_systems/icd10/management/commands/generate_icd10_hierarchy_fixture.py
@@ -1,0 +1,50 @@
+"""
+Generate a JSON fixture containing every Concept related to Concepts with given codes.
+
+The generated fixture can be loaded with `loaddata`.
+"""
+from operator import attrgetter
+
+from django.core import serializers
+from django.core.management import BaseCommand
+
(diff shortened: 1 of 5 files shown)


coding_systems/icd10/management/commands/generate_icd10_hierarchy_fixture.py
```diff
@@ -0,0 +1,50 @@
...
+            concept = todo.pop()
+            print(concept.code, concept.term)
+            concepts.add(concept)
+
+            todo.extend(concept.children.all())
+
+        concepts = sorted(concepts, key=attrgetter("pk"))
+
+        with open(path, "w") as f:
+            serializers.serialize("json", concepts, stream=f, indent=2)
```
ghickman, line 50: Why serializers here instead of `json.dump(…)`?

ghickman (approved):

coding_systems/icd10/management/commands/generate_icd10_hierarchy_fixture.py
inglesp: Because `json.dump(instance)` doesn't know how to produce data in the format that `loaddata` can read.

PR #644, opened 2021-04-13
Drop ctv3tpp coding system

inglesp:
It's a relic.  TPP's Y-Codes are now included in the ctv3 coding system.

Note: before deploying, run

    Codelist.objects.filter(coding_system_id="ctv3tpp").update(coding_system_id="ctv3")


ghickman (approved):
🎉

PR #666, opened 2021-04-21
More API tweaks

inglesp:


codelists/tests/test_api.py
```diff
@@ -171,9 +166,21 @@ def test_versions_post_permission_denied(
...
 
     assert rsp.status_code == 403
+
+
+def post(client, url, data, user):
+    if user is None:
+        headers = {}
+    else:
+        headers = {"HTTP_AUTHORIZATION": f"Token {user.api_token}"}
+    return client.post(url, data, content_type="application/json", **headers)
```
ghickman, line 186: Not sure this abstraction gains you much in the long run?  If one is digging into a broken test they now have to look around for this function, when all you're really doing is saving the addition of passing a dictionary to `client.post`!

ghickman (approved):

PR #671, opened 2021-04-23
Change how searching by code works

inglesp:
Instead of having two buttons, one for searching by code and one for
searching by term (which seems to be causing confusion) we now have a
single "search" button.  By default, we search by term, but if the
search string matches "code:xxx", we search for code xxx instead, and if
the search string matches "code:xxx*", we search for concepts whose code
begins with xxx.


builder/views.py
```diff
@@ -163,14 +163,12 @@ def update(request, draft):
 @require_http_methods(["POST"])
 @load_draft
 def new_search(request, draft):
-    if request.POST["field"] == "term":
-        term = request.POST["search"]
-        code = None
-    elif request.POST["field"] == "code":
+    term = request.POST["search"].strip()
+    if term.startswith("code:"):
+        code = term[5:].strip()
```
ghickman, line 168: Would `str.removeprefix("code:")` be of use here?

codelists/models.py
```diff
@@ -392,7 +392,7 @@ class Meta:
 
     @property
     def term_or_code(self):
-        return self.term or self.code
+        return self.term or f"code: {self.code}"
```
ghickman, line 395: Is there supposed to be a space after the colon here?

coding_systems/icd10/coding_system.py
```diff
@@ -20,18 +20,17 @@ def search_by_term(term):
...
-        return set()
-
-    # The UI does not support top-level concepts being included/excluded (and in any
-    # case, they cannot appear on a patient record) so if a user searches for a chapter
-    # code, we instead return the codes for all children of that chapter.
-    if concept.kind == "chapter":
-        return {child.code for child in concept.children.all()}
+    code = code.upper()
+    if code[-1] == "*":
+        kwargs = {"code__startswith": code[:-1]}
```
evansd, line 25: nano-nit: If you used `code.endswith("*")` you'd be robust against having an empty string (although presumably that's enforced elsewhere. And you used `code.rstrip("*")` you'd be robust against a user accidentally supplying multiple wildcards.

ghickman (approved):

evansd (approved):
Looks like a very sensible UI decision and a clean implementation.

I made one tiny comment which should you probably ignore.

codelists/models.py
inglesp: Yes, this is for display.

builder/views.py
inglesp: TIWR (today I was reminded...)

coding_systems/icd10/coding_system.py
inglesp: Nice idea.

builder/views.py
inglesp: Except we're not running on 3.9 :(

ghickman: One to come back to later then 🙂

PR #676, opened 2021-04-26
Fix ICD-10 importer to handle modified codes

inglesp:
I have verified that the codes output by this change matches the codes
published by the WHO in their own browser at
https://icd.who.int/browse10/2019/en.

### DIFF
--- a/coding_systems/icd10/import_data.py
+++ b/coding_systems/icd10/import_data.py
@@ -1,6 +1,8 @@
 """Import ICD-10 data from
 https://apps.who.int/classifications/apps/icd/ClassificationDownload/DLArea/Download.aspx"""
 
+from collections import defaultdict
+
 from django.db import transaction
 from lxml import etree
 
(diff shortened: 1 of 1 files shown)


coding_systems/icd10/import_data.py
```diff
@@ -17,8 +19,67 @@ def import_data(release_path):
...
+    """
+
     root = doc.getroot()
 
+    modifiers = defaultdict(list)
+
+    for e in root.findall("ModifierClass"):
+        modifier = e.get("modifier")
+        code = e.get("code")
+        if code[0] != ".":
```
ghickman, line 73: Would `str.startswith` be more approrpriate here, to avoid problems with `code` being an empty string?  Or do we know that `code` will always be 1+ long?

ghickman (approved):
Looks good, just a small query, which is probably more data related than anything.

coding_systems/icd10/import_data.py
inglesp: Good suggestion, but (a) `code` is never empty and (b) we're not running 3.9 yet.

evansd: `startswith` was added at some point between 1.6 and 2.0.

inglesp: You're very right.  I was thinking of `str.removeprefix`, or rather, I was not thinking.

Issue #678, opened 2021-04-27
Allow users to change codelist name or owner


inglesp:
We want to allow users to rename a codelist or move it to another owner.  At the moment, we just need to consider changing ownership from an individual user to an organisation that they are a member of.

When making this kind of change, old URLs should redirect.  (And this will allow us to remove the hard-coded redirects from `urls.py`)

### Implementation notes

* Add new model `Handle`, with `name`, `slug`, `organisation`, `user` fields copied from `Codelist`, with a foreign key to `Codelist` and a boolean `is_current` field
* Add data migration to create relevant `Handle` models
* Remove fields that are now on `Handle` from `Codelist`
* Add `name`, `slug`, `organisation`, `user` properties to `Codelist`
* Update `create_codelist` and `create_codelist_*` actions to create `Handle` instances (inside a transaction)
* Update `update_codelist` action to take a `owner`, `name` (and maybe `slug`?) parameters, and to create a new `Handle` if required
* Update `CodelistUpdateForm` to add `owner`, `name` (and maybe `slug`) fields
  * See `owner_choices` in `opencodelists.forms.CodelistCreateForm` for inspiration
* Update `load_codelist` and `load_version` decorators to redirect to current handle
* Create `Handle` objects for redirects that are hard-coded in `urls.py`

sebbacon:
Just to note that when we're thinking about organisation membership, we should do so with an eye to unifying one day with the work @ghickman is doing.  Sorry if repeating myself - pretty sure this has been discussed before!

For example, the permission to change owner or set organisation could be looked up on the new API. At some point.

PR #683, opened 2021-04-29
Add Handle model

inglesp:
This PR adds a `Handle` model.  A `Handle` belongs to a `Codelist`, and holds the fields used to identify the codelist (name, slug, organisation, user).

This prepares the ground for implementing #678 (Allow users to change codelist name or owner).


codelists/models.py
```diff
@@ -130,6 +130,42 @@ def latest_version(self):
...
+        related_name="handles",
+        on_delete=models.CASCADE,
+    )
+    user = models.ForeignKey(
+        "opencodelists.User",
+        null=True,
+        related_name="handles",
+        on_delete=models.CASCADE,
+    )
+    is_current = models.BooleanField()
```
ghickman, line 153: Should this have a default or `null=True` set?

codelists/models.py
```diff
@@ -130,6 +130,42 @@ def latest_version(self):
...
+        unique_together = [("organisation", "name", "slug"), ("user", "name", "slug")]
+        constraints = [
+            models.CheckConstraint(
+                name="%(app_label)s_%(class)s_organisation_xor_user",
+                check=(
+                    models.Q(organisation_id__isnull=False, user_id__isnull=True)
+                    | models.Q(user_id__isnull=False, organisation_id__isnull=True)
+                ),
+            )
+        ]
```
ghickman, line 168: It's _so_ nice to see these constraints in the wild 💪

codelists/migrations/0034_auto_20210429_1029.py
```diff
@@ -0,0 +1,24 @@
...
+            is_current=True,
+        )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("codelists", "0033_auto_20210429_1029"),
+    ]
+
+    operations = [migrations.RunPython(create_handles)]
```
ghickman, line 24: Can you add a reverse step to this (even if it's a no-op)?  Makes the migration a little easier to use 🙂

codelists/actions.py
```diff
@@ -110,17 +90,60 @@ def create_codelist_with_codes(
...
-        slug=slug, name=name, coding_system_id=coding_system_id
+    codelist = Codelist.objects.create(
+        coding_system_id=coding_system_id,
+        description=description,
+        methodology=methodology,
     )
-    codelist.versions.create(draft_owner=draft_owner)
+    owner.handles.create(codelist=codelist, slug=slug, name=name, is_current=True)
+
+    for reference in references or []:
```
ghickman, line 141: The `or []` here is a really nice touch for keeping the kwarg immutable 👍

ghickman (changes requested):
A wonderfully readable PR!  Love some of the paradigms you've used here.

Couple of questions/suggestions which are just nits but I think the BooleanField on Handle needs either a default or to be nullable in case it's ever accessed outside of the actions (or you could just comment as to why it's not got either!)

The thing I don't think I really understand is what the need for Handle is?  I feel like it's me missing the final piece in the puzzle though.

codelists/migrations/0034_auto_20210429_1029.py
inglesp: Good idea.

codelists/models.py
inglesp: No, because I want to always be set explicitly.

inglesp:
> The thing I don't think I really understand is what the need for Handle is? I feel like it's me missing the final piece in the puzzle though.

I hadn't spelt this out, sorry.  We want to allow users to be able to change a codelist's name, slug, or owner.  Doing so will change the codelist's URL, but it's important that we can continue to serve codelists from the old URL.  So when a user changes a codelist's name/slug/owner, we'll create a new handle, and set `is_current` to `False` on the old handle.  Does that make sense?

codelists/migrations/0034_auto_20210429_1029.py
inglesp: I think I'm going to leave this.  The [next migration (0035)](https://github.com/opensafely-core/opencodelists/pull/683/files#diff-bf77995fe00a8f8fe70a14fbb5c14e0194bf2e4b3ada400006ed6fc2ee16d91d) removes a bunch of fields from `Codelist`, some of which are required.  As such, 0035 is not reversible.  So to make a completely reversible chain of migrations, I'd have to have something like:

migration | forwards | backwards
-|-|-
0033 | add Handle model | remove Handle model (automatic)
0034 | make fields nullable on Codelist | make fields required on Codelist (automatic)
0035 | create Handle instance for each Codelist | populate Codelist fields from Handle
0036 | delete fields from Codelist | create fields on Codelist (automatic)

This is all doable, but given that we're not doing development here in parallel, I don't think it's worth it.

codelists/actions.py
evansd: Setting the default to empty tuple (`references=()`) is another similar trick you can use in this context. Though here it would require using it as the default all the way up the call-chain so it would be more awkward.

ghickman (approved):

ghickman:
Yep, that's exactly what I needed to connect the dots 👍

PR #694, opened 2021-05-06
Rework tests

inglesp:
This is going to be stupefyingly dull to review, so I promise to review the next stupefyingly dull PR to come along.


evansd (approved):
This was actually quite pleasant to review (lots of beautiful red everywhere) apart from that "remove factory_boy" commit.

PR #724, opened 2021-05-19
Update OpenCodelists URL

StevenMaude:
In the README and documentation index page.


madwort (approved):

inglesp:
Thanks for doing this @StevenMaude!  The remaining references should not be changed, as it's important we continue to serve codelists from the original domain.

PR #736, opened 2021-05-26
Improve builder performance

inglesp:
This PR significantly speeds up the backend responses to updates from the builder.  This makes the frontend more responsive.  Additionally it speeds up other views (although this is less important).

It does this by reducing the amount of work done to process updates, and then by caching the `Hierarchy` instance belonging to each `CodelistVersion`.


codelists/migrations/0043_auto_20210526_0952.py
```diff
@@ -0,0 +1,51 @@
...
+        CachedHierarchy.objects.create(version=version, data=hierarchy.data_for_cache())
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("codelists", "0042_cachedhierarchy"),
+    ]
+
+    operations = [migrations.RunPython(set_cached_hierarchy)]
```
ghickman, line 51: Without a reverse_code (even migrations.RunPython.noop) this migration will be irreversible, did you mean to do that?

ghickman (approved):
Just the one point on the migration reversal 👍

PR #819, opened 2021-07-21
Delete backups older than 30 days

inglesp:


deploy/bin/backup.sh
```diff
@@ -13,3 +13,10 @@ BACKUP_PATH="/mnt/volume-lon1-01/opencodelists-backups/core-data-$(date +%F).jso
         --indent 2 \
         --verbosity 0 \
         --output "$BACKUP_PATH"
+
+# Keep only the last 30 backups
+# The disabled check recommends using find instead of ls, but (a) the reasons for doing
+# so (non-alphanumeric filenames) don't apply here and (b) the "proper" way of doing it
+# is unreadable (https://superuser.com/a/608889).
+# shellcheck disable=SC2012
+ls -tr "$BACKUP_PATH/core-data-*.json.gz" | head -n -30 | xargs rm
```
madwort, line 22: This is fine, alternatively we could do something like 
```
find "$BACKUP_PATH" -name "core-data-*.json.gz" | sort | head -n -30 | xargs rm
```
so that we can remove the comment & shellcheck disable

madwort: (because our filenames should sort correctly by date)

inglesp: Ingenious

madwort (approved):

PR #851, opened 2021-08-19
Fix backup cleanup

inglesp:


madwort (approved):
looks like it'll work, let's try it out

PR #946, opened 2021-10-20
Bump all Python requirements

inglesp:


StevenMaude (approved):

requirements.txt
```diff
@@ -1,37 +1,37 @@
 #
-# This file is autogenerated by pip-compile
+# This file is autogenerated by pip-compile with python 3.6
```
StevenMaude, line 2: NB (but you probably know): Python 3.6 support [ends at the end of 2021](https://devguide.python.org/#status-of-python-branches).

PR #965, opened 2021-11-02
Allow versions to be deleted

inglesp:
This PR adds a new "Delete version" button for users who can edit a codelist.

image

When clicked, the user sees an "are you sure?" modal:

image

The button is greyed out if the version has already been published, since we do not allow a published version to be changed.


iaindillingham (approved):
🚀

PR #966, opened 2021-11-02
Support concepts with duplicate Fully Specified Names

inglesp:


StevenMaude (approved):
Does this behaviour need noting anywhere that's user-facing?

Or are these just small variations in the fully specified name that don't really matter for users?

StevenMaude:
Aside from my query, looks fine :+1:

inglesp:
> Does this behaviour need noting anywhere that's user-facing?

Good question.  No, I don't think so.  It affects ~15 concepts (out of ~500k), and in all but two cases the selected terms are identical.  I'll update the comment to reflect this.

PR #1063, opened 2022-02-01
feat: Clean csv header

### LINKED ISSUE #1062: Strip leading/trailing spaces from column headers
re https://ebmdatalab.slack.com/archives/C33TWNQ1J/p1643387639934759 and this codelist https://www.opencodelists.org/codelist/nhsd/oral-steroid-drugs-pra-dmd/4d081f73/#full-list 

the dmd_id column header has a leading space " dmd_id" which is not clearly visible on the rendered html of the codelist page but still mismatches with "dmd_id" when used within cohortextractor.

suggest stripping trailing and leading spaces from column headers upon code list creation/csv upload.

Jongmassey:
closes  #1062

### DIFF
--- a/codelists/forms.py
+++ b/codelists/forms.py
@@ -94,7 +94,8 @@ class CSVValidationMixin:
         data = self.cleaned_data["csv_data"].read().decode("utf-8-sig")
 
         reader = csv.reader(StringIO(data))
-        num_columns = len(next(reader))  # expected to be headers
+        header = next(reader)  # expected to be headers
+        num_columns = len(header)
         errors = [i for i, row in enumerate(reader) if len(row) != num_columns]
 
         if errors:
@@ -105,6 +106,10 @@ class CSVValidationMixin:
                     for i in errors
                 ]
             )
+
+        cleaned_header = [h.strip() for h in header]
+        if header != cleaned_header:
+            data = data.replace(",".join(header), ",".join(cleaned_header), 1)
 
         return data
 
--- a/codelists/tests/test_forms.py
+++ b/codelists/tests/test_forms.py
@@ -25,7 +25,9 @@ def test_csvvalidation_correct_csv_column_count():
 
     # wrap CSV up in SimpleUploadedFile to mirror how a Django view would
     # handle it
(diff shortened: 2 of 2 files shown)


codelists/tests/test_forms.py
```diff
@@ -50,3 +50,19 @@ def test_codelistform_incorrect_csv_column_count():
...
+    form = CSVValidationMixin()
+
+    # wrap CSV up in SimpleUploadedFile to mirror how a Django view would
+    # handle it
+    csv_data = "code , description\n1067731000000107,Injury whilst swimming (disorder)"
+    upload_file = csv_builder(csv_data)
+    uploaded_file = SimpleUploadedFile("our csv", upload_file.read())
+    form.cleaned_data = {"csv_data": uploaded_file}
+
+    cleaned_data = form.clean_csv_data()
```
inglesp, line 65: How about:

```
cleaned_data = form.clean_csv_data()
header_line = cleaned_data.splitlines()[0]
assert header_line == "code,description"
```

inglesp (approved):
Nice, thanks for diving in and fixing this.

codelists/tests/test_forms.py
Jongmassey: That's certainly more concise

codelists/forms.py
```diff
@@ -106,6 +107,10 @@ def clean_csv_data(self):
                 ]
             )
 
+        cleaned_header = [h.strip() for h in header]
+        if header != cleaned_header:
+            data = data.replace(",".join(header), ",".join(cleaned_header), 1)
```
madwort, line 112: are all your CSV files guaranteed not to have quoted items? I think this would fail if they do? 
i.e. your test case is
```csv
 code , description
1234, stuff
```
but this would also be a valid csv file:
```csv
" code ", " description"
"1234","stuff"
```
and csv parsers typically parse `" code", " description"` to `[" code", "description"]`, not `["\" code\"", "\"description\""]` - I think the replace would then fail?

madwort (changes requested):

PR #1065, opened 2022-02-01
feat: Clean csv header

Jongmassey:
take 2 of #1063 with @madwort  suggestions

### DIFF
--- a/codelists/tests/test_forms.py
+++ b/codelists/tests/test_forms.py
@@ -83,6 +83,6 @@ def test_codelistform_header_stripping_quoted():
 
     cleaned_data = form.clean_csv_data()
     header_line = cleaned_data.splitlines()[0]
-    assert header_line == "code,description"
+    assert header_line == '"code","description"'
     data_line = cleaned_data.splitlines()[1]
     assert data_line == '"1067731000000107","Injury whilst swimming (disorder)"'


Jongmassey:
Second go at this. Don't feel great about needing a regex to do the stripping, happy to take advice if there's a more efficient or pythonic way of doing it.

madwort (dismissed):
I'm happy with it, would be nice to tweak the tests though

codelists/forms.py
```diff
@@ -105,6 +107,12 @@ def clean_csv_data(self):
                     for i in errors
                 ]
             )
+        cleaned_header = [
+            "".join(re.match(r"(\"?)\s*(\w+)\s*(\"?)", h).groups()) for h in header
```
madwort, line 111: nb. again I have to lean on your knowledge of the source data - obviously that `(\w+)` won't match any space chars - so if your header was two words `" my data "` it would not match. Hopefully they're all one word & this is fine?

Jongmassey: That's another good catch! I think we need to allow for it, but I wasn't thinking about that when I wrote that pattern. I'll fix it.

codelists/forms.py
```diff
@@ -106,6 +107,14 @@ def clean_csv_data(self):
                 ]
             )
 
+        cleaned_header = [
+            '"' + h.strip('" ') + '"' if h[0] == '"' and h[-1] == '"' else h.strip()
```
madwort, line 111: yeah, this looks better!

madwort: although you'd still have a problem with escaped quotes in strings - does your source data contain those? `" my \"data\""`

madwort: csvs are annoying

Jongmassey: I'm not sure what the reasonable expectation of user input is here. This isn't a particularly oft-used bit of functionality but we've had one instance already where some leading spaces in the header caused confusion and I thought I'd just put a quick fix in to stop it from happening again!

Jongmassey: This explicitly only removes the first and last chars of the string, which are the ones being tested  for presence of quotes. Notwithstanding a library to do this for us, I think this might be right?

codelists/forms.py
```diff
@@ -108,7 +108,7 @@ def clean_csv_data(self):
             )
 
         cleaned_header = [
-            '"' + h.strip('" ') + '"' if h[0] == '"' and h[-1] == '"' else h.strip()
+            '"' + h[1:-1].strip() + '"' if h[0] == '"' and h[-1] == '"' else h.strip()
```
Jongmassey, line 111: This explicitly only removes the first and last chars of the string, which are the ones being tested for presence of quotes. Notwithstanding a library to do this for us, I think this might be right? @madwort

PR #1082, opened 2022-02-24
Allow organisation admins to add users to organisation

### LINKED ISSUE #1048: Allow admins to add users to organisations
The [`Membership` model](https://github.com/opensafely-core/opencodelists/blob/main/opencodelists/models.py#L190-L201) links a user to an organisation.

There is an [action](https://github.com/opensafely-core/opencodelists/blob/main/opencodelists/actions.py#L40-L43) that adds users to an organisation.

This should be wired up to a view to allow admins to add users to their organisation.

rebkwok:
Fixes #1048 

If a user is a member of one or more organisations, there's now a navbar link to a view that shows a list of the users orgs, with links to the org's codelists and (if the user is an admin for that org) the org's users.

Screenshot from 2022-02-24 09-53-30

The organisation users page shows a list of users and a minimal form to add a user by email.  It assumes an admin will only add a user who's already signed up to opencodelists.

Screenshot from 2022-02-24 09-54-12

Error messages are shown if you try to add a user that doesn't exist or is already a member

Screenshot from 2022-02-24 09-54-31
Screenshot from 2022-02-24 09-54-51

When a user is successfully added, they're highlighted green in the user table and shown first:

Screenshot from 2022-02-24 10-20-08

### DIFF
--- a/opencodelists/forms.py
+++ b/opencodelists/forms.py
@@ -146,3 +146,36 @@ class RegisterForm(forms.ModelForm):
         if commit:
             user.save()
         return user
+
+
+class MembershipCreateForm(forms.Form):
+    email = forms.EmailField(
+        help_text="Enter an email address for an existing user",
+        widget=forms.TextInput(
+            attrs={"class": "form-control", "placeholder": "Email address"}
+        ),
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.organisation = kwargs.pop("organisation")
+        self.user = None
+        super().__init__(*args, **kwargs)
+
+        if "email" in self.errors:
+            self.fields["email"].widget.attrs.update(
+                {"class": "form-control border-danger"}
+            )
+
+    def clean_email(self):
+        email = self.cleaned_data["email"]
+        try:
+            self.user = User.objects.get(email=email)
+        except User.DoesNotExist:
+            self.add_error("email", f"User with email address {email} does not exist")
+
+        if self.user and self.user.is_member(self.organisation):
+            self.add_error(
+                "email", f"User with email address {email} is already a member"
+            )
+        else:
+            return email
--- a/opencodelists/forms.py
+++ b/opencodelists/forms.py
@@ -149,10 +149,11 @@ class RegisterForm(forms.ModelForm):
 
 
 class MembershipCreateForm(forms.Form):
-    email = forms.EmailField(
-        help_text="Enter an email address for an existing user",
+    user_idenitfier = forms.CharField(
+        max_length=255,
+        help_text="Enter a username or email address for an existing user",
         widget=forms.TextInput(
-            attrs={"class": "form-control", "placeholder": "Email address"}
+            attrs={"class": "form-control", "placeholder": "Username/Email"}
         ),
     )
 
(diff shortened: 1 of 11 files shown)


rebkwok:
I'm not entirely sure if we should be showing all the users in an organisation, but this view is only accessible by admin users, and presumably it's useful for them to be able to see who's already a member.

Also, there's no pagination on the table of users.  I'm not sure how many users an organisation is likely to have, and if we should paginate it.

opencodelists/models.py
```diff
@@ -32,11 +34,12 @@ def create_user(self, username, email, password, **extra_fields):
...
         email = self.normalize_email(email)
-        user = self.model(
-            username=username, email=email, organisation=org, **extra_fields
-        )
+        user = self.model(username=username, email=email, **extra_fields)
         user.set_password(password)
         user.save()
+        user.memberships.create(
+            organisation=org, date_joined=datetime.today(), is_admin=False
+        )
```
inglesp, line 42: Does `User.objects.create_user(...)` fail without this change?

Users no longer need to belong to an organisation, so I think it'd be better to remove any organisation-handling code from this function.

opencodelists/forms.py
```diff
@@ -146,3 +146,36 @@ def save(self, commit=True):
...
         return user
+
+
+class MembershipCreateForm(forms.Form):
+    email = forms.EmailField(
+        help_text="Enter an email address for an existing user",
+        widget=forms.TextInput(
+            attrs={"class": "form-control", "placeholder": "Email address"}
+        ),
+    )
```
inglesp, line 157: Given that users use usernames instead of email addresses to log in, can I suggest using username instead of email address in this form?

inglesp (approved):
This is great, thank you!  You've taken all the undocumented idioms in the codebase and have done exactly the right thing with them.

A couple of small suggestions, but nothing to hold up merging if you want to get on with other things.

opencodelists/models.py
rebkwok: Yes, because it's trying to create a user with an organisation kwarg that doesn't exist anymore.  I made it do the old thing of creating a default org, but I could just make it skip the organisation bit, if it's not necessary.  (I only discovered it because I tried to do `./manage.py createsuperuser` for my local env, which failed)

opencodelists/forms.py
rebkwok: I thought it was more likely that an admin user would know their email address than username.  Would it make sense to allow them to enter either?

inglesp: >  Would it make sense to allow them to enter either?

Yes, if that's easy!

opencodelists/models.py
inglesp: > I could just make it skip the organisation bit, if it's not necessary

I'd ditch it.

inglesp:
> Also, there's no pagination on the table of users. I'm not sure how many users an organisation is likely to have, and if we should paginate it.

At the moment an organisation might have tens of users tops, so no need for pagination.

opencodelists/forms.py
rebkwok: Done :slightly_smiling_face:

PR #1084, opened 2022-02-28
Move deployment to dokku

### LINKED ISSUE #1049: Deploy on dokku
The site is currently hosted directly on smallweb1, and deployment is currently done manaully via fabric (see fabfile.py).  It would be good it was hosted and deployed in the same way that our other sites are.

### LINKED ISSUE #1050: Upgrade to latest Python
Currently running 3.6 which is now unsupported.

I'm not sure whether it makes sense to do this before or after #1049.

rebkwok:
Fixes #1049 
Set up deployment on dokku via github actions.
As a side effect, this also:
- upgrades to python 3.8 (fixes #1050), because the Dockerfile uses our base-docker image where the default python is 3.8
- upgrades Node to v 16
- replaces Makefile with justfile in line with how we set up repos now
- sets up options for formatting/linting/testing/running local server in docker
- runs all the things in docker for CI


app.json
```diff
@@ -0,0 +1,8 @@
+{
+    "cron": [
+      {
+        "command": "/app/deploy/bin/backup.sh",
+        "schedule": "00 01 * * *"
+      }
+    ]
+  }
```
rebkwok, line 8: This is how dokku says to do cron jobs...but I'm not convinced it will work with deployment via docker image rather than git push :shrug:

.github/workflows/main.yml
```diff
@@ -1,225 +1,187 @@
...
-        run: |
-          source ${{ github.workspace }}/venv/bin/activate
-          make sort
+      - uses: extractions/setup-just@aa5d15c144db4585980a44ebfdd2cf337c4f14cb  # v1.4.0
+      - name: Check js linting
+        run: just check-js
 
   test-py:
-    needs: [format-py, lint-py, sort]
+    needs: [check]
```
ghickman: I've stopped gating tests on tooling checks in a number of repos since they often end up with a linear run time, so it turns out to be a tigher feedback loop to run them in parallel!

.github/workflows/main.yml
```diff
@@ -1,225 +1,187 @@
...
         with:
-          path: ./node_modules
-          key: install-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-      - run: make format-js
+          dockerfile: docker/Dockerfile
 
-  lint-js:
-    needs: [install-js]
+  docker-test:
+    needs: [test-py, test-js]
```
ghickman: Do we need the Python/JS tests to run before this job?  Or could we parallelise a bit here by letting this run without them as dependencies?

ghickman (approved):
Just a couple of notes about possible CI changes, I'll leave those up to you!  Great to see this being moved onto Dokku finally 🎉

.github/workflows/main.yml
rebkwok: This is weird - that `docker-test` shouldn't be a thing anymore

rebkwok: Yeah, it's not there https://github.com/opensafely-core/opencodelists/blob/dokku/.github/workflows/main.yml
There is just `test` (which does both the python and js tests) and then the `deploy` step depends on that.   Odd that the PR has shown this in the diff

ghickman: Ah, classic "reading through the commits" issue, sorry for the noise!

.github/workflows/main.yml
rebkwok: This almost halves the workflow time, thanks for the suggestion :slightly_smiling_face:

app.json
bloodearnest: Oh, interesting question. I guess we'll find out?

I suspect it might though, as *all* dokku apps are based running a docker container. It's just the the push way builds the container locally with a buildpack, whereas the `from-image` way uses a pre-built docker image.

inglesp (approved):
Have kicked the tyres locally, and everything seems to be working!

Can you add to the docs:

* where can I pull the production database from?
* where do backups go?

Also, who's going to check that backups get made?

Thanks for taking this on!

PR #1099, opened 2022-03-02
Make dokku checks work

rebkwok:


iaindillingham (approved):
I don't have enough context, really, but it looks like Dokku's checks weren't working and changing the host and port fixes them.

opencodelists/settings.py
```diff
@@ -41,13 +41,14 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DEBUG", False)
 
-BASE_URLS = env.list("BASE_URLS", default=["http://localhost:8000"])
+BASE_URLS = env.list("BASE_URLS", [])
+# note localhost is required on production for dokku checks
+BASE_URLS += ["http://localhost:8000"]
```
bloodearnest, line 46: I would have expected this to be `http://localhost:7000`?

rebkwok: It's 8000 for local development, and in ALLOWED_HOSTS it gets turned into just localhost

bloodearnest: Ah. 

So, I think that ideally if we are choosing a port to run on for an app, it should run on the same port in dev and prod. That's kinda the point of choosing a different port that the default, so we an run them side by side in dev easily.

If that's not something worth pursuing, then we should just run on the default port?

rebkwok:
That's easy enough to do

PR #1103, opened 2022-03-04
Allow users to change and reset their own passwords

### LINKED ISSUE #1047: Allow users to reset their own passwords
On the login page, there should be an "I forgot my password" button which lets a user request a link to reset their password.

There is already [a view](https://github.com/opensafely-core/opencodelists/blob/main/opencodelists/views/user_set_password.py) that lets a user set their password.  [This was added](https://github.com/opensafely-core/opencodelists/commit/1d617342962a618711735f90892a2b32e8b6b17a#diff-399d753ed281242a40bc79c6b334a8a659494b692462012dce893cd2c7688eea) before users could sign up themselves (the workflow was that I'd create an account, send them this link, and they'd set their own password) and this might be repurposable for this.

rebkwok:
Fixes #1047 

Uses django's auth urls/views, so just requires adding relevant templates, and setting up email.

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -271,5 +271,5 @@ ANYMAIL = {
 EMAIL_BACKEND = env.str(
     "EMAIL_BACKEND", default="django.core.mail.backends.console.EmailBackend"
 )
-DEFAULT_FROM_EMAIL = "you@example.com"
-SERVER_EMAIL = "your-server@example.com"
+DEFAULT_FROM_EMAIL = "no-reply@opencodelists.org"
+SERVER_EMAIL = "tech@opensafely.org"
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -80,6 +80,7 @@ INSTALLED_APPS = [
     "debug_toolbar",
     "rest_framework",
     "rest_framework.authtoken",
+    "anymail",
     "django.contrib.admin",
     "django.contrib.auth",
     "django.contrib.contenttypes",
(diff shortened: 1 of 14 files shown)


opencodelists/settings.py
```diff
@@ -259,3 +260,16 @@
...
+ANYMAIL = {
+    "MAILGUN_API_KEY": env.str("MAILGUN_API_KEY", default=None),
+    "MAILGUN_API_URL": "https://api.eu.mailgun.net/v3",
+    "MAILGUN_SENDER_DOMAIN": "mg.opencodelists.opensafely.org",
+}
+EMAIL_BACKEND = env.str(
+    "EMAIL_BACKEND", default="django.core.mail.backends.console.EmailBackend"
+)
+DEFAULT_FROM_EMAIL = "you@example.com"
+SERVER_EMAIL = "your-server@example.com"
```
ghickman, line 275: We should probably set these to something 😁

ghickman (changes requested):
Couple of config values need fixing.

You've added an HTML email template, do you know how Django handles text-only content for this email?

Really happy to see this being added!

opencodelists/settings.py
rebkwok: We should probably also change them on job-server, where I copied the settings from :smile:

rebkwok:
> You've added an HTML email template, do you know how Django handles text-only content for this email?

It's an html template to override the one that the default view uses, but I think it just sends it as text
https://github.com/django/django/blob/main/django/contrib/auth/views.py#L222

opencodelists/settings.py
rebkwok: Are these sensible values?
```
DEFAULT_FROM_EMAIL = "no-reply@opencodelists.org"
SERVER_EMAIL = "tech@opensafely.org"
```

ghickman: I think those make sense 👍

ghickman:
Oh yes, I see, there's an `html_email_template_name` too, perfect 👍

ghickman (approved):

Issue #1111, opened 2022-03-08
Codelists with no published versions show up in search


inglesp:
For instance, [Monoamine Oxidase Inhibitors DM&D](https://www.opencodelists.org/codelist/opensafely/monoamine-oxidase-inhibitors-dmd/) shows up in a [search for "antidepressants"](https://www.opencodelists.org/?q=antidepressants), but is not accessible to anybody outside the opensafely organisation.

Implementation notes for @lucyb:

* [Here's a failing test case](https://gist.github.com/inglesp/c6bf4efc1bd6c2b9b43b0b8ee959c5ed)
* The relevant part of the [data model](https://github.com/opensafely-core/opencodelists/blob/main/codelists/models.py) is:
    * A `Codelist` has one or more `Version`s
    * A `Version` has a status which can be "published", "under review", or "draft"
    * A `Codelist` has one or more `Handle`s, which link a `Codelist` to a `User` or `Organisation`
* Searches are handled by [this view function](https://github.com/opensafely-core/opencodelists/blob/e93142091131443e96b1271f07dacb6143e55457/codelists/views/index.py#L9-L32).
* You probably want to filter `codelists`, defined on [line 30](https://github.com/opensafely-core/opencodelists/blob/e93142091131443e96b1271f07dacb6143e55457/codelists/views/index.py#L30) to remove any codelists with no published versions

lucyb:
Fix has been deployed. "Monoamine Oxidase Inhibitors DM&D" is no longer visible in the search results (and neither are other codelists without a published version).

Can this issue be closed?

inglesp:
> Can this issue be closed?

Yes!

PR #1112, opened 2022-03-10
Display only codelists with a published version

lucyb:
Codelists with no published versions are currently being returned in
search results, but these are not accessible to anyone outside of the
OpenSafely organisation.

An example is given in [Github Issue #1111](https://github.com/opensafely-core/opencodelists/issues/1111)

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -157,6 +157,9 @@ class Codelist(models.Model):
 
         return self.visible_versions(user).first()
 
+    def has_published_versions(self):
+        return self.versions.filter(status="published").exists()
+
 
 class Handle(models.Model):
     codelist = models.ForeignKey(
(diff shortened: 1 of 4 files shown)


codelists/models.py
```diff
@@ -157,6 +157,9 @@ def latest_visible_version(self, user):
 
         return self.visible_versions(user).first()
 
+    def has_visible_versions(self):
+        return True if self.versions.filter(status="published") else False
```
inglesp, line 161: This would be more idiomatic:

```
return self.versions.filter(status="published").exists()
```

Alternatively, if `.exists()` didn't, um, exist, then you could do

```
return bool(self.versions.filter(status="published"))
```

which I think is more idiomatic than the ternary conditional.

inglesp (approved):
One small naming change, and one small suggestion for style.

Apart from that: great!

Please could you fix the name, and consider the style suggestion.  Then amend the single commit (or rebase), and merge.

It should deploy automatically, although I don't think this is logged anywhere -- so five minutes after merging, can you check that the bug in the original issue is fixed, and then close that issue.

codelists/models.py
lucyb: Oh, of course that's much neater! 

I've amended the commit with this change and the renaming of functions suggested below.

Issue #1128, opened 2022-03-18
Prevent/restrict owner changes on published codelists?


rebkwok:
Metadata can be updated for codelists that are either under review or published.  This includes changing the owner of a codelist, which could potentially cause problems when a codelist is owned by an organisation. 

E.g.
1) User1 creates and publishes a codelist "foo" under the OpenSAFELY organisation
2) User1 creates a new draft
3) User2 is a member of the OpenSAFELY organisation, so can also edit the codelist.  User2 changes the owner of the codelist to themselves.
4) Now the codelist which was available at `codelists/opensafely/foo` is no longer accessible
5) User1 can still see the draft they created, but the codelist itself  is now owned by User2.  User1 can still update the draft, put it into review, and publish it.

I don't think that anyone in an organisation should be able to arbitratily change the owner of a codelist that can be shared by others - possibly not the slug either (basically anything that makes up the canonical URL).  I'd propose that after first publication, the owner/name/slug for an organisation-owned codelist should be locked (or perhaps restricted to organisation admins), so if another user has a URL for a codelist, it might point to a different version in future, but it shouldn't become invalid (the URL issue per se might be handled by codelist DOIs in future, so we might be able to allow name/slug changes, but the owner controls access, so shouldn't change).

inglesp:
> Now the codelist which was available at `codelists/opensafely/foo` is no longer accessible

Changing the owner of a codelist creates a new `Handle`, but the old `Handle` still exists -- so requests to codelists/opensafely/foo will redirect to codelists/user2/foo.

There's still an edge case about people taking ownership of organisation-owned codelists when there is a draft that's owned by another member of an organisation.  I'm not sure what we should do here.

rebkwok:
OK, I didn't realise the old handles would redirect.  In that case, I guess this is not an issue, and I'd expect that changing ownership (especially from org to user) isn't something that would happen often

PR #1129, opened 2022-03-18
Maintain author on all codelist versions past draft status

### LINKED ISSUE #1117: Codelists owned by an organisation disappear when a user saves them for review
Users who create codelists under an organisation owner should still be able to see/edit those codelists once they are save for review.  Currently they only see codelists that are in draft.

Steps to reproduce:
1) As a user who belongs to an organisation, create a codelist and select the organisation as the owner
2) Save the codelist as draft; it appears under "Your drafts"
3) Save the codelist for review; it is no longer shown under "Your drafts", but also doesn't appear under "Your codelists"
(Note that a codelist with the user as owner moves from "Your drafts" to "Your codelists" when saved for review)

A `CodelistVersion` for a draft codelist has a `draft_owner`.  When the codelist is moved to "under review", the `draft_owner` is set to None.  If this was maintained when moved to the "under review" status, we could easily continue to list "under review" codelists for a user.

rebkwok:
Fixes #1117 

Changes `draft_owner` on a `CodelistVersion` to `author` (via a series of migrations).
We now maintain the author on a codelist version so we: 
a) have a record of who created a version 
b) allow users to continue to see codelists that they've created for an organisation after they put them into under-review
c) Let users see a list of organisation codelist that they've created or edited in the past
This also involved updating some code where the presence of a `draft_owner` was used to determine if a version was in draft (now uses the already available `in_draft` property instead).  

Previously, the `draft_owner` was changed to None when the codelist was saved for review.  This was fine if the codelist was owned by the user, but if it was owned by an organisation, it disappeared from the user's "Your codelists" page.

The "Your Codelists" page now also splits visible codelists by user-owned, org-owned, under review and draft.

Note that there can be some weirdness if a codelist was to be authored by a user for an organisation,  published, and then changed to be owned by a different user (see #1128).  I've not addressed this here as I'm assuming it's unlikely to happen, and I think the right thing to do is to prevent codelists from being changed to user-specific ownership after publication anyway.


codelists/migrations/0050_set_codelistversion_author.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def set_author_from_draft_owner(apps, schema_editor):
+    CodelistVersion = apps.get_model("codelists", "CodelistVersion")
+    versions_with_draft_owner = CodelistVersion.objects.filter(
+        draft_owner__isnull=False
+    )
+    for version in versions_with_draft_owner:
+        version.author = version.draft_owner
+        version.save()
```
ghickman, line 13: Is it worth using an `.update()` call here?

ghickman (approved):

codelists/migrations/0050_set_codelistversion_author.py
rebkwok: It's not slow, but it could

ghickman: Yeah I imagine we don't have anywhere near enough versions to see perf gains here!

rebkwok: I added it anyway, just because :slightly_smiling_face:

opencodelists/models.py
```diff
@@ -110,12 +110,28 @@ def is_admin_member(self, organisation):
...
+        """Codelists owned by user"""
+        return Codelist.objects.filter(handles__is_current=True, handles__user=self)
 
     @property
     def drafts(self):
+        """Codelist versions authored by user and in draft"""
         return self.versions.filter(status=Status.DRAFT)
 
+    @property
+    def reviews(self):
```
inglesp, line 122: I think this is a slightly misleading name, and I'd prefer `versions_under_review`.  This is because we might introduce the concept of something like a code review in the future.

opencodelists/views/user.py
```diff
@@ -1,19 +1,40 @@
...
-        "codelists": codelists.order_by("handles__name"),
-        "drafts": user.drafts.select_related("codelist").order_by(
-            "codelist__handles__name"
+        "codelists": owned_codelists.order_by("handles__name"),
+        # note that name is a property on a codelist, not an attribute, and it comes from the current handle.
+        # If we want to order codelists or versions by codelist name, we actually need to order them by handle name.
+        # We can't use a queryset order_by in the following cases (where reviews/versions are querysets of
+        # CodelistVersion instances), as a codelist can have multiple versions and multiple handles,
+        # and this results in duplicates in the returned queryset.
+        # See https://code.djangoproject.com/ticket/18165
```
inglesp, line 28: Exemplary comment!

opencodelists/tests/views/test_user.py
```diff
@@ -0,0 +1,118 @@
...
+    update_codelist,
+)
+from opencodelists.actions import add_user_to_organisation
+from opencodelists.models import User
+
+
+def test_get(client, organisation_user):
+    user = User.objects.create(
+        username="test", name="test", email="test@test.com", password="test"
+    )
```
inglesp, line 16: I'd prefer it if we created a `user_with_no_codelists` fixture, instead of creating it here.

opencodelists/tests/views/test_user.py
```diff
@@ -0,0 +1,118 @@
...
+    for codelist_category in [
+        "codelists",
+        "authored_for_organisation",
+        "under_review",
+        "drafts",
+    ]:
+        assert not response.context[codelist_category]
+
+
+def test_user_codelists(client, organisation):
```
inglesp, line 31: Can we piggyback off existing fixtures here?

inglesp (commented):
This is great, thank you!  Have made a handful of small suggestions, but feel free to ignore if you've moved on to other things.

PR #1132, opened 2022-03-18
Fix bugs with discarding drafts

### LINKED ISSUE #867: Handle.DoesNotExist
[Sentry1](https://sentry.io/organizations/ebm-datalab/issues/2503446649/?project=5248013&query=is%3Aunresolved)

[Sentry2](https://sentry.io/organizations/ebm-datalab/issues/2502181784/?project=5248013&query=is%3Aunresolved)

### LINKED ISSUE #868: CodelistVersion.cached_hierarchy.RelatedObjectDoesNotExist
[Sentry](https://sentry.io/organizations/ebm-datalab/issues/2445693249/?project=5248013&query=is%3Aunresolved)

### LINKED ISSUE #922: Server error (500) when discarding draft codelist
Moved from https://github.com/opensafely/documentation/issues/439.

@rmgpanw:
> I get a webpage saying 'server error (500)' when I discard a draft codelist. The codelist draft is successfully removed, but ideally there wouldn't be the error page? Thanks

rebkwok:
Discarding a draft was resulting in a 500 error to the user, although the draft was successfully deleted.

In the case where a draft was the sole version for a codelist, the codelist itself was also deleted, and the error was caused by an attempt to redirect to the non-existed codelist URL (`Handle.DoesNotExist` error)
Fixes #867 

In the case where there were multiple drafts, the error was caused by an attempt to cache the hierarchy on a discarded version (`CodelistVersion.cached_hierarchy.RelatedObjectDoesNotExist`)
Fixes #868
Fixes #922

### DIFF
--- a/builder/views.py
+++ b/builder/views.py
@@ -4,6 +4,7 @@ from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, redirect, render
+from django.urls import reverse
 from django.views.decorators.http import require_http_methods
 
 from codelists.search import do_search
@@ -45,7 +46,10 @@ def _handle_post(request, draft):
     elif action == "discard":
         messages.add_message(request, messages.INFO, "Your changes have been discarded")
         actions.discard_draft(draft=draft)
-        return redirect(draft.codelist)
+        # check if the discarded draft's codelist still exists (i.e. this was not the only version)
+        if draft.codelist.id is not None:
+            return redirect(draft.codelist)
+        return redirect(reverse("user", args=(request.user.username,)))
     else:
         return HttpResponse(status=400)
 
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -313,6 +313,13 @@ class CodelistVersion(models.Model):
         kwargs["other_tag_or_hash"] = other_clv.tag_or_hash
         return reverse(f"codelists:{self.codelist_type}_version_diff", kwargs=kwargs)
 
+    def exists(self):
+        """
+        Check that a CodelistVersion exists.  It may not, in the event that it's just
+        been discarded
+        """
+        return CodelistVersion.objects.filter(id=self.id).exists()
+
     @property
     def url_kwargs(self):
         kwargs = self.codelist.url_kwargs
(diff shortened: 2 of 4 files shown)


rebkwok:
Note to self: this will conflict with #1129; whichever one gets merged second will need rebasing

builder/views.py
```diff
@@ -45,7 +46,10 @@ def _handle_post(request, draft):
     elif action == "discard":
         messages.add_message(request, messages.INFO, "Your changes have been discarded")
         actions.discard_draft(draft=draft)
-        return redirect(draft.codelist)
+        # check if the discarded draft's codelist still exists (i.e. this was not the only version)
+        if draft.codelist.id is not None:
```
ghickman, line 50: Ooh, I'd not thought of checking the ID before!  Could you optimise this further by using `draft.codelist_id` or would that not have been updated because it's on the now-deleted Draft instance?

ghickman (approved):

builder/views.py
rebkwok: Hmm.  I don't know - I'll try and see

rebkwok: Nope, apparently you can't.  `codelist_id` is still not None at this point.

codelists/models.py
```diff
@@ -313,6 +313,13 @@ def get_diff_url(self, other_clv):
         kwargs["other_tag_or_hash"] = other_clv.tag_or_hash
         return reverse(f"codelists:{self.codelist_type}_version_diff", kwargs=kwargs)
 
+    def exists(self):
```
inglesp, line 316: Would prefer it if this was called `has_any_versions`.  It's a bit long-winded, but it communicates the intent more clearly.

inglesp: Also: I can't see where this is used -- am I missing something?

inglesp (commented):
Great stuff -- just one query about an apparently-unused new method.

codelists/models.py
rebkwok: This method is on a `CodelistVersion` though, so `has_any_versions` doesn't seem like the right name.  It's used here: https://github.com/opensafely-core/opencodelists/blob/main/builder/decorators.py#L20

inglesp: Oh sorry, my mistake!

codelists/models.py
```diff
@@ -313,6 +313,13 @@ def get_diff_url(self, other_clv):
...
+        """
+        Check that a CodelistVersion exists.  It may not, in the event that it's just
```
inglesp, line 318: Might be clearer if this said "Check that this CodelistVersion exists in the database".

PR #1133, opened 2022-03-21
Give the app a bit longer to start up

rebkwok:


iaindillingham (approved):
Why does the app need 20 more (seconds?) to start up?

rebkwok:
> Why does the app need 20 more (seconds?) to start up?

30 seconds is a bit arbitrary and is just what job-server users.  It needed increased because 10s wasn't always long enough if it had to run migrations.  This one took 3 reruns before it managed to get the migrations done quickly enough to complete deployment.
https://github.com/opensafely-core/opencodelists/actions/runs/2015184619

iaindillingham:
Got it, thank you!

inglesp:
Sigh!  Could we add a comment about where 30s has come from, and why we think it's enough?

PR #1134, opened 2022-03-21
Remove unused user/auth code and add some admin fields 

### LINKED ISSUE #1107: Remove old user activation and set password views
Users can now create their own accounts, and change and reset passwords (using the default django auth views), so we should be able to remove the old activation and password views.

rebkwok:
1) Remove some now unused code around user creation, activation and password setting.

Fixes #1107 
Originally users could not create their own accounts - instead an account would be created for them (inactive by default), an organisation admin would go to the activation url to activate them and generate a set password url, which they'd send to the user.  Now, users can register to create their own accounts, and the default django auth views are used for password changes and resets.

2) Add some useful extra admin fields
- users' memberships on the User list view
- member counts on the Organisation list view
- membership inlines on change forms for both


ghickman (approved):

codelists/tests/views/assertions.py
```diff
@@ -18,8 +18,8 @@ def assert_post_unauthenticated(client, url):
 
 
 def assert_unauthorised(client, method, url):
-    user = create_user(
-        username="ursula", name="Ursula", email="ursula@test.ac.uk", is_active=True
```
inglesp, line 22: I may have been watching a lot of the Little Mermaid when I wrote this test.

inglesp (commented):
`+69 −363`

:100:

PR #1137, opened 2022-03-22
Various updates from recent PRs

rebkwok:
Post-merge updates from 
#1132 
#1133 
#1129

### DIFF
--- a/CHECKS
+++ b/CHECKS
@@ -1,3 +1,4 @@
+# Wait for 30 seconds to allow time for migrations, if necessary
 WAIT=30
 TIMEOUT=60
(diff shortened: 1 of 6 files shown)


inglesp (approved):
Thanks, I appreciate you taking the time to revisit these!

CHECKS
```diff
@@ -1,3 +1,4 @@
+# Wait for 30 seconds to allow time for migrations, if necessary
```
rebkwok, line 1: I'm not 100% sure dokku will like the comments in this file, but I've no way of checking that I can think of

PR #1138, opened 2022-03-22
Organisation members can see organisation-owned under-review codelists

rebkwok:
Now that organisation members have a "My organisations" page, we can give them a list of both published and under-review codelists.  This means that any organisation-owned under-review codelists that don't have authors (either because they predate the author field, or because they were uploaded via the API) can be accessed easily by members of that organisation.

### DIFF
--- a/codelists/views/index.py
+++ b/codelists/views/index.py
@@ -3,10 +3,10 @@ from django.shortcuts import get_object_or_404, render
 
 from opencodelists.models import Organisation
 
-from ..models import Handle
+from ..models import Handle, Status
 
 
-def index(request, organisation_slug=None):
+def index(request, organisation_slug=None, status=Status.PUBLISHED):
     handles = Handle.objects.filter(is_current=True)
 
     q = request.GET.get("q")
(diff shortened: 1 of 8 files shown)


codelists/views/index.py
```diff
@@ -38,3 +44,13 @@ def _all_published_codelists(handles):
...
     ]
+
+
+def _all_under_review_codelist_versions(handles):
+    return sum(
+        [
+            list(handle.codelist.versions.filter(status=Status.UNDER_REVIEW))
+            for handle in handles
+        ],
+        [],
```
inglesp, line 55: Could you create something like `list_utils.flatten`, and call that here?

codelists/views/index.py
```diff
@@ -27,9 +27,15 @@ def index(request, organisation_slug=None):
...
-    codelists = _all_published_codelists(handles)
-    ctx = {"codelists": codelists, "organisation": organisation, "q": q}
-    return render(request, "codelists/index.html", ctx)
+    if status == Status.PUBLISHED:
+        codelists = _all_published_codelists(handles)
+        ctx = {"codelists": codelists, "organisation": organisation, "q": q}
+        return render(request, "codelists/index.html", ctx)
+    else:
+        assert status == Status.UNDER_REVIEW
+        codelists = _all_under_review_codelist_versions(handles)
```
inglesp, line 36: Could you rename `codelists` to `versions`?

inglesp (approved):
Great.

PR #1140, opened 2022-03-23
Show versions on builder page, and latest published version on user codelists page

### LINKED ISSUE #1127: User can't view published codelist if they have a version in draft
The default view for a codelist returns the latest visible version for that user.  If the user has a version that's in draft, that is returned as the latest visible version, in the builder view.  However, as there are no links to other versions on the builder view, a user with a new draft version has no way to see the current published version (without logging out and searching for it, and then it'll only be visible if it's an OpenSAFELY-owned codelist).

I think there are two possible solutions:
1) Continue to return the lastest draft version for the user as the default, if there is one, but add links to other versions on the builder page
2) Make the default version returned for a codelist the latest published version (which will contain links to drafts/reviews if the user has any); only return draft or review versions if there is no published one

On the users "My codelists" page, we list their published codelists and codelists in draft.  If there's both a published and draft version of the same codelist for a user, it appears in both sections.  The urls for published codelists are `/codelist/<owner>/<codelist_id>` and for drafts `codelists/<owner>/<codelist_id>/<version_id>` but currently they'll take you to the same builder page for the draft.

rebkwok:
Fixes #1127 

- Show versions on the builder page
The code for this was there already, but was only showing up if a draft had searches.

- Use latest published version URLs on user's codelists page
The `Codelist.get_absolute_url` returns the latest visible version for that user, but when a codelist has both a published version and a version in draft or under review, we expect the link under "My codelists" to show the latest published version, and the link under "My codelists under review/My drafts" to show the relevant unpublished version.

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -1,3 +1,4 @@
+from django.contrib.auth.models import AnonymousUser
 from django.db import models
 from django.urls import reverse
 from django.utils.functional import cached_property
@@ -83,6 +84,16 @@ class Codelist(models.Model):
             f"codelists:{self.codelist_type}_codelist", kwargs=self.url_kwargs
         )
 
+    def get_latest_published_url(self):
+        """
+        Return the URL for the latest published version, or if no such version exists,
+        return the absolute URL for this codelist, which will redirect to the user's latest
+        visible version
+        """
+        if self.latest_published_version():
+            return self.latest_published_version().get_absolute_url()
+        return self.get_absolute_url()
+
     def get_update_url(self):
         return reverse(
             f"codelists:{self.codelist_type}_codelist_update", kwargs=self.url_kwargs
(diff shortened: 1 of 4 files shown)


static/src/js/builder/codelistbuilder.jsx
```diff
@@ -189,16 +189,12 @@ class CodelistBuilder extends React.Component {
...
-                  ))}
-                </ul>
-              </>
-            )}
+            <h6>Versions</h6>
+            <ul className="pl-3">
+              {this.props.versions.map((version) => (
+                <Version key={version.hash} version={version} />
+              ))}
+            </ul>
```
inglesp, line 197: Does this match what we do on the non-builder (need a better name for that...) page?  That is, if there's only one version, do we show a list of versions?

codelists/models.py
```diff
@@ -157,6 +168,12 @@ def latest_visible_version(self, user):
 
         return self.visible_versions(user).first()
 
+    def latest_published_version(self):
```
inglesp, line 171: I think "published" and "visible" mean slightly different things, so could you rename this to `latest_visible_version`?

inglesp (approved):

codelists/models.py
rebkwok: There's already a `latest_visible_version`, which shows the last visible version for the user, and defaults to the last version with status "published" if the user is anonymous.  This method is intended to return the last published version, even if there's a non-published version that would be visible for this user.

static/src/js/builder/codelistbuilder.jsx
rebkwok: yes, we do

PR #1145, opened 2022-03-24
Delete searches

### LINKED ISSUE #358: Term Search Delete
It would be helpful if term searches could be deleted when building a codelists if you decide they aren't necessary.

For example when building an [antiplatelet](https://codelists.opensafely.org/builder/brian-mackenna/antiplatelet-medication/search/0209/) codelist  I entered `0209` and got ostomy bag. I know for definite then that my first term `antiplatelet` identified everything and I do not need anything from `0209` term. It would be preferable to delete 9~this is simple example) but only way I can see to get rid is to manually exclude everything irrelevant from `0209`

image

rebkwok:
Makes searches deletable.

Code are deleted from the codelist along with the search only if:
1) they only belong to this search (i.e. they aren't returned by any other searches)
2) they are not included
3) they do not have any ancestors that are included

Fixes #358

### DIFF
--- a/builder/actions.py
+++ b/builder/actions.py
@@ -68,17 +68,12 @@ def delete_search(*, search):
     hierarchy = search.version.hierarchy
     codes_to_keep = set()
 
+    codes_to_keep = []
     for code_obj in search_only_code_objs:
-        # Keep a code if it is explicitly included or a descendant of an explicitly included code
-        if code_obj.code in all_included_codes:
-            codes_to_keep.add(code_obj.code)
-            codes_to_keep = codes_to_keep | hierarchy.descendants(code_obj.code)
-        else:
-            # if it's not explicitly included, we still need to keep it if any of its parents are included
-            ancestors = hierarchy.ancestors(code_obj.code)
-            ancestors_in_included = ancestors & all_included_codes
-            if ancestors_in_included:
-                codes_to_keep.add(code_obj.code)
+        ancestors = hierarchy.ancestors(code_obj.code)
+        ancestors_in_included = ancestors & all_included_codes
+        if ancestors_in_included or code_obj.code in all_included_codes:
+            codes_to_keep.append(code_obj.code)
 
     # Delete any code objs that belong to this search only, and are not in the codes_to_keep set
     search_only_code_objs.exclude(code__in=codes_to_keep).delete()
--- a/builder/actions.py
+++ b/builder/actions.py
@@ -51,10 +51,37 @@ def delete_search(*, search):
     # Grab the PK before we delete the instance
     search_pk = search.pk
 
-    # Delete any codes that only belong to this search
-    search.version.code_objs.annotate(num_results=Count("results")).filter(
-        results__search=search, num_results=1
-    ).delete()
+    # Delete any codes that:
+    # - only belong to this search
+    # - are not included
+    # - are not descendants of an included code
+
+    # Find all code objs that belong to this search and no others
+    search_only_code_objs = search.version.code_objs.annotate(
+        num_results=Count("results")
+    ).filter(results__search=search, num_results=1)
+    # Get all explicitly included codes on the codelist
+    all_included_codes = search.version.codeset.codes()
+
+    # Build a set of codes_to_keep from this search, consisting of any included codes, their descendants,
+    # and any codes with ancestor that is included
+    hierarchy = search.version.hierarchy
+    codes_to_keep = set()
+
+    for code_obj in search_only_code_objs:
+        # Keep a code if it is explicitly included or a descendant of an explicitly included code
+        if code_obj.code in all_included_codes:
+            codes_to_keep.add(code_obj.code)
+            codes_to_keep = codes_to_keep | hierarchy.descendants(code_obj.code)
+        else:
+            # if it's not explicitly included, we still need to keep it if any of its parents are included
+            ancestors = hierarchy.ancestors(code_obj.code)
+            ancestors_in_included = ancestors & all_included_codes
+            if ancestors_in_included:
+                codes_to_keep.add(code_obj.code)
+
+    # Delete any code objs that belong to this search only, and are not in the codes_to_keep set
+    search_only_code_objs.exclude(code__in=codes_to_keep).delete()
 
     # Delete the search
     search.delete()
--- a/builder/tests/test_actions.py
+++ b/builder/tests/test_actions.py
@@ -90,60 +90,116 @@ def test_duplicate_search(version_from_scratch):
     assert draft.code_objs.count() == 2
 
 
-@pytest.mark.xfail
-def test_delete_search():
-    pass
-    # # Arrange: create a draft with codes and a search
-    # codelist = CodelistFactory()
-    # owner = UserFactory()
-    # draft = actions.create_draft_with_codes(
-    #     codelist=codelist,
-    #     owner=owner,
-    #     codes=["1067731000000107", "1068181000000106"],
-    # )
-    # s = actions.create_search(
-    #     draft=draft, term="synchronised", codes=["1068181000000106"]
-    # )
+def test_delete_search(version_from_scratch):
 
-    # # Act: delete the search
-    # actions.delete_search(search=s)
+    draft = version_from_scratch
 
-    # # Assert...
-    # # that the codelist has 0 searches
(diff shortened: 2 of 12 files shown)


builder/tests/test_actions.py
```diff
@@ -90,60 +90,116 @@ def test_duplicate_search(version_from_scratch):
...
+
+    # Act: delete the search for "epicondylitis"
+    actions.delete_search(search=s1)
+
+    # Assert...
+    # that the codelist has no searches
+    assert draft.searches.count() == 0
+    # This search matched 2 starting codes which are now deleted.  Others codes that were not
+    # associated with the search remain
+    assert draft.code_objs.count() == starting_code_count - 2
```
rebkwok, line 202: Is this what should happen?  The 2 epicondylitis codes were on the codelist when it had no searches, the search means that they become associated with the search, so when you delete it, those 2 codes get deleted.

Should the delete action check the status of the `code_obj` and only delete it if it's unresolved?

inglesp:
When deleting a `Search`, I think the least surprising behaviour is to delete any of the related `CodeObj`s that satisfy all three of:

(a) are not related to other `Search` objects
(b) are not included
(c) are not descendants of an included `CodeObj`

For instance, suppose we have a `CodelistVersion` with the following `CodeObj`s:

```
  a   b
 / \ / \
c   d   e
```

where codes in the first row are parents of codes in the second row.

Suppose we also we have two searches, `Sa` (which matches `a`) and `Sb` (which matches `b`).

If we delete `Sb`, then we should not delete any `a`, `c`, `d`, because they are related to `Sa`.

But what we do with `b` or `e`?

Here's a table indicating whether they should be deleted or not:

| b | e   | delete b? | delete e? |
|---|-----|-----------|-----------|
| + | (+) | no        | no        |
| + | +   | no        | no        |
| + | -   | no        | no        |
| ? | +   | yes       | no        |
| ? | -   | yes       | yes       |
| - | (-) | yes       | yes       |
| - | -   | yes       | yes       |
| - | +   | yes       | no        |

This is all a bit fiddly to reason about, so I'd be interested in a second opinion!

rebkwok:
After I put the frontend part in and played about with deleting a bit, I think that (a) is already covered.  I made this codelist with 3 searches on asthma, acute asthma and asthma in pregnancy.
Screenshot from 2022-03-24 14-57-45

Asthma is the parent of the other two, but deleting it leave the codes related to the other two in place.
Screenshot from 2022-03-24 15-06-33

Keeping code objs that are included/descendents of included makes sense.  That should also handle the ones that are imported from csv (or whatever) not via the builder, as they'd be marked as included already when a new draft version is created.

rebkwok:
> After I put the frontend part in and played about with deleting a bit, I think that (a) is already covered.
Of course, this is expected, the action checks for codes that only belong to the search being deleted, and only deletes those.  I think I misinterpreted what[ this filter ](https://github.com/opensafely-core/opencodelists/blob/main/builder/actions.py#L55) was doing.

builder/actions.py
```diff
@@ -51,10 +51,37 @@ def delete_search(*, search):
...
+    all_included_codes = search.version.codeset.codes()
+
+    # Build a set of codes_to_keep from this search, consisting of any included codes, their descendants,
+    # and any codes with ancestor that is included
+    hierarchy = search.version.hierarchy
+    codes_to_keep = set()
+
+    for code_obj in search_only_code_objs:
+        # Keep a code if it is explicitly included or a descendant of an explicitly included code
+        if code_obj.code in all_included_codes:
```
inglesp, line 73: I'm fairly sure we don't need this branch, and that the following will do what we need:

```
    codes_to_keep = []
    for code_obj in search_only_code_objs:
        ancestors = hierarchy.ancestors(code_obj.code)
        ancestors_in_included = ancestors & all_included_codes
        if ancestors_in_included:
            codes_to_keep.append(code_obj.code)
```

builder/tests/test_actions.py
```diff
@@ -90,60 +90,175 @@ def test_duplicate_search(version_from_scratch):
...
+    assert draft.searches.count() == 0
+    # This search matched 2 codes, one included, one not included, but a descendant of an included
+    # code, they are not deleted
+    assert draft.code_objs.count() == len(codelist_codes)
+
+    # Arrange: create new search for not-included code
+    s3 = actions.create_search(
+        draft=draft,
+        term="arthritis of elbow",
+        codes={arthritis_of_elbow_code},
```
inglesp, line 251: This isn't quite right, although it's almost certainly fine.

The view code that creates a search passes all the codes that are found by [`codelists.search.do_search`](https://github.com/opensafely-core/opencodelists/blob/main/codelists/search.py) -- so not just the codes that match, but their descendants too.

https://github.com/opensafely-core/opencodelists/blob/5e46909d7ae2b8e2a112435029852978200281d4/builder/views.py#L212-L214

So in this case, the `codes` argument should also include the "lateral epicondylitis" code, since that's a descendant of "arthritis of elbow".

builder/tests/test_actions.py
```diff
@@ -90,60 +90,175 @@ def test_duplicate_search(version_from_scratch):
...
+    arthritis_of_elbow_code = "439656005"
+
+    assert epicondylitis_included_code in included_codes
+    assert epicondylitis_excluded_code not in included_codes
+    for code in epicondylitis_codes:
+        assert code in codelist_codes
+    assert tennis_toe_code not in codelist_codes
+    assert tennis_toe_code not in included_codes
+    assert arthritis_of_elbow_code in codelist_codes
+    assert arthritis_of_elbow_code not in included_codes
```
inglesp, line 196: I'm sorry you've had to do this!  Do you think it'd be worth spending time to produce some fixtures that are more legible?

builder/views.py
```diff
@@ -217,3 +219,17 @@ def new_search(request, draft):
...
+@login_required
+@require_http_methods(["POST"])
+@load_draft
+@require_permission
+def delete_search(request, draft, search_slug):
+    search = get_object_or_404(Search, version=draft, slug=search_slug)
+    actions.delete_search(search=search)
+    messages.info(
+        request,
+        f'Search for "{search_slug}" deleted.  Any included codes and their descendants have been maintained.',
```
inglesp, line 233: I would change this to something like:

> Search for "{search_slug}" deleted.  Any codes matching "{search_slug}" that do not match any other search term and which are not included in the codelist have been removed

This is not strictly accurate, since it doesn't mention not deleting codes which are descendants of included codes, but I think it gets the balance about right between accuracy and comprehensibility.

inglesp (approved):
Well done, this is really tricky to think about.

builder/actions.py
rebkwok: Almost - it needed to be:
```
    codes_to_keep = []
    for code_obj in search_only_code_objs:
        ancestors = hierarchy.ancestors(code_obj.code)
        ancestors_in_included = ancestors & all_included_codes
        if ancestors_in_included or code_obj.code in all_included_codes:
            codes_to_keep.append(code_obj.code)
```
I've updated this, and also added another test since the previous one didn't catch that extra `or code_obj.code in all_included_codes` that was needed.

builder/tests/test_actions.py
rebkwok: updated

Auselan:
Just wanted to thank you for working on this. I notice the current behaviour is that when a search term is deleted, the orphaned codes appear under [no search term] which is fine.

I just want to backtrack a little bit here, and explore the use-case for this function. Have we defined what we are actually trying to achieve?  To me, the problem that I faced was that if I searched for a keywords I sometimes discovered that it took me too long to deal with the consequences.  For example, if my keyword picks up several hundred Inactive codes, as their relationships have been lost during inactivation, I have to manually click on each and every single one.

That being my problem, the functionality gap is actually about how to rapidly deal with large keyword search results. Do we have a separate thread running on that or shall I elaborate here?

I'm also keen to note the other issue for me, which is the collaborative aspect. If someone asks me to collaborate on a list but gives me either a code definition, or even worse an extensional list, I then have to add search terms or browse the hierarchies myself in order to create my own list for comparison.  In an ideal world, I would like to know which search terms were used by the person generating the code list. I think that means that I want to see which terms were searched for, even if they were then decided to be irrelevant. Not quite sure of the best way to record that, but deleting the search term for me is about shredding a misspelt word.

PR #1161, opened 2022-03-30
Update footer styles and links

tomodwyer:
...to include Bennett and Oxford links.

This also makes the footer stick to the bottom of the screen, when the main content does not fill the screen height.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -238,7 +238,7 @@
 
         <div class="row">
           <div class="col-12 offset-md-1 col-md-10 offset-lg-2 col-lg-8 text-center mt-3 mb-0 small">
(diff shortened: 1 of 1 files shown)


templates/base.html
```diff
@@ -206,10 +206,43 @@
...
+          <li class="list-inline-item mb-3 mb-md-0">
+            <a href="https://www.ox.ac.uk/">
+              University of Oxford
+            </a>
+          </li>
+        </ul>
+
+        <div class="row">
+          <div class="col-12 offset-md-1 col-md-10 offset-lg-2 col-lg-8 text-center mt-3 mb-0 small">
+            <p>© University of Oxford for the Bennett Institute for Applied Data Science 2022. This work may be copied freely for non-commercial research and study. If you wish to do any of the other acts restricted by the copyright you should apply in writing to <a href="mailto:ebmdatalab@phc.ox.ac.uk">ebmdatalab@phc.ox.ac.uk</a>.</p>
```
rebkwok, line 241: Should we have a different email address than `ebmdatalab@`?

The "get in touch" link on the homepage goes to the opensafely contact page, which has `team@opensafely.org`, should we use that here too?

rebkwok (approved):
One question about email addresses

templates/base.html
tomodwyer: Good point! I will use the opensafely.org email.

PR #1173, opened 2022-04-06
Add coding system docs

inglesp:


coding_systems/snomedct/README
```diff
@@ -0,0 +1,25 @@
...
+New releases are published irregularly -- there was a schedule, but it's been disrupted
+by covid.
+
+To import the data, unzip the zip file and run:
+
+        ./manage.py import_data coding_systems.snomedct path/to/dir/SnomedCT_InternationalRF2_PRODUCTION_*
+        ./manage.py import_data coding_systems.snomedct path/to/dir/SnomedCT_UKClinicalRF2_PRODUCTION_*
+        ./manage.py import_data coding_systems.snomedct path/to/dir/SnomedCT_UKEditionRF2_PRODUCTION_*
+
+[0] https://isd.digital.nhs.uk/trud/users/authenticated/filters/0/categories/26/items/101/releases
```
rebkwok, line 25: I feel like I'm being stupid, but I can't see where on this page you actually download the data (same goes for the DM+D TRUD link).  The link also goes to a releases page for the UK Clinical Edition - I think that includes the international and UK clinical extension, but maybe it'd be good to clarify.  

I signed up for an account in order to view the page - should we be using a specific account, or is that how you'd expect someone else to do it?  Do you have to subscribe in order to have download options?

inglesp: Good point, thanks.  Have updated the instructions -- I hope they're clearer.

rebkwok (approved):
Thanks, this is really helpful.  Maybe worth adding a comment in `DEVELOPERS.md` to point people at these READMEs too

inglesp:
> Thanks, this is really helpful. Maybe worth adding a comment in `DEVELOPERS.md` to point people at these READMEs too

Thanks, have done.

PR #1178, opened 2022-04-07
Paginate index pages

rebkwok:
We have over 2000 published codelists now, and they all get loaded on the main index page.  This adds pagination for the home page, organisation codelists page, and organisation under-review page.

Numbers are arbitrary; 30 for under-review pages and 15 for codelist pages, because under-review pages only show the codelist name, and codelist pages have a bit of descriptive text that makes them take up more space on the page.

Screenshot from 2022-04-07 14-28-44

### DIFF
--- a/opencodelists/tests/fixtures.py
+++ b/opencodelists/tests/fixtures.py
@@ -118,6 +118,7 @@ from codelists.actions import (
     export_to_builder,
 )
 from codelists.coding_systems import CODING_SYSTEMS
+from codelists.models import Status
 from codelists.search import do_search
 from opencodelists.actions import (
     add_user_to_organisation,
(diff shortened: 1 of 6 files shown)


opencodelists/tests/fixtures.py
```diff
@@ -637,6 +638,27 @@ def new_style_version(universe, request):
     return type(version).objects.get(pk=version.pk)
 
 
+@pytest.fixture
+def create_codelists(organisation):
```
evansd, line 642: This is a just a drive-by comment so feel free to ignore, but I doesn't look like we're actually using the fixture properties here at all (i.e. the closure over `organisation`) because we always supply the `owner` argument. So I think the whole thing could just be a helper function. But maybe you've got plans to use it in other contexts later where the default organisation will be useful?

evansd (approved):
Seems like a _very_ sensible performance improvement to me! I don't know the codebase that well, but I think the change is mostly standard Django pagination code which all looks good to me.

opencodelists/tests/fixtures.py
rebkwok: Yes, I mostly added it to the fixtures like this to be more generally useful later, intending to use the default org in the tests, but then I passed the owner arg explicitly just to make it clearer.

PR #1192, opened 2022-04-13
feat: add opentelemetry

madwort:
* send spans to honeycomb.io
* @madwort  has added the new envvars to the app config on dokku1

### DIFF
--- a/deploy/gunicorn/conf.py
+++ b/deploy/gunicorn/conf.py
@@ -1,10 +1,10 @@
-from services.logging import logging_config_dict
-
 from opentelemetry import trace
 from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
 from opentelemetry.sdk.resources import Resource
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+from services.logging import logging_config_dict
 
 bind = "0.0.0.0:7000"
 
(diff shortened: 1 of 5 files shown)


deploy/gunicorn/conf.py
```diff
@@ -12,3 +18,19 @@
 # Configure log structure
 # http://docs.gunicorn.org/en/stable/settings.html#logconfig-dict
 logconfig_dict = logging_config_dict
+
+
+def post_fork(server, worker):
```
rebkwok, line 23: Can you add a comment about what this is doing?

rebkwok (approved):

deploy/gunicorn/conf.py
```diff
@@ -12,3 +18,24 @@
...
+# in gunicorn's post_fork method in order to instrument our application process, see:
+# https://opentelemetry-python.readthedocs.io/en/latest/examples/fork-process-model/README.html
+def post_fork(server, worker):
+    server.log.info("Worker spawned (pid: %s)", worker.pid)
+
+    resource = Resource.create(attributes={"service.name": "opencodelists"})
+
+    trace.set_tracer_provider(TracerProvider(resource=resource))
+    span_processor = BatchSpanProcessor(
+        OTLPSpanExporter(endpoint="https://api.honeycomb.io")
```
madwort, line 33: this may be duplicated in `OTEL_EXPORTER_OTLP_ENDPOINT` in the environment, I'd like to get it working first & then test removing it. Experimenting with these kind of config tweaks is much easier in production at the moment...!

rebkwok (approved):

PR #1193, opened 2022-04-14
feat: add a robots.txt

madwort:
* ask nicely for scraper bots to go easy on us 
* redirect from the required location opencodelists.org/robots.txt to the static file at opencodelists.org/static/robots.txt

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,0 +1,8 @@
+User-agent: SemrushBot
+Crawl-delay: 60
+
+User-agent: MJ12bot
+Crawl-delay: 60
+
+User-agent: bingbot
+Crawl-delay: 60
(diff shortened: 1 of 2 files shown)


static/robots.txt
```diff
@@ -0,0 +1,8 @@
+User-agent: SemrushBot
+Crawl-delay: 60
+
+User-agent: MJ12bot
+Crawl-delay: 60
+
+User-agent: bingbot
+Crawl-delay: 60
```
inglesp, line 8: Can we ask all bots to be nice, instead of just these?

inglesp (approved):
I know nothing about robots.txt, but if this reduces load on the server then I'm all for it!

static/robots.txt
madwort: yeah, we could address it to `User-agent: *` although it's up to the bots to choose whether to obey it & I think these ones do (from experience with other sites).

PR #1194, opened 2022-04-14
feat: instrument index with opentelemetry

madwort:
* this should add more span detail to the traces currently generated by auto_instrumentation
* see https://docs.honeycomb.io/getting-data-in/opentelemetry/python/ and https://opentelemetry.io/docs/instrumentation/python/cookbook/

### DIFF
--- a/codelists/views/index.py
+++ b/codelists/views/index.py
@@ -1,11 +1,10 @@
 from django.core.paginator import Paginator
 from django.db.models import Q
 from django.shortcuts import get_object_or_404, render
+from opentelemetry import trace
 
 from opencodelists.list_utils import flatten
 from opencodelists.models import Organisation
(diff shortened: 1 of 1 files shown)


codelists/views/index.py
```diff
@@ -9,46 +10,57 @@
 
 
 def index(request, organisation_slug=None, status=Status.PUBLISHED):
-    handles = Handle.objects.filter(is_current=True)
+    tracer = trace.get_tracer(__name__)
```
ghickman, line 13: We probably want this at the top since `__name__` is the module name, that matches how loggers are typically invoked in Python too.

ghickman (approved):
Looks good, just one note on where `tracer` is constructed.

Issue #1189, opened 2022-04-13
Add an option to OpenCodelists to filter by code type


lucyb:
The OpenCodelists API should be extended to allow filtering by codelist type.

For the Interactive tool we'll need to get a list of codelists from within the OpenSafely organisation, which are of a specified type, e.g SNOMED CT.

PR #1200, opened 2022-04-19
Add coding_system_id parameter to codelist list API endpoint

inglesp:
Fixes https://github.com/opensafely-core/opencodelists/issues/1189.

### DIFF
--- a/codelists/api.py
+++ b/codelists/api.py
@@ -29,6 +29,9 @@ def codelists_get(request, owner):
     HTPP response contains JSON array with one item for each codelist owned by the
     organisation.
 
+    request.GET may contain `coding_system_id` parameter, which will be used to filter
+    the returned codelists.
+
     Eg:
 
     [
@@ -36,6 +39,7 @@ def codelists_get(request, owner):
             "full_slug": "opensafely/asthma-diagnosis",
             "slug": "asthma-diagnosis",
             "name": "Asthma Diagnosis",
+            "coding_system_id": "snomedct",
             "versions": [
                 {
                     "hash": "66f08cca",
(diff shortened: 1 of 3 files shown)


codelists/api.py
```diff
@@ -48,15 +48,22 @@ def codelists_get(request, owner):
...
     for cl in sorted(
-        owner.codelists.prefetch_related("handles", "versions"), key=lambda cl: cl.slug
+        owner.codelists.filter(**filter_kwargs).prefetch_related("handles", "versions"),
+        key=lambda cl: cl.slug,
     ):
         record = {
             "full_slug": cl.full_slug(),
             "slug": cl.slug,
             "name": cl.name,
+            "coding_system_id": cl.coding_system_id,
```
lucyb, line 66: Really minor thing, but the comment at the start of this function could be updated with the new parameter details.

inglesp: Good point.  Did a quick fix without engaging brain.

codelists/api.py
```diff
@@ -48,15 +52,22 @@ def codelists_get(request, owner):
...
+    filter_kwargs = {}
+
+    if "coding_system_id" in request.GET:
+        filter_kwargs["coding_system_id"] = request.GET["coding_system_id"]
+
     records = []
 
     for cl in sorted(
-        owner.codelists.prefetch_related("handles", "versions"), key=lambda cl: cl.slug
+        owner.codelists.filter(**filter_kwargs).prefetch_related("handles", "versions"),
```
lucyb, line 63: If someone passed in "coding_system_id": "" would that cause unexpected behaviour here? Might be worth checking that the value is truthy in the if-statement above?

lucyb (approved):

PR #1203, opened 2022-04-20
Add Codelist.is_private, and hide private codelists from search

inglesp:
This is to allow us to hide the automatically converted codelists from
users, because these codelists have not been reviewed or assured.

This is another tiny step towards implementing
https://github.com/opensafely-core/roadmap/discussions/17.


inglesp:
Existing tests pass, and I've verified manually that private codelists don't show up in search results.

But there are no new tests -- the existing behaviour is not well-tested (sorry!) and when we implement [Improve workflow for creating and publishing codelists](https://github.com/opensafely-core/roadmap/discussions/17) we'll want to revisit the testing strategy properly here.

PR #1226, opened 2022-05-05
Handle quotes in CSV files as expected

### LINKED ISSUE #1210: Don't ignore quotes when reading codelists from csv files
When reading codelists from csv files, quotes are ignored:

https://github.com/opensafely-core/opencodelists/blob/8a0cb0caeab57f303e25dc20e3821b901dd7bf65/codelists/forms.py#L96

Unfortunately, this means that a line like this:

```
a,"b,c",d
```

will be read as

```
["a", "b", "c", "d"]
```

rather than as

```
["a", "b,c", "d"]
```

and raise an error.

@alexwalkerepi suggests removing the quoted commas, to work around the issue, meaning the line should look like this:

```
a,"bc",d
```

Reported by @andrewscolm in [this Slack thread](https://bennettoxford.slack.com/archives/C33TWNQ1J/p1651050772425919).

inglesp:
We turned off quote handling in CSV uploading to handle confusing cases
where headers contained extraneous whitespace (see #1065).  However,
that meant that we could not handle CSVs with commas (#1210).

Instead of attempting to handle CSVs with whitespace in headers, we now
raise a ValidationError, which allows us to to handle quotes properly.

Fixes #1210.

### DIFF
--- a/codelists/forms.py
+++ b/codelists/forms.py
@@ -93,8 +93,15 @@ class CSVValidationMixin:
     def clean_csv_data(self):
         data = self.cleaned_data["csv_data"].read().decode("utf-8-sig")
 
-        reader = csv.reader(StringIO(data), quoting=csv.QUOTE_NONE)
+        reader = csv.reader(StringIO(data))
         header = next(reader)  # expected to be headers
+
+        for i, value in enumerate(header):
+            if value != value.strip():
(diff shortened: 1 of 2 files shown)


CarolineMorton (approved):
Looks good to me

PR #1268, opened 2022-06-01
Force new version

CarolineMorton:
This is a small change the API to allow us to force new versions of codelists to be created even if they have not changed from the old version. This is a feature request from PRIMIS.

### DIFF
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -58,7 +58,7 @@ def create_or_update_codelist(
     methodology=None,
     references=None,
     signoffs=None,
-    force_create=False
+    force_create=False,
 ):
     slug = slug or slugify(name)
     references = references or []
@@ -78,10 +78,7 @@ def create_or_update_codelist(
             signoffs=signoffs,
         )
         create_version_with_codes(
-            codelist=codelist,
-            codes=codes,
-            tag=tag,
-            force_create=force_create
+            codelist=codelist, codes=codes, tag=tag, force_create=force_create
         )
 
         return codelist
@@ -212,7 +209,7 @@ def create_version_with_codes(
     hierarchy=None,
     codeset=None,
     author=None,
-    force_create=False
+    force_create=False,
 ):
     """Create a new version of a codelist with given codes.
     Returns the new version, or None if no version is created because the codes are the
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -58,7 +58,7 @@ def create_or_update_codelist(
     methodology=None,
     references=None,
     signoffs=None,
-    force_create=False,
+    always_create_new_version=False,
 ):
     slug = slug or slugify(name)
     references = references or []
@@ -78,7 +78,10 @@ def create_or_update_codelist(
             signoffs=signoffs,
         )
         create_version_with_codes(
-            codelist=codelist, codes=codes, tag=tag, force_create=force_create
+            codelist=codelist,
+            codes=codes,
+            tag=tag,
+            always_create_new_version=always_create_new_version,
         )
 
         return codelist
@@ -209,11 +212,11 @@ def create_version_with_codes(
     hierarchy=None,
     codeset=None,
     author=None,
-    force_create=False,
+    always_create_new_version=False,
 ):
     """Create a new version of a codelist with given codes.
     Returns the new version, or None if no version is created because the codes are the
-    same as the previous version.  (Future work could allow a force_create parameter.)
+    same as the previous version.
     `hierarchy` and `codeset` may be passed in if they have already been calculated.
     """
 
@@ -221,7 +224,7 @@ def create_version_with_codes(
         raise ValueError("No codes")
 
     prev_clv = codelist.versions.order_by("id").last()
-    if force_create is True or set(codes) != set(prev_clv.codes):
+    if always_create_new_version or set(codes) != set(prev_clv.codes):
         return _create_version_with_codes(
             codelist=codelist,
             codes=codes,
@@ -431,14 +434,13 @@ def publish_version(*, version):
 
 
 @transaction.atomic
-def delete_version(*, version, force=False):
+def delete_version(*, version):
     """Delete a version.
     If this is the only version, the codelist will also be deleted.
     Returns a boolean indicating whether the codelist was deleted.
     """
 
-    if not force:
-        assert version.is_under_review
+    assert version.is_under_review
 
     codelist = version.codelist
     if codelist.versions.count() == 1:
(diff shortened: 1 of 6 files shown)


codelists/actions.py
```diff
@@ -210,20 +209,19 @@ def create_version_with_codes(
...
-
     `hierarchy` and `codeset` may be passed in if they have already been calculated.
     """
 
     if not codes:
         raise ValueError("No codes")
 
     prev_clv = codelist.versions.order_by("id").last()
-    if set(codes) == set(prev_clv.codes):
+    if set(codes) == set(prev_clv.codes) and force_create is False:
```
rebkwok, line 224: Minor nit - you could avoid having to do the set comparison if `force_create` is `True`, since in that case we know we're going to have to create a new version anyway

```
if force_create is True or set(codes) != set(prev_clv.codes):
    return _create_version_with_codes(...)
return None
```

codelists/actions.py
```diff
@@ -440,13 +434,11 @@ def publish_version(*, version):
...
 def delete_version(*, version):
     """Delete a version.
-
     If this is the only version, the codelist will also be deleted.
-
     Returns a boolean indicating whether the codelist was deleted.
     """
 
-    assert version.is_under_review
+    # assert version.is_under_review
```
rebkwok, line 441: Should this be commented out?  We don't want to allow published versions to be deleted, do we?

codelists/api.py
```diff
@@ -166,11 +168,17 @@ def versions(request, codelist):
 
     try:
         if "codes" in data:
-            clv = create_version_with_codes(
-                codelist=codelist,
-                codes=set(data["codes"]),
-                tag=data.get("tag"),
-            )
+            if "force_create" in data:
```
rebkwok, line 171: I don't think this needs to be checked, we can just always pass the `force_create` parameter, with a defaul:
```
clv = create_version_with_codes(
                codelist=codelist,
                codes=set(data["codes"]),
                tag=data.get("tag"),
                force_create=data.get("force_create", False)
            )
```

rebkwok (changes requested):
A couple of suggestions and one question
Could you add a test for this too?

inglesp (changes requested):
In addition to Becky's review, I'd like to see a new test for the updated action.

The action is tested by [these tests](https://github.com/opensafely-core/opencodelists/blob/35faed5a9a3777e1aef01b29c8120daff47d694a/codelists/tests/test_actions.py#L121-L176).  There should be a new test (called something like `test_create_or_update_codelist_force_update_no_change_to_codes`) which should be like `test_create_or_update_codelist_update_no_change_to_codes`, except with `force_True`.  See `test_create_or_update_codelist_update` for an example of a test that verifies that a new version has been created.

codelists/actions.py
CarolineMorton: Yes this is commented out as I am trying to create an API endpoint to delete published primis codelists

CarolineMorton: But I will try to find a way to pass it in as an option arg

codelists/actions.py
```diff
@@ -58,6 +58,7 @@ def create_or_update_codelist(
     methodology=None,
     references=None,
     signoffs=None,
+    force_create=False,
```
inglesp, line 61: In a function called `create_or_update_codelist`, a parameter called `force_create` is a bit confusing.  (For instance: if it's true, are we ensuring that we create a new codelist instead of updating an existing one?)

I'd suggest calling it something like `always_create_new_version`.

codelists/actions.py
```diff
@@ -210,36 +209,33 @@ def create_version_with_codes(
     hierarchy=None,
     codeset=None,
     author=None,
+    force_create=False,
 ):
     """Create a new version of a codelist with given codes.
-
     Returns the new version, or None if no version is created because the codes are the
     same as the previous version.  (Future work could allow a force_create parameter.)
```
inglesp, line 216: Oh right, `force_create` was my name.  Please could you update this comment?

codelists/tests/test_actions.py
```diff
@@ -176,6 +176,38 @@ def test_create_or_update_codelist_update_no_change_to_codes(
...
+        owner=organisation,
+        name=new_codelist_name,
+        coding_system_id="snomedct",
+        codes=disorder_of_elbow_codes,
+        description="This is a test",
+        methodology="This is how we did it",
+        references=[{"text": "Some reference", "url": "http://example.com"}],
+        signoffs=[{"user": user.username, "date": "2021-04-21"}],
+    )
+    assert cl.versions.count() == 1
```
inglesp, line 193: Instead of asserting the number of versions here, you could use `with assert_difference(...)` as we do [here](https://github.com/opensafely-core/opencodelists/blob/0bbfc721c24f9e531022c4afe68342d5454b50ad/codelists/tests/test_actions.py#L144).

inglesp (approved):
New test looks good, thanks.

There are a couple of small changes I'd like you to make, but I don't need to review this again.

codelists/tests/test_actions.py
```diff
@@ -176,6 +176,34 @@ def test_create_or_update_codelist_update_no_change_to_codes(
...
```
rebkwok, line 193: the `assert cl.versions.count() == 1` line is unnecessary now that you're doing `with assert_difference`

codelists/api.py
```diff
@@ -170,6 +172,7 @@ def versions(request, codelist):
                 codelist=codelist,
                 codes=set(data["codes"]),
                 tag=data.get("tag"),
+                always_create_new_version=data.get("always_create_new_version", False),
```
rebkwok, line 175: There should probably be a test for this too  - a version of the `test_versions_post_no_difference` [here](https://github.com/opensafely-core/opencodelists/blob/main/codelists/tests/test_api.py#L153) which passes the `force_create` or `always_create_new_version` flag and does a `with assert_difference`

codelists/api.py
```diff
@@ -135,6 +136,7 @@ def codelists_post(request, owner):
         "methodology",
         "references",
         "signoffs",
+        "force_create",
```
rebkwok, line 139: Should this also be `always_create_new_version`?

opencodelists/settings.py
```diff
@@ -39,11 +39,11 @@
 SECRET_KEY = env.str("SECRET_KEY")
 
 # SECURITY WARNING: don't run with debug turned on in production!
```
rebkwok, line 41: Are these settings changed deliberate?

CarolineMorton: sorry - that was not deliberate. I added a file that I was playing around with!

codelists/api.py
CarolineMorton: changed

rebkwok (approved):

PR #1280, opened 2022-06-14
Draft codelists bug

### LINKED ISSUE #1275: Draft codelists leading to blank page
After logging into opencodelists, navigating to My Codelists>Your drafts, the links to some draft codelists lead to webpages with the codelist name but no other information (as in the screenshot below). 

The codelists which this happens for and which were fully populated in february are [here](https://www.opencodelists.org/builder/3d3644f8/) and [here](https://www.opencodelists.org/builder/58f87823/). @HelenCEBM  also has this issue with codelists containing codes, created at various periods in over the last few years. 

First raised in this [slack thread](https://bennettoxford.slack.com/archives/C33TWNQ1J/p1654766413226869).

rebkwok:
Fixes #1275 

The issue here was when the frontend attempts to find a code in the `codeset.code_to_status` dict, and that code doesn't exist (because of an update to the coding system).  The builder frontend can deal with codes with unknown status, even if they have a parent that is included or excluded, but it can't deal with codes that are missing altogether.


codelists/actions.py
```diff
@@ -491,30 +492,79 @@ def export_to_builder(*, version, author):
...
+        )["all_codes"]
         builder_actions.create_search(
             draft=draft, term=search.term, code=search.code, codes=codes
         )
+    cache_hierarchy(version=draft)
 
-    # This assert will fire if new matching concepts have been imported.  At the moment,
-    # the builder frontend cannot deal with a CodeObj with status ?  if any of its
-    # ancestors are included or excluded.  We will have to deal with this soon but for
-    # now fail loudly.
```
rebkwok, line 512: This comment was a red herring; new matching concepts are actually fine.  The issue is when there are new concepts that don't match a search, but are a descendent of an included/excluded code, because the search recreation will not create `CodeObj`s on the draft for those codes

rebkwok:
A proposed follow up: https://github.com/opensafely-core/opencodelists/issues/1281

codelists/actions.py
```diff
@@ -491,30 +492,84 @@ def export_to_builder(*, version, author):
...
+    checked it, so we don't have to redo it every time.
+    """
+    cached_updates = cache.get("updated_drafts", [])
+    new_codes = []
+    if draft.pk not in cached_updates:
+        _recreate_draft_searches(source_version or draft, draft)
+        new_codes = _check_new_concepts(draft)
+        cached_updates.append(draft.pk)
+        # cache for 24 hrs
+        cache.set("updated_drafts", cached_updates, timeout=60 * 60 * 25)
```
iaindillingham, line 519: Why `25` and not `24`?

builder/views.py
```diff
@@ -60,6 +63,15 @@ def _draft(request, draft, search_slug):
     if request.method == "POST":
         return _handle_post(request, draft)
 
+    # Currently we have no way to record the releases for a code system
+    # At the point of any GET for a draft, we can't tell if the codes may have changed,
+    # so we need to update the draft again.  This will recreate all the searches and
+    # check and update the codeset for any new concepts.
+    # We cache the updated draft ids for 60 mins so this isn't done on every single GET request
```
iaindillingham, line 70: I thought we cached them for 24 hours?

iaindillingham (commented):
I'm going to comment, rather than approve; I know @inglesp would like to 👀 and his review will be much more useful than mine, not least because this is the first time I've looked at OpenCodelists.

codelists/tests/test_actions.py
```diff
@@ -425,6 +427,96 @@ def test_export_to_builder(organisation_user, new_style_version):
     assert draft.searches.count() == new_style_version.searches.count()
 
 
+def test_update_draft(organisation_user, version_with_complete_searches):
```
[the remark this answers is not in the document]
rebkwok: Hmm. It's callled `test_update_draft` because that's what I was initially going to test, and then I realised that it was tricky to mock `_recreate_draft_searches` to pretend it had returned new concepts, because the test fixtures use `export_to_builder`, which needs it to do the right thing.  I can rename to `test_check_new_concepts` (I don't suppose the `__` is necessary

codelists/actions.py
rebkwok: typo :)

builder/views.py
rebkwok: I changed my mind about how long to cache for and decided 60 mins was overkill, given that coding systems are updated quite infrequently.  Forgot to update the comment

codelists/actions.py
iaindillingham: Relief! This caused me to doubt what little I thought I knew about computers 😆

inglesp (changes requested):
I'm marking this as "request changes", but I'm not sure what changes are going to be possible in the scope of this PR.

First though: this is a really fearless piece of work, engaging with a lot of gnarly parts of the code.  @evansd would call it a bloody difficult problem -- well done.

Unfortunately, the frontend cannot deal with new codes with unresolved status.  This is a particular issue with a more general problem of how to deal with the structure of the SNOMED hierarchy changing between releases.

A couple of examples that we've seen:

* Sometimes, the hierarchy changes so that two concepts are switch places.  For instance, in v1, A is the parent of B, but in v2, B is the parent of A.  So if with v1, A was included and B was indirectly included, then with v2, B is still indirectly included, but has no ancestors.
* Sometimes, the hierarchy changes so that a concept is replaced in the hierarchy by another concept.  For instance, in v1, A is the parent of B, which is the parent of C, but in v2, B is replaced by B'.  If with v1, A was included, B was excluded, and C was indirectly excluded, then with v2, C is still indirectly excluded, but if you click on the three dots, you'll see a message saying that it's excluded by A (which is included).

(There's also other example of the frontend failing to cope with unexpected question marks in a way that's not dependent on the SNOMED hierarchy changing in weird ways, but unfortunately I cannot remember it.  I have a feeling we've discussed it, but I cannot find a reference on #1145, which is the only PR of yours that I can find that might have come close to opening this can of worms.  Does that ring any bells?)

This is the kind of problem that caching the hierarchy related to a codelist was designed to solve, but I'd not thought about the case where a codelist is being built while somebody's working on a draft.  I think the only real solution is as you've discussed in #1281: to version codelists against a release of a coding system.  The solution may be a little bit more involved though, as we'll need to keep old versions of a coding system around -- I had thought that one way of achieving this would be by having a separate SQLite database for each release, but this requires more thought.

In the immediate future, we've got the question of how to unblock Rose.  Relatedly, we should work out how many other codelists are affected by this.  Could you create issues for these please?

rebkwok:
OK, so I misunderstood the comment about the frontend not being able to deal with unresolved codes.  I thought it was that the frontend couldn't display them (hence the blank page), and this PR makes them displayable.  I'd spent a while trying to figure out how to assign new codes based on their ancestors and I thought this was a simpler solution.  

I'll have another think about how to fix these specific codelists

rebkwok:
Updated (possible) solution:
1) update searches properly for updating drafts
2) After updating the searches - which creates the missing code objs that `_check_new_concepts` was doing before anyway - recalculate the codeset statuses according to the new hierarchy, by resetting the code_to_status dict, and re-running `codeset.update()` with all codes that are explicitly included/excluded.  This means that new decendents get assigned parent statuses (and conflicting statuses, if there are any).
3) Fix some permission issues - I was misled by the fact that we were setting edit permissions on the builder view to allow only draft authors to edit, whereas it should be any organisation user (I think)

I think there will still be issues around codes that no longer exist in a new version of a coding system (assuming that happens), but with this method the 2 codelists that raised this issue now work.  One of them has a couple of new codes that are unresolved, but they have no ancestors in the hierarchy, and can be manually included/excluded after the update (I checked that works this time)

rebkwok:
> Sometimes, the hierarchy changes so that two concepts are switch places. For instance, in v1, A is the parent of B, but in v2, B is the parent of A. So if with v1, A was included and B was indirectly included, then with v2, B is still indirectly included, but has no ancestors.

> Sometimes, the hierarchy changes so that a concept is replaced in the hierarchy by another concept. For instance, in v1, A is the parent of B, which is the parent of C, but in v2, B is replaced by B'. If with v1, A was included, B was excluded, and C was indirectly excluded, then with v2, C is still indirectly excluded, but if you click on the three dots, you'll see a message saying that it's excluded by A (which is included).

I've added tests for these two scenarios, plus the simpler appearance of a new code that has an included or excluded ancestors.  I've also run this with the current prod db to check that Rose's two codelists do get properly updated and the frontend works (there are some new codes that need to be resolved, but none that have included or excluded ancestors that would break the frontend).

codelists/tests/test_actions.py
```diff
@@ -425,6 +428,164 @@ def test_export_to_builder(organisation_user, new_style_version):
...
+    # 2) Assume that 73583000 (Epicondylitis) and 35185008 (Enthesopathy of elbow region)
+    # have swapped places - ie. 73583000 used to be the parent of 35185008 and 202855006,
+    # and was explicitly included, with 35185008 implicitly included
+
+    # i.e. in previous version
+    # +   73583000        │ ├ Epicondylitis
+    # (+) 35185008        │ │ └ Enthesopathy of elbow region
+    # (+) 202855006       │ │   └ Lateral epicondylitis
+    # in new version
+    # +   35185008        │ ├ Enthesopathy of elbow region
```
inglesp, line 513: This comment doesn't match the assertion on line 539.

codelists/actions.py
```diff
@@ -491,30 +492,116 @@ def export_to_builder(*, version, author):
...
+    cache_hierarchy(version=draft)
 
-    # This assert will fire if new matching concepts have been imported.  At the moment,
-    # the builder frontend cannot deal with a CodeObj with status ?  if any of its
-    # ancestors are included or excluded.  We will have to deal with this soon but for
-    # now fail loudly.
-    assert not draft.code_objs.filter(status="?").exists()
 
-    cache_hierarchy(version=draft)
+def _check_new_concepts(draft):
```
inglesp, line 541: I'm surprised that a function called `_check_something` makes changes to the database.

codelists/actions.py
```diff
@@ -491,30 +492,116 @@ def export_to_builder(*, version, author):
...
+        (code, status)
+        for code, status in draft.codeset.code_to_status.items()
+        if initial_statuses[code] == "?"
+    }
+    changed_code_statuses = all_changes - new_code_statuses
+    return {
+        "added": new_code_statuses,
+        "changed": changed_code_statuses,
+        "removed": old_codes,
+    }
```
inglesp, line 604: I would prefer it if this behaviour was captured in a couple of new methods on a codeset:

```
class Codeset:
    ...

    def update_hierarchy(self, new_hierarchy):
        ...
        return new_codeset

    def diff(self, other_codeset):
        ...
        return {
            "added": new_code_statuses,
            "changed": changed_code_statuses,
            "removed": old_codes,
        }
```

I feel like this would be easier to understand/test/maintain, which would offset a slightly less efficient implementation.

(The new `test_update` tests are very good, but I think they'd be easier to understand if they were tests against new `Codeset` methods.)

builder/decorators.py
```diff
@@ -32,7 +32,9 @@ def require_permission(view_fn):
 
     @wraps(view_fn)
     def wrapped_view(request, draft, **kwargs):
-        if request.user == draft.author:
+        if request.user == draft.author or draft.codelist.can_be_edited_by(
+            request.user
+        ):
```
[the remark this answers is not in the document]
rebkwok: Ah, I had thought that this was an error I'd introduced when I did the work on authors, but it was just a name change.  I'll remove it, but I think we should have some sort of indication on the builder page that you don't have permission to edit it.  It took me a while to realise that the +/- buttons weren't working because I wasn't the draft's author, not because of a frontend bug.  It wasn't until I noticed that the search field wasn't there that I figured it out.  I think it's quite likely that someone might make a draft and then send the link to someone else to look at, which will work for them, but won't be editable

codelists/actions.py
```diff
@@ -491,30 +492,116 @@ def export_to_builder(*, version, author):
...
+    checked it, so we don't have to redo it every time.
+    """
+    cached_updates = cache.get("updated_drafts", [])
+    updates = {}
+    if draft.pk not in cached_updates:
+        _recreate_draft_searches(source_version or draft, draft)
+        updates = _check_new_concepts(draft)
+        cached_updates.append(draft.pk)
+        # cache for 24 hrs
+        cache.set("updated_drafts", cached_updates, timeout=60 * 60 * 24)
```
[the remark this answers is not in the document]
rebkwok: yes, that's definitely wrong.  I'd originally put it on the session, and didn't properly think about the difference it made when it was in a cache

codelists/actions.py
rebkwok: I agree with putting this in methods on a codeset, but I'm unsure what you intended `update_hierarchy` on a `Codeset` to do.  At this point, the hierarchy has already been updated (with `cache_hierarchy`, after the searches are rerun) - I don't think that belongs in `update_hierarchy`, because it needs the `CodelistVersion` and its searches.   Would `update_hierarchy` just call the existing `Codeset.update` method instead of having the `reset` flag I added to it?  Or did you intend this to do the part where it deleted CodeObjs that no longer exist?
Sorry if this should be obvious

rebkwok: Actually, I don't think the deleting CodeObjs bit works as a Codeset method either, because it's using the draft to filter the CodeObjs as well as the draft's current codeset.  And then it uses `builder_actions.update_code_statuses` to create the new codeset on the draft, which calls the codeset's `update` method and then updates the relevant statuses of the draft's CodeObjs.  Currently a Codeset doesn't have any knowledge itself of a CodelivstVersion.

codelists/tests/test_actions.py
rebkwok: "new version" here is the changes made in the test setup, not what it looks like after the updates.  That's not clear, I'll reword

PR #1292, opened 2022-06-22
Update draft codesets if coding system has changed

rebkwok:
Fixes https://github.com/opensafely-core/opencodelists/issues/1275

When a coding system is updated, it can result in a mismatch between the draft version's hierarchy, and its codeset's `code_to_status` dict (e.g. there are new codes that are implicitly included by others, codes that have swtiched places in the hierarchy, codes that have been replaced).

 `export_to_builder` re-creates the searches from the source version on the new draft, and raises an assertion error if there are any resulting codes with unresolved status - i.e. if any search returned a new code.  However, a draft may have been created before a coding system change, and may not have been saved for review or published yet.  

With this PR:

- `export_to_builder` remains unchanged; it recreates searches, and raises and assertion error if there are unresolved codes
- adds a new builder action, `create_or_update_search` so that searches can be re-run on existing drafts as well as being created for new ones
- adds a new codelists action `update_draft_codeset`.  This is currently only called by a management command, to allow us to manually fix broken drafts.  It:
    1. re-runs the searches on the draft; this creates new code objs with unresolved status if there are new matching concepts
    2. deletes any code objs that no longer exist in the coding system
    3. updates the codeset by applying all the known explicitly included/excluded codes again, and recalculating all inherited statuses
    4. outputs a dict of the changes, for reporting to the user
- adds a management command `update_draft` that takes a draft hash and runs  `update_draft_codeset` as above

### DIFF
--- a/builder/views.py
+++ b/builder/views.py
@@ -65,6 +65,18 @@ def _draft(request, draft, search_slug):
     codeset = draft.codeset
     hierarchy = codeset.hierarchy
 
+    # Ensure all descendants of included/excluded codes are in the codeset and are resolved
+    # If any descendants are missing, it's due to new/modifed concepts in a coding system,
+    # and the frontend can't deal with it.
+    # For now, we just fail loudly with a reminder of how to fix
+    for code in codeset.all_codes():
+        if codeset.code_to_status[code] not in ["?", "!"]:
+            for descendant in hierarchy.descendants(code):
+                assert codeset.code_to_status.get(descendant, "?") != "?", (
+                    f"Unresolved or missing descendant codes in the codeset for draft {draft.hash}; "
+                    "fix by running the 'update_draft' manage command"
+                )
+
     if search_slug is None:
         search = None
         displayed_codes = list(codeset.all_codes())


codelists/codeset.py
```diff
@@ -191,3 +195,24 @@ def update(self, updates):
...
+        """
+        removed_codes = set(other_codeset.code_to_status) - set(self.code_to_status)
+        all_changes = set(self.code_to_status.items()) - set(
+            other_codeset.code_to_status.items()
+        )
+        new_code_statuses = {
+            (code, status)
+            for code, status in self.code_to_status.items()
+            if other_codeset.code_to_status[code] == "?"
+        }
```
inglesp, line 212: If `code` is not in `other_codeset`, we get a `KeyError`.  Could this happen?

At the moment, this method is tested very indirectly, and I'd like to see it thoroughly tested directly.  Not only would this give us more confidence that it's correct, but this would also provide helpful clues about how we expect it to work.

rebkwok:
Superceded (for now at least) by #1294

PR #1294, opened 2022-06-23
Add a manage command to fix a draft version after a coding system update

rebkwok:
Fixes https://github.com/opensafely-core/opencodelists/issues/1275

When a coding system is updated, it can result in a mismatch between the draft version's hierarchy, and its codeset's code_to_status dict (e.g. there are new codes that are implicitly included by others, codes that have swtiched places in the hierarchy, codes that have been replaced).

This PR is an interim strategy to allow us to fix up drafts that were created prior to a coding system update, but defers the full implementation for future us.
- Adds a management command that can be run manually to update a draft
- Modifications to the main codebase are as minimal as possible.  The only addition is `Codeset.reapply_statuses`, and this is only called from the management command.  This means there's some duplicated code in the management command, but the idea was to keep and document all the logic here so that it can be extracted later if/when we get to making the proper solution.

- The management command:

    1. Logs the codes and statuses of the original draft
    2. re-runs the searches on the draft; this creates new code objs with unresolved status if there are new matching concepts
    3. deletes any code objs on the draft that no longer exist in the coding system
    4. updates the codeset by applying all the known explicitly included/excluded codes again, and recalculating all inherited statuses

### DIFF
--- a/codelists/codeset.py
+++ b/codelists/codeset.py
@@ -150,6 +150,22 @@ class Codeset:
 
         yield from helper(self.defining_tree())
 
+    def reapply_statuses(self):
+        """
+        Build a new Codeset by reapplying the directly included and excluded status, and
+        recalculating inherited statuses
+        """
+        # Find all explicitly included/excluded codes
+        explicitly_included_or_excluded = {(code, "+") for code in self.codes("+")} | {
+            (code, "-") for code in self.codes("-")
+        }
+
+        # Reset the code_to_status dict so all codes are now unresolved
+        self.code_to_status = {code: "?" for code in self.code_to_status}
+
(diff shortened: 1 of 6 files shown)


rebkwok:
Note: this keeps the tests from #1292; the individual methods on the manage command could be split out and tested, especially the `diff` method, but I'd rather leave that for later and get this PR in.

I also haven't done the checks in the builder view (to raise an exception if we know the draft won't be properly displayable) - given we only know of 2 codelists this applies to atm, I'd also like to make that a future issue.

inglesp (approved):
One small change regarding transactions, but apart from that, this looks good to go!

Thanks again for all you efforts in getting to this stage.

PR #1342, opened 2022-08-03
Include wildcard search in helptext for ICD-10 only

rebkwok:
For an ICD-10 codelist, users can search for codes that start with a prefix, using a wildcard (`code:xyz*`).  For other conding systems, this doesn't make sense (see [this slack thread](https://bennettoxford.slack.com/archives/C63UXGB8E/p1659525543519349)) and isn't implemented, but the help text under the search field still refers to it, which is confusing, especially as it will never return anything (i.e. for a BNF codelist, `code:01` returns  lots of results (all, `code:01*` returns nothing).  This PR just removes the reference to "startswith" searches for non-ICD-10 codelists.

### DIFF
--- a/static/src/js/builder/codelistbuilder.jsx
+++ b/static/src/js/builder/codelistbuilder.jsx
@@ -165,7 +165,10 @@ class CodelistBuilder extends React.Component {
             {this.props.isEditable && (
               <>
                 <h6>New search</h6>
-                <SearchForm searchURL={this.props.searchURL} />
+                <SearchForm
+                  searchURL={this.props.searchURL}
+                  codingSystemName={this.props.metadata.coding_system_name}
+                />
                 <hr />
               </>
             )}
@@ -385,7 +388,7 @@ function Version(props) {
 }
 
 function SearchForm(props) {
-  const { searchURL } = props;
+  const { searchURL, codingSystemName } = props;
 
   return (
     <form method="post" action={searchURL}>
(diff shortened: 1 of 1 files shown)


static/src/js/builder/codelistbuilder.jsx
```diff
@@ -413,12 +416,20 @@ function SearchForm(props) {
...
-            instance, use <code>code:xyz</code> to find the concept with code{" "}
-            <code>xyz</code>, or <code>code:xyz*</code> to find all concepts
-            with codes beginning <code>xyz</code>.
-          </p>
+          {codingSystemName === "ICD-10" ? (
+            <p>
+              To search by code, prefix your search with <code>code:</code>. For
+              instance, use <code>code:xyz</code> to find the concept with code{" "}
+              <code>xyz</code>, or <code>code:xyz*</code> to find all concepts
+              with codes beginning <code>xyz</code>.
```
inglesp, line 424: Could you add something like "(Prefix search only available for ICD-10.)"?

inglesp (approved):

static/src/js/builder/codelistbuilder.jsx
rebkwok: Just to the ICD-10 version?  Or to all?
"prefix search" might be confusing, since the text already refers to prefixing searches with "code:" to search by code.  "wildcard search" instead?

inglesp: > Just to the ICD-10 version? Or to all?

I think to just the ICD-10 version.

And "wildcard search" is better, thanks.

PR #1371, opened 2022-09-20
Upgrade requirements

rebkwok:
Upgrading all the things; necessitated a couple of other changes:
- deprecated `USE_L1ON` setting removed
- replacing structlog.threadlocal which is now [considered legacy](https://www.structlog.org/en/stable/thread-local.html)


opencodelists/settings.py
```diff
@@ -176,8 +176,6 @@
 
 USE_I18N = True
 
-USE_L10N = True
-
```
rebkwok, line 180: deprecated; defaults to True since 4.1, will be removed in 5.0

ghickman (approved):

PR #1374, opened 2022-09-26
Confirmation prompt on deleting draft

rebkwok:


rebkwok:
I made this work and then refactored it to be like the way the other popups work; will squash commits before merging

inglesp (approved):
I've not written any React for a long time, but this all looks plausible!

PR #1392, opened 2022-10-11
Support diffing dm+d codelists

### LINKED ISSUE #1387: Support diffing dm+d codelists
Currently we don't support diffing dm+d codelists (I think it gives 500s) because we do not know which column contains the codes.  As part of codelist rot investigations, we've identified the [headings that have been used for columns with dm+d codes](https://github.com/opensafely/medication-codelists/blob/main/normalise_codelists.py#L57-L59).

Note that some dm+d codelists contain invalid codes, whose terms we won't be able to look up.  In the first instance, we should display something like "Invalid code" for these codes, but we may be able to work out from the codelist what the intended term was.

rebkwok:
Fixes #1387

We have a view already for diffing codelists, e.g. https://www.opencodelists.org/codelist/primis-covid19-vacc-uptake/bmi_stage/v1.2/diff/08641331/

This currently only works for codelist versions that have code objects - i.e. new-style codelists that have been imported from CSV using the "create a codelist" form, or built using searches in the codelist builder.  dm+d codelists aren't currently buildable via the builder, so only contain uploaded data in the `CodelistVersion.csv_data` field.  To complicate things further, dm+d uploaded csv data may have different headers, so we have to figure out which column contains the code.  

This PR updates the codelist diff view to support dm+d codelists.  This involves:
1) retrieving codes from the uploaded csv_data for a dm+d codelist verision.
I checked all the headers in the database for dm+d codelists, and there are a limited number of possible headers (most versions have been created by downloading a BNF codelist as dm+d and so have standardised headers). If they have no headers, we raise an error for now (I think this is only a few very old codelist versions)
2) Adding a new `code_to_term` function for dm+d, in line with other coding systems.  The diff view uses this to build the summaries of code differences.
3) Adds some new dmd fixtures and tests

### DIFF
--- a/coding_systems/dmd/coding_system.py
+++ b/coding_systems/dmd/coding_system.py
@@ -1,2 +1,26 @@
+from .models import AMP, AMPP, VMP, VMPP, VTM
+
 name = "Dictionary of Medicines and Devices"
 short_name = "dm+d"
+
+
+def lookup_codes(codes):
+    # A code is a unique identifier in dm+d which corresponds to a SNOMED-CT code
+    # It could be the identifier for any of AMP, VMP, VTM, VMPP, AMPP
+    # dm+d codelists generally only include AMPs and VMPs, so we attempt to match codes in
+    # these models first
+    # If not found, look up VTM, VMPP, AMPPs also, in case of user-uploaded codelists that
+    # might contain these
+    for code in codes:
+        for model_cls in [AMP, VMP, AMPP, VMPP, VTM]:
+            try:
+                yield code, model_cls.objects.get(id=code).nm
+                continue
+            except model_cls.DoesNotExist:
+                ...
+
+
+def code_to_term(codes):
+    lookup = {code: term for code, term in lookup_codes(codes)}
+    unknown = set(codes) - set(lookup)
+    return {**lookup, **{code: "Unknown" for code in unknown}}
--- a/coding_systems/dmd/coding_system.py
+++ b/coding_systems/dmd/coding_system.py
@@ -4,7 +4,7 @@ name = "Dictionary of Medicines and Devices"
 short_name = "dm+d"
 
 
-def lookup_codes(codes):
+def lookup_names(codes):
     # A code is a unique identifier in dm+d which corresponds to a SNOMED-CT code
     # It could be the identifier for any of AMP, VMP, VTM, VMPP, AMPP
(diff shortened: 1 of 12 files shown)


inglesp (approved):
Nice, thanks.  One question about performance impact.

coding_systems/dmd/coding_system.py
```diff
@@ -1,2 +1,26 @@
...
+    # these models first
+    # If not found, look up VTM, VMPP, AMPPs also, in case of user-uploaded codelists that
+    # might contain these
+    for code in codes:
+        for model_cls in [AMP, VMP, AMPP, VMPP, VTM]:
+            try:
+                yield code, model_cls.objects.get(id=code).nm
+                continue
+            except model_cls.DoesNotExist:
+                ...
```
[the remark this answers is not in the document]
rebkwok: In practice, it's likely to be at most 2, except for unknown codes.  I've tried it with a few different codelists, and it seems reasonable.  I did consider checking for a dmd_type column (the ones that have been converted from bnf), and only checking the matching type.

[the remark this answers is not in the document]
rebkwok: Refactored now so it should be at most 5 trips to the db in total, not 5 per code.  This is better, I think.

coding_systems/dmd/coding_system.py
```diff
@@ -11,13 +11,14 @@ def lookup_names(codes):
...
-            except model_cls.DoesNotExist:
-                ...
+    codes = set(codes)
+    for model_cls in [AMP, VMP, AMPP, VMPP, VTM]:
+        matched = dict(model_cls.objects.filter(id__in=codes).values_list("id", "nm"))
+        for code, name in matched.items():
+            yield code, f"{name} ({model_cls.__name__})"
+        codes = codes - set(matched.keys())
+        if not codes:
+            break
```
inglesp, line 21: Does this `break` mean that, if no AMPs are matched, we won't try to find objects of any other type?

rebkwok: No, it should mean that if ALL codes match AMPs, we won't check any others.

inglesp: Ah, I didn't read that properly.  Thanks!

inglesp (approved):

PR #1436, opened 2022-11-11
Skip coverage for third party ECL parser utils

rebkwok:
The parser utils for the SNOMED CT Expression Constraint Language are a set of tools that we[ download](https://github.com/opensafely-core/opencodelists/blob/main/coding_systems/snomedct/parser_utils/README).  We have a small ecl_parser module that makes use of them, but the tools themselves are third-party shouldn't need to be included in our overall coverage.

This PR adds a couple more tests for the ecl_parser and omits the third party toold from coverage.

### DIFF
--- a/coding_systems/snomedct/tests/test_ecl_parser.py
+++ b/coding_systems/snomedct/tests/test_ecl_parser.py
@@ -24,6 +24,27 @@ def test_handle(subtests):
             assert handled["excluded"] == excluded
 
 
-def test_parse_error():
+def test_parse_error_syntax_error():
     with pytest.raises(ParseError):
         handle("(111111 OR 222222) MINUS (333333 OR 444444")
+
+
+def test_parse_error_attempting_full_context():
+    with pytest.raises(ParseError, match="Attempting full context"):
+        handle("111111 < 222222")
+
+
+def test_error_nested_minus_operation():
+    node = {
+        "type": "minus",
+        "lhs": {"type": "expr", "operator": None, "value": "222222"},
+        "rhs": {"type": "expr", "operator": None, "value": "333333"},
+    }
+    with pytest.raises(AssertionError, match=str(node)):
+        handle("111111 MINUS (222222 MINUS 333333)")
+
+
+def test_error_child_of():
+    # Child of operator "<!" is not allowed
+    with pytest.raises(AssertionError):
+        handle("<!111111")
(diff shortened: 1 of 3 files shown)


coding_systems/snomedct/tests/test_ecl_parser.py
```diff
@@ -24,6 +26,40 @@ def test_handle(subtests):
...
     with pytest.raises(ParseError):
         handle("(111111 OR 222222) MINUS (333333 OR 444444")
+
+
+def test_parse_error_attempting_full_context():
+    with pytest.raises(ParseError, match="Attempting full context"):
+        handle("111111 < 222222")
+
+
+def test_parse_error_ambiguity():
```
inglesp, line 39: This is the same as the test above.

coding_systems/snomedct/tests/test_ecl_parser.py
```diff
@@ -24,6 +26,40 @@ def test_handle(subtests):
...
+    with pytest.raises(ParseError, match="Attempting full context"):
+        handle("111111 < 222222")
+
+
+@patch("coding_systems.snomedct.ecl_parser.parse")
+def test_error_unknown_tree_type(mock_parse):
+    bad_tree = {"type": "unk", "operator": None, "value": "111111"}
+    mock_parse.return_value = bad_tree
```
inglesp, line 47: I'm not sure what this is really testing.  Are we mocking to ensure full coverage?

In pure computation (eg parsing) if we can't hit a branch without mocking, I'd rather remove the branch, or at least exclude it from coverage.

inglesp (approved):
I like where this is going!

coding_systems/snomedct/tests/test_ecl_parser.py
rebkwok: I'll exclude it - I couldn't find a simple test that would return an unexpected tree type, but I don't feel like I understand the ECL enough to be sure that it couldn't happen.

coding_systems/snomedct/tests/test_ecl_parser.py
rebkwok: Ah yes, this one should have been deleted; copied and pasted to add a test for the ambiguity error, but I couldn't find a simple example that would raise one.  Afaict, Ambiguity Error is an antlr4 thing, where the expression could be parsed in more than one way, but I couldn't find an expression that would cause an ambiguity without hitting a different error first.

PR #1440, opened 2022-11-16
Ensure the format of database_alias always follows the expected pattern

rebkwok:
A CodingSystemRelease is associated with a database_alias, which is expected to be in the format `<coding_system>_<release_name>_<valid_from>`.  The database alias is set on model save, however, it was previously possible for it to be specified on the model as something different.  We don't want to allow that, as it's used to create the release's database file, and it's used by the database router (which expects it to be in this format).  

This PR also adds a test to complete the test coverage on the database router.

### DIFF
--- a/coding_systems/versioning/models.py
+++ b/coding_systems/versioning/models.py
@@ -26,10 +26,17 @@ class CodingSystemRelease(models.Model):
         unique_together = ("coding_system", "release_name", "valid_from")
 
     def save(self, *args, **kwargs):
-        if not self.database_alias:
-            self.database_alias = slugify(
-                f"{self.coding_system}_{self.release_name}_{self.valid_from.strftime('%Y%m%d')}"
-            )
+        # CodingSystem methods and the database router depend on the format
+        # of database_alias following an expected pattern
+        database_alias = slugify(
+            f"{self.coding_system}_{self.release_name}_{self.valid_from.strftime('%Y%m%d')}"
+        )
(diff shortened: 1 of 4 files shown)


coding_systems/versioning/models.py
```diff
@@ -26,10 +26,11 @@ class Meta:
         unique_together = ("coding_system", "release_name", "valid_from")
 
     def save(self, *args, **kwargs):
-        if not self.database_alias:
-            self.database_alias = slugify(
-                f"{self.coding_system}_{self.release_name}_{self.valid_from.strftime('%Y%m%d')}"
-            )
+        # CodingSystem methods and the database router depend on the format
+        # of database_alias following an expected pattern, therefore we overwrite any
+        # attempts to manually set it.
```
inglesp, line 31: The approach taken elsewhere is to only make changes through functions in actions.py.  This isn't enforced though, so perhaps it's right to be cautious here.

Or would it be better to assert that if `self.database_alias` is not `None`, then it's as expected?  That way, we can catch any code that's doing the wrong thing.

inglesp (approved):

coding_systems/versioning/models.py
rebkwok: Hmm, maybe asserting the format is correct is better than silently overwriting it. It _should_ only be set by the import_data function for each coding system (which will use a common function in the next PR), but that's also not enforced

PR #1443, opened 2022-11-17
Import coding system releases into new databases

rebkwok:
This ensures that all the `import_data` scripts that import new coding system releases now import to their own new database.

- Adds some additional parameters to the main `import_data` command to specify coding system release info (release name, valid from date, references)
- Wraps the various coding systems' `import_data` functions in a context manager that handles setting up the new database and CodingSystemRelease object, and cleaning things up if anything goes wrong.  
- Updates the actual import_data code (which generally required minimal change, other than to specify the database to use). 
- gets rid of unused readv2 models and import_data code


opencodelists/management/commands/import_data.py
```diff
@@ -14,23 +19,94 @@ def iter_possible_modules():
...
+                self.stdout.write(possibility)
 
             sys.exit(1)
 
+        module_name = dataset.rsplit(".", 1)[-1]
+        if module_name in CODING_SYSTEMS:
+            import_data_kwargs = self.get_coding_system_kwargs(
+                module_name, release_name, valid_from, import_ref, force
+            )
+        else:
```
inglesp, line 78: Feels like we've got two separate commands here: `import_coding_system_data` and `import_mapping_data`.  If we separated them, would the resulting code be simpler?

opencodelists/management/commands/import_data.py
```diff
@@ -14,23 +19,94 @@ def iter_possible_modules():
...
+            import_data_kwargs = {}
+
         fn = getattr(mod, "import_data")
-        fn(release_dir)
+        fn(release_dir, **import_data_kwargs)
+
+    def get_coding_system_kwargs(
+        self, coding_system, release_name, valid_from, import_ref, force
+    ):
+        if not all([release_name, valid_from]):
```
inglesp, line 87: This is an example of some code that would be simplified if we split out the two commands, because we could make `release_name` and `valid_from` required arguments with argparse.

coding_systems/bnf/README
```diff
@@ -8,10 +8,18 @@ For more details, see our most popular blog post [0].
...
 
-        ./manage.py import_data coding_systems.bnf path/to/dir/
+    ./manage.py import_data coding_systems.bnf path/to/zip/file
+    --release <release_name> --valid-from YYYYMMDD --import-ref <ref>
 
+- `release` is a short label for this coding system release - the official version number/reference if there is one.
+- `valid-from` is the date that this release is valid from (NOT the date it is being
+  imported into OpenCodelists), in YYYYMMDD format
+- `import-ref` is an optional long-form reference field for any other potential
+  references or comments on this import if desired, e.g. full filenames
```
inglesp, line 22: Can you be more explicit here?  Given a filename like `20220701_1656687539491_BNF_Code_Information.zip`, how would I know what to use for `release` and `valid-from`?  And can you offer some more prescriptive advice for `import-ref` too?

coding_systems/bnf/import_data.py
```diff
@@ -1,33 +1,45 @@
...
+            records = {type: set() for type in TYPES}
+            with open(path) as f:
+                for r in csv.DictReader(f):
+                    parent_code = None
+                    for type in TYPES:
+                        name = r[f"BNF {type}"]
+                        code = r[f"BNF {type} Code"]
+                        if "DUMMY" not in name:
+                            records[type].add((code, name, parent_code))
+                            parent_code = code
```
inglesp, line 37: It might be clearer if everything above this point was moved out of the `with CodingSystemImporter` context manager, since (afaict) it doesn't need the context.  It would also let us dedent the `with CodingSystemImporter` line and everything below it, because it would have finished using `tempdir`.

coding_systems/base/import_data_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    def setup_new_import_database(self):
+        print("Setting up database...")
+        # get or create the CodingSystemRelease; note we don't update an existing
+        # CodingSystemRelease with a new import_timestamp or import_ref yet, in case
+        # anything goes wrong during the import
+        # Note that this method is wrapped in transaction.atomic, but this will ONLY
+        # roll back the CodingSystemRelease object in case of errors raised in this
+        # method itself.  The __exit__ method deals with reverting the state of the
+        # release database if anything goes wrong in the import itself.
+        self.cs_release, new = CodingSystemRelease.objects.get_or_create(
```
inglesp, line 87: Under what circumstances would we be getting an already-created `CodingSystemRelease`?

inglesp: Looks like we're supporting reimporting a release.  Given the complexity this introduces (with multiple setup and teardown paths) is this worth it?

(I think the code is correct!  But I'm having a slightly hard time following it.)

coding_systems/conftest.py
```diff
@@ -22,3 +23,31 @@ def coding_systems_tmp_path(settings, tmp_path):
...
         (tmp_path / coding_system).mkdir(parents=True, exist_ok=True)
     yield tmp_path
+
+
+def mock_migrate_coding_system(*args, **kwargs):
+    """
+    `import_data` runs `migrate` for the coding system app, to set up the the newly
+    created coding system version database.  However, tests run with --no-migrations,
+    so this function uses the schema from the test database for the relevant coding
+    system to set up the new db in tests (note that the default ab can't be used
```
inglesp, line 34: s/ab/db/

coding_systems/ctv3/README
```diff
@@ -6,5 +6,24 @@ CTV3 is a retired system of clinical terms.
 It is the successor to Read v2, and is replaced by SNOMED CT.  (The "CT" stands for
 "Clinical Terms" in both CTV3 and SNOMED CT.)
 
-No new CTV3 codes are published, although sometimes TPP have published new "Y codes",
-which look like CTV3 codes.
+No new CTV3 codes are published, although sometimes TPP have published new "Y codes", which look like CTV3 codes.
+
+The last release of CTV3 was on 1st Aug 2018.  This data is no longer available to
+download online, but can be retrieved from the zip file on dokku1: `nhs_readctv3_25.0.0_20180401000001.zip`.
```
inglesp, line 12: Isn't dokku1 going away?  Should we move this somewhere more permanent?

inglesp (approved):
Lots to like here!  There are some really good explanatory comments and tests.

The only thing I'd like to see changed is a bit more guidance in the READMEs about how exactly to find the versions/dates/refs etc for each release.

opencodelists/management/commands/import_data.py
rebkwok: Possibly - I'll take a look

coding_systems/ctv3/README
rebkwok: yes, all refs to dokku1 should be dokku3 now - there are a few other places in docs to update too

coding_systems/base/import_data_utils.py
rebkwok: > Under what circumstances would we be getting an already-created `CodingSystemRelease`?

If we're overwriting an existing import.  It's been extremely useful in testing, but I'd hope it wouldn't be used often in production.  There could be a case where someone downloads the wrong release file and wants to re-import it.

I'll try to add some better comments!  I did make sure this is all fully covered by tests :)

rebkwok:
@inglesp This is updated with your suggestions, and some README improvements, including instructions on where to save the data files on the server (we were planning to keep them there for reference) and what to use for `--release` and `--valid-from`.  I don't have any response yet from the clin-informatics team about BNF, but given that the two files I have from this year (2022-07-01 and 2022-11-01) contain different data, I think that probably updates to a release happen without any change to the release number, and are only idenitifiable by the date in the downloaded file.  If that turns out not to be the case, I'll update that readme.

For now, assumptions on release name to use:
- snomedct: has a straightforward version number
- dmd: has a version number based on the month/week, release name "YYYY release-number "
- icd10: releases are versioned by year, except in odd cases.  We can get the version as referenced by ICD-10 (and used in the ICD-10 browser) from the XML file; usually it's just something like "2016"; the current one is "2019-covid-expanded"
- BNF: probably referenced by release number (as on download page) and date (from filename) - e.g. "82 (2022-11-01)" 
- CTV3: I've suggested referencing this with both the last CTV3 release date and the date the TPP codes were extracted, e.g. "2018-04-01 with TPP extensions 2022-11-28".  Is that too wordy? Do we want to make some sort of incremental release name/number for it?

rebkwok:
> For another day, maybe we could parse the release names etc out of the filename.

I thought about it, and decided to leave it for now, because we can't get everything out of the filename for all of them - in fact I think only dm+d would allow that, so we still need the options for passing both release name and valid date in manually.

PR #1457, opened 2022-11-24
Check newly imported coding system releases for compatibility with existing codelist versions

### LINKED ISSUE #1441: Record coding system releases compatibility for codelist versions
A codelist version has a record of the coding system release that was used to create it.  It may be compatible with subsequent releases of the coding system, and we should record that.

How can we tell if a codelist version is compatible?
1) If it was created with searches, rerun the searches with the new release.  If all searches bring back identical results, the codelist version is compatible.
2) For versions without searches, if it has a cached hierarchy, we can recreate the hierarchy using the new coding system release and the version's codes.  If the old and new hierarchies match, the codelist version is compatible.
3) If there are no searches and no cached hierarchy, the best we can do is to confirm that the version is not _incompatible_ - ie. there are no codes on the version that are missing in the new relase.  This is probably not enough to be able to say that it is compatible, since we can't know if any relationships between codes formed part of the author's intention, or if any of those relationships have changed.

We need to:
- check for compatible versions after importing a new release (we can limit this to versions that were created or compatible with the previous release)
- record releases that the version is compatible with
- display compatible releases on the version/builder pages

rebkwok:
Adds a `compatible_releases` field to CodelistVersion, so every CodelistVersion now knows:
1) which coding system release it was created with and
2) which subsequent coding system releases it is compatible with

Compatibility is checked after importing a new release, by looking for all CodelistVersions that are compatible/created with the _previous_ release (determined by the `valid_from` date).  If a version wasn't compatible with the previous release, we assume it also isn't compatible with this release.  This should mean that the number of versions that we need to check after each import is manageable (for the first new import, where every existing codelist version is marked as "unknown", we'll need to check all of them; for snomed, this is ~3000 versions, and takes a few minutes).

A CodelistVersion is deemed to be compatible with a new release if:
- It has searches, and rerunning the searches with the new release returns identical results OR
- It has a cached hierarchy (this applies to all coding systems that we can import new releases for, except dm+d), and a hierarchy created from its codes with the new release is identical 

The hierarchy check is more restrictive - e.g. there could be a new parent code in the hierarchy that isn't returned by a search, so we do the search check first, and the hierarchy check only on codelists that don't have searches.

On the version and builder pages, the codelist version details also list compatible releases (in an expandable section, since it could be quite a long list for some versions, especially when dm+d ones - with weekly releases - have hierarchies/searches):

Screenshot from 2022-11-29 16-31-20

Fixes #1441


coding_systems/base/import_data_utils.py
```diff
@@ -136,3 +149,118 @@ def setup_new_import_database(self):
...
+    1) versions with searches by rerunning the searches with the new release and checking
+       the results are the same
+    2) versions without searches, but with hierarchies, by recreating the hierarchy with the
+       new release and checking if it is identical to the version's cached hierarchy
+    """
+    previous_release = (
+        CodingSystemRelease.objects.filter(coding_system=coding_system_id)
+        .exclude(id=new_coding_system.release.id)
+        .latest("valid_from")
+    )
```
inglesp, line 179: I'm quite paranoid...  I feel it would be helpful to assert here that `previous_release.valid_from` is earlier than new_coding_system.release.valid_from`.

coding_systems/base/import_data_utils.py
```diff
@@ -136,3 +149,118 @@ def setup_new_import_database(self):
...
+    print(
+        f"Found {len(versions_to_check)} versions compatible with previous release '{previous_release.database_alias}'."
+    )
+    # Find any versions with searches
+    print("Checking compatibility by search:")
+    versions_with_searches = set(
+        versions_to_check.filter(
+            codelist__coding_system_id=coding_system_id, searches__isnull=False
+        )
+    )
```
inglesp, line 194: I don't think this is enough: you can have a codelist version that was created by uploading a CSV, and where a user has done some searches, but where these searches don't cover all of the codes in the codelist.

So should we be only checking the hierarchies?

coding_systems/base/import_data_utils.py
```diff
@@ -136,3 +149,118 @@ def setup_new_import_database(self):
...
+    for counter, version in enumerate(versions, start=1):
+        update_progress(counter, number_of_versions, comment="Checking: ")
+        is_compatible = compatibility_check_function(coding_system, version)
+        if is_compatible:
+            version.compatible_releases.add(coding_system.release)
+            compatible_count += 1
+    return compatible_count, number_of_versions
+
+
+def _check_version_by_search(coding_system, version):
```
inglesp, line 230: This needs a docstring explaining why the searches being the same implies compatibility.  (Although actually I think it doesn't -- see above.)

coding_systems/base/import_data_utils.py
```diff
@@ -136,3 +149,118 @@ def setup_new_import_database(self):
...
+        current_codes = set(search.results.values_list("code_obj__code", flat=True))
+        updated_codes = do_search(coding_system, term=search.term, code=search.code)[
+            "all_codes"
+        ]
+        if current_codes != updated_codes:
+            return False
+    return True
+
+
+def _check_version_by_hierarchy(coding_system, version):
```
inglesp, line 241: This needs a thorough docstring and some really explicit tests.  My nervousness here is that we're turning some functionality that was designed for to support performance into something that we're depending on for data correctness.

New tests should demonstrate the ways that a full coding system hierarchy can change, and the ways that this changes whether or not a codelist version's hierarchy (a `Hierarchy` instance) needs to change.  By tweaking `Hierarchy.from_codes` so that it calls through to a new class method that doesn't know anything about coding systems, you should be able to test this just by calling this new class method.  I'd be happy to pair on this if you like.

coding_systems/base/import_data_utils.py
rebkwok: hmmm...it should be.  Let me check if I've forgotten to push something

coding_systems/base/import_data_utils.py
rebkwok: So should we only check by searches for codelist versions where the searches cover all codes?

coding_systems/base/import_data_utils.py
rebkwok: I've added that now - I had updated it so it filtered by valid_from < new_coding_system.release.valid_from, but that seems to have gone missing in a rebase somewhere

rebkwok:
@inglesp This PR is updated and I think ready for another review; I've added commits with new changes rather than rewriting the existing ones; the last 5 are new since you reviewed last.

As discussed, the compatibility check is now by both hierarchies and searches.  A codelist version is only compatible with a new release if the hierarchy built with the new and old release are identical AND the search results with the new and old release are identical.

I added tests for `Hierarchy.from_codes`, using a mock coding system which is essentially a larger hierarchy, which can be used to build the `Hierarchy` from code(s), which is a subset of it, and also tests to verify changes to the coding system that do or don't result in changes to the hierarchy.

I also realised that we only want to check and update compatibility for codelist versions that are under-review or published.  Codes for those won't change, whereas for drafts, they could change, and we don't want to have to re-check compatibility every time a draft is updated. So the compatibility check on importing new data only looks at under-review and published versions, and when a draft is saved for review, it's checked against any newer releases of the coding system at that point.

inglesp (approved):
Great stuff, thanks.  I want to rewrite all the existing tests with the new mock coding system!

coding_systems/base/import_data_utils.py
```diff
@@ -180,57 +200,95 @@ def update_codelist_version_compatibility(coding_system_id, new_database_alias):
...
+
+    # All codelist versions that existed prior to implementation of coding system releases
+    # have an "unknown" coding system release, which represents the coding system data at that
+    # point. For older codelist versions, that is not the actual coding system release that was
+    # used to create the codelist version, and we do not have a record of, or access to, the
+    # coding system data at the time of its creation.
+    # For any codelist with the "unknown" coding system release, we fall back to using
+    # its cached hierarchy.
+    # For any codelist created with a known coding system release, we can re-create its hierarchy
+    # from the original coding system release and use that for comparison with the new release.
```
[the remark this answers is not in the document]
rebkwok: It is, of course, much slower, so I've changed it back to always using the cache for the original one

PR #1459, opened 2022-11-30
Validate old-style CSV uploads

### LINKED ISSUE #1388: Validate dm+d codes on upload
We currently don't do any validation of dm+d codelists on upload.  We should do so!  This will involve adding some validation to [`codelists.forms.CodelistCreateForm`](https://github.com/opensafely-core/opencodelists/blob/main/codelists/forms.py#L120-L130), along the lines of the validation in [`opencodelists.forms.CodelistCreateForm`](https://github.com/opensafely-core/opencodelists/blob/main/opencodelists/forms.py#L74-L99).

(Having two forms is a bit of a mess, sorry.)

rebkwok:
Fixes #1388 

Old-style CSV data can be uploaded at `/codelist/<user/username or org>add`; this adds the uploaded csv data to `CodelistVersion.csv_data` rather than generating `CodeObj` instances for each code.  CodelistVersions without hierarchies in OpenCodelists (i.e. dm+d, OPCS-4, ReadV2) can only be uploaded via this method, but all coding systems are allowed.  It's mainly used for uploading dm+d codelists now.  No validation was being done on the uploaded data, beyond checking that headers don't contain whitespace, and all rows have the expected number of columns.

This PR adds validation of the CSV data, by checking that all codes can be looked up in the latest version of the coding system, in the same way that the normal "create a codelist" form does (reports the number of invalid codes, and where the first one is found).  If validation fails, it also reports the coding system release used to validate.  Currently this is always the most recent release that OpenCodelists has.  Even though it can't be changed via the interface yet; it may useful for identifying why an upload doesn't work.  

Validation is also added to the CodelistVersionForm, which is used to upload new CSV data for an existing old-style codelist.

I also decided to limit valid code columns in the form, so old-style CSV data requires a specific code column, which must be either "code" or "dmd_id"; this won't affect any existing CodelistVersions with csv_data that has different headers, and makes it simpler to determine with confidence which column contains the codes.  (`dmd_id` is allowed because the most likely use for this form now is for codelists that have been created in BNF and downloaded as dm+d; `dmd_id` is the header used for the dmd codes in those downloads). 

Also required some updates to tests because they need to use valid codes that exist in the test fixtures.

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -472,10 +472,17 @@ class CodelistVersion(models.Model):
             return self._new_style_codes()
 
     def _old_style_codes(self):
-        if self.coding_system_id in ["ctv3", "icd10", "snomedct"]:
+        if self.coding_system_id in ["bnf", "ctv3", "icd10", "snomedct"]:
             headers, *rows = self.table
 
-            for header in ["CTV3ID", "CTV3Code", "ctv3_id", "snomedct_id", "id"]:
+            for header in [
+                "CTV3ID",
+                "CTV3Code",
+                "ctv3_id",
+                "snomedct_id",
+                "id",
+                "code",
+            ]:
                 if header in headers:
                     ix = headers.index(header)
                     break
--- a/codelists/forms.py
+++ b/codelists/forms.py
@@ -186,6 +186,7 @@ class CodelistUpdateForm(forms.Form):
 
 
 class CodelistVersionForm(forms.Form, CSVValidationMixin):
+    coding_system_id = forms.CharField(widget=forms.HiddenInput())
     csv_data = forms.FileField(label="CSV data")
 
     class Meta:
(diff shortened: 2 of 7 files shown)


codelists/models.py
```diff
@@ -472,10 +472,17 @@ def codes(self):
             return self._new_style_codes()
 
     def _old_style_codes(self):
-        if self.coding_system_id in ["ctv3", "icd10", "snomedct"]:
+        if self.coding_system_id in ["bnf", "ctv3", "icd10", "snomedct"]:
```
rebkwok, line 475: I don't know if there was a reason that `bnf` is missing here, but it is a coding system with a hierarchy, so old-style uploads for bnf codelists will try to build a hierarchy, but have no codes to build from,  and will [error](https://sentry.io/organizations/ebm-datalab/issues/3777833872/?project=5248013)

inglesp (approved):
Great!

Issue #1449, opened 2022-11-23
Short-term solution for superceded dm+d codes


rebkwok:
Problem:
dm+d VMP IDs can change.  When they do, they have a previous VMP ID recorded against them.  If a codelist contains only the current ID, or only the previous ID, queries on `with_these_medications` will fail to match all relevant matching medications.

dm+d codelists are "old style" codelists. Updating them to new-style codelists would mean making a new version of each codelist, which has implications for any studies or individuals that might be using them.  We don't currently have a straightforward way to notify consumers of codelists.

The proposed short term solution is to leave current dm+d codelists as they are, and have cohort-extractor/databuilder add in the previous/new VMP IDs.  This includes the following tasks:
 
- [ ] Add an API endpoint to OpenCodelists that returns a mapping of current:previous VMP IDs #1460 
- [ ] Job server proxies this so that it can be retrieved on the secure server 
- [ ] The mapping is used to build a new database table/view that does the expansion of the medication table to include alternative (prev/subsequent) codes (TBC where - as a job-runner script in the first instance?)
- [ ] Update cohort-extractor methods that use the Medication table (`with_these_medications`**) to use the new expanded medication table instead
- [ ]  Similarly, in databuilder, use the new expanded medications table (currently `beta.tpp.medications`.)
- [ ] A release of dm+d only has the ability to track previous ids for previous ids if the previous ID is still a current code (otherwise it can't be looked up in the release DB).  Currently all but 2 VMP IDs have previous codes that are not still current.  We should be able to check against older releases from OpenPrescribing to see if codes change multiple times - if they do, we might still be missing codes.  Since we will now be tracking releases, we do have a way to maintain the previous code mapping from now on.  
- [ ] All dm+d codelists on OpenCodelists need to be accompanied by a disclaimer that says that there could be previous or new codes and that OpenSafely will include them
- [x] #1388 Even though we're not making dm+d codelists new-style codelists just yet, we should have a way to check newly uploaded ones to ensure they're valid and consistent with the latest dm+d coding system release (that means we can also assign a dm+d release to a new dm+d codelist version and have a record of the release at the time the version was created). 
- [ ] Regularly import dm+d data (in the short term, this can be manual, with a slack reminder).  After the import is complete, also run the script to update the medications table on the server.  If possible, automate it with the available API #1469 

[...]

PR #1460, opened 2022-12-01
API endpoint to map dm+d current VMP Ids to previous ones

### LINKED ISSUE #1470: API endpoint to map dm+d current VMP Ids to previous ones

rebkwok:
Fixes #1470

Currently builds the mapping as a list of `(current, previous)` pairs, as that's how the `dmd_previous` table looks in @Jongmassey's[ proposed SQL](https://github.com/opensafely-core/opencodelists/issues/1449#issuecomment-1327366376).  If a VMP 1 has a previous VMP 2 which also has a previous ID VPM 3, VMP 1 appears twice in the list with each of its previous ids. i.e.
```
[
    (1, 2),
    (1, 3),
    (2, 3
]
```

We have found that VMP IDs can change more than once, and there's no  record in the current release of anything further back than the one previous ID.  This means we need to keep track of all id:prev_id for each release.  We can do that now, with the releases in OpenCodelists, but for earlier ones, we need to retrieve them from OpenPrescribing.  This PR contains a script to do that, but my expectation is that it should only ever be run once, the data loaded into OCL, and then subsequent updates come from each dm+d import into OCL.  I've put the script in here though, in case it's ever needed again.

The mapping from ID:PrevID is now in a separate Mapping app, along with the code to expand it into a mapping of ID to all previous IDs, which we need for the api endpoint.  There's a mapping import_data script, which just imports the data fetched from OP, and an small extra function in the dm+d codingsystem import_data script, whcih adds the mappings after a dm+d release.


mappings/dmdvmpprevmap/README
```diff
@@ -0,0 +1,24 @@
...
+(See scripts/fetch_vmp_prev.py for re-creating/updating this file.)
+
+For releases after 12 Dec 2022, the mappings are updated when each new release is
+imported.
+
+If the historical data needs to be re-imported, it can be run with:
+
+    dokku run opencodelists python manage.py
+    import_mapping_data mapping.dmdvmpprevmap
+    /storage/data/dmdvmpprevmap/dmd_vpid_vpidprev_mapping_20221208.csv.gz
```
inglesp, line 24: If this is going to be copy/pasted, could you include backslashes so that it's handled correctly by bash?

inglesp (approved):

Issue #1463, opened 2022-12-05
Error converting BNF to dm+d


iaindillingham:
@speed-vm is trying to convert [this BNF codelist][1] into a dm+d codelist. However, OpenCodelists errors with

> CSV file contains 1465 unknown code -- the first (321112001) is on line 1 (dm+d release unknown, valid from 1900-01-01)

@richiecroker thinks the issue might relate to a new drug. For more information, see [this Slack thread][2].

---

![Screenshot 2022-12-05 at 18-01-21 OpenCodelists](https://user-images.githubusercontent.com/477263/205709157-4894f9c7-a67f-46d1-b684-dce1031f67fb.png)

[1]: https://www.opencodelists.org/codelist/user/speed-vm/medications-addictive-medicines/79636c0f/
[2]: https://bennettoxford.slack.com/archives/C31D62X5X/p1670258003662399

Jongmassey:
But those dm+d codes *do* appear to be valid, though?

richiecroker:
@Jongmassey yes they are - at least some of them are, which is not what this is saying!

rebkwok:
Fixed by #1465

Issue #1433, opened 2022-11-10
Mappings between coding systems


rebkwok:
Questions arising from the work on coding system releases:

We have two mapping apps, which map ctv3 to snomed, and bnf to dmd.  Both of these import data from a single mapping file; although it seems unlikely that there will ever be any more updates for ctv3/snomed.
https://nhs-digital.citizenspace.com/standards-assurance/mapping-tables-read-v2-snomed/

BNF to dm+d mappings are released as excel spreadsheets [here](https://www.nhsbsa.nhs.uk/prescription-data/understanding-our-data/bnf-snomed-mapping), however, there isn't any indication of which version of each coding system they relate to.

Currently, (or at least, as of #1415) the converters assume that we're always talking about the latest coding system release.  Given that there aren't likely to be more ctv3 to snomed mappings, we can probably limit concerns to BNF to dm+d mappings.

1) Can we identify a release of dm+d and BNF from the mapping files?  Or can we at least identify whether a mapping release is incompatible with any particular dm+d or BNF release (i.e. because there are codes in the mapping that aren't in a dm+d/BNF release)
2) Should we also be keeping historic releases of the mapping?
3) The mapping is used for a dm+d download of a BNF codelist; if we've been able to identify mappings that are compatible with releases of BNF/dm+d, do we want users to be able to select the dm+d release that's used for downloading?
4) Do the mappings change?  If we have the same code mapped differently in 2 mapping files, but the mapping files are compatible with multiple versions of the coding systems, which mapping do we choose?

Jongmassey:
>there isn't any indication of which version of each coding system they relate to.

there is, however, a "valid from" date, which should let us reasonably infer this

>Can we identify a release of dm+d and BNF from the mapping files?
 ...

> can we at least identify whether a mapping release is incompatible with any particular dm+d or BNF release (i.e. because there are codes in the mapping that aren't in a dm+d/BNF release)

I think we can, and this is a useful bit of info to present to users.

>Should we also be keeping historic releases of the mapping?

Probably, but at least we should record what release of the mapping a mapped codelist was created using.

Jongmassey:
>The mapping is used for a dm+d download of a BNF codelist; if we've been able to identify mappings that are compatible with releases of BNF/dm+d, do we want users to be able to select the dm+d release that's used for downloading?

The most likely path is the latest of each, I think. However, for audit purposes it would be nice to be able to map using arbitrary releases of the coding systems and mapping files.

[...]

PR #1464, opened 2022-12-06
Bugfixes for dm+d and bnfdmd mapping imports

rebkwok:
Some fixes for dm+d imports, needed to address #1463 

1) dm+d data downloads are a zip file; in the most recent downloads, one of the zip's subfolders is now also zipped, so the importer needs to unzip that too.

2) bnf to dmd mappings are released monthly; I'm not sure when the mappings were last updated in opencodelists, but the import_data command errored due to a unique constraint, because it was trying to bulk import Mapping objects without deleting the existing ones first.  Since we only use the most recent dm+d release currently (both for converting bnf codelists to dm+d, and for uploading dm+d codelists), we want the mapping to also be the most recent.  Mappings might change between releases, so i think it's safest to delete them all and start from a clean slate with the new import.  

Note that we have an issue #1433 with some questions for later around whether we should be maintaining releases of mappings as well as bnf/dm+d coding systems.

### DIFF
--- a/mappings/bnfdmd/README
+++ b/mappings/bnfdmd/README
@@ -1,0 +1,22 @@
+BNF to DM+D (SNOMED) Mappings
+------------------------------------------
+
+dm+d codes are used to identify medicines and devices in use across the NHS.
+
+Codelists are typically built using pseudoBNF and converted to dm+d for use in study definitions.
+
+For more details, see our blog posts [0] and [1].
+
+We obtain the mapping data from the NHS Business Services Authority [2].  New mapping data is released monthly.
+
+On the download page, download the latest release, which will be an zip file called something like `SNOMED - BNF mapping document September 2022.zip`, containing a single
+xlsx file.
+
+To import the data, unzip the file and run:
+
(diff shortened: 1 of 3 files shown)


mappings/bnfdmd/README
```diff
@@ -0,0 +1,23 @@
+BNF to DM+D (SNOMED) Mappings
+------------------------------------------
+
+dm+d codes are used to identify medicines and devices in use across the NHS.
+
+However, dm+d codes are not very human-readable, so codelists are typically
+build using pseudoBNF and converted to dm+d for use in study definitions.
```
inglesp, line 7: As I understand it, the main reason for constructing a BNF codelist and converting it to dm+d is that BNF has a richer structure than dm+d.

inglesp (approved):

mappings/bnfdmd/README
rebkwok: Brian said this in the blog post, but maybe I'll just refer to that :)

PR #1466, opened 2022-12-06
Add note to restart app after importing data to coding system READMEs

rebkwok:
After importing new coding system data into the prod dokku app, we need to restart the app in order for the new database connections to be picked up


inglesp (approved):
Can you make it so that there's a warning in the output of the command too?

rebkwok:
> Can you make it so that there's a warning in the output of the command too?

Good idea, will do

PR #1475, opened 2022-12-12
Skip CSV validation for coding systems without data to validate

rebkwok:
We have a couple of coding systems that we don't maintain data for, and can't validate (OPCS4 and ReadV2) - this just makes the CSV upload skip code validation for those.

### DIFF
--- a/opencodelists/forms.py
+++ b/opencodelists/forms.py
@@ -89,6 +89,13 @@ class CodelistCreateForm(forms.Form):
 
 
 def validate_csv_data_codes(coding_system, codes):
+    # Fully implemented codings systems have a `lookup_names` method that is used to
+    # validate the codes in the CSV upload.  However, we also support uploads for some
+    # coding systems that we don't maintain data for (e.g. OPCS4, ReadV2).  Skip code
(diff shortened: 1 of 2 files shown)


opencodelists/forms.py
```diff
@@ -89,22 +89,27 @@ def clean_csv_data(self):
...
-        else:
-            num = len(unknown_codes_and_ixs)
-            msg = f"CSV file contains {num} unknown code -- the first ({code}) is on line {line}"
-        msg += f" ({coding_system.short_name} release {coding_system.release_name}, valid from {coding_system.release.valid_from})"
-        raise forms.ValidationError(msg)
+    # Fully implemented codings systems have a `lookup_names` method that is used to
+    # validate the codes in the CSV upload.  However, we also support uploads for some
+    # coding systems that we don't maintain data for (e.g. OPCS4, ReadV2).  Skip code
+    # validation for these systems, and just allow upload of the CSV data as it is.
+    if coding_system.has_database:
```
evansd, line 96: I might be tempted to make this a guard clause and then you don't need to indent everything i.e.
```py
if not coding_system.has_database:
    return
```

But that's ultimately a matter of taste.

PR #1477, opened 2022-12-13
Load vmp previous mapping from the filtered queryset

rebkwok:


inglesp (approved):
Is there a low-effort thing we could do to test this?

PR #1478, opened 2022-12-13
Fetch dm+d releases with the TRUD API

rebkwok:
dm+d releases happen weekly, and currently we need to import them manually.  This lets us at least skip going to the TRUD page, downloading the release and uploading it to dokku3, by fetching it directly from the TRUD API in the import_coding_system_data command.

The `import_coding_system_data` still requires that you provide a release name and a valid-from date, so you can download and import a release that is not the latest one if you want.  If you do want the latest, you also still need to provide the release name and date, and the `--latest` flag will fetch just the latest release, and check that it matches the release/date provided.


rebkwok:
With not much more work I could make the release and date optional with the --latest flag, and extract them from the latest filename, but I'm wary of doing more that necessary at the moment.  This will save some time for the weekly updates.

inglesp (approved):

PR #1501, opened 2023-01-10
Automate weekly dm+d imports

rebkwok:
This consists of a script and template cronfile to be run as a cron job each Monday to:
1) Notify slack that dm+d import has started
2) Run the manage command to import the latest dm+d release and restart the dokku app
3) Notify slack about success/failure

We can't use the `app.json` to run the import via dokku-managed cron, because we need to also restart the app when it's done.  So once this is deployed, we'll need to manually copy the script to /var/lib/dokku/data/storage/opencodelists/ on dokku3 and add the cronfile at /etc/cron.d (including updating the SLACK_WEBHOOK_URL with the #tech-noise webhook [found in the Bennett Bot config](https://api.slack.com/apps/A03UM1N45JN/incoming-webhooks?)

### DIFF
--- a/coding_systems/dmd/README
+++ b/coding_systems/dmd/README
@@ -12,8 +12,14 @@ On the TRUD release page, subscribe to the releases.
 For local testing, you will need your API key.  Find it on your [account page](https://isd.digital.nhs.uk/trud/users/authenticated/filters/0/account/manage), and set the
 `TRUD_API_KEY` variable in your local .env file.
 
+## Automated updates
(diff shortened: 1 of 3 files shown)


Jongmassey (approved):
AFAICT looks all good to me - how is `$SLACK_WEBHOOK_URL` set to something other than `dummy-url` ?

rebkwok:
> AFAICT looks all good to me - how is `$SLACK_WEBHOOK_URL` set to something other than `dummy-url` ?

Manually - the script and cronfile need to be manually copied anyway, so when we do that, we can also update the `SLACK_WEBHOOK_URL=` line

PR #1507, opened 2023-01-17
Convert PseudoBNF codelist versions to dm+d

### LINKED ISSUE #1334: Autoconvert BNF codes to dm+d codes
In order to use `with_these_medications` in OpenSAFELY you must use a codelist in the dm+d form.

However most medications codelists are built using pseudoBNF code structure, then converted to dm+d using a feature on OpenCodelists, then downloaded and finally reuploaded to the site on a hidden page that allows editors to add raw csvs.

This takes a little time, isn't well documented and also introduces the chance of errors when someone downloads and opens with excel corrupting the identifier before uploading and a video with some of the workflow.

**Would it be possible to add an autoconvert feature?**

There is some background in this [draft blog](https://docs.opensafely.org/codelist-creation/#build-a-simple-medication-codelist) on differences between medicine terminoliges.

### LINKED ISSUE #1496: Write script to convert BNF codelists to dm+d
@kevwing is working on a study with a lot of BNF medication codelists.  He's asked for help converting these to dm+d codelists so that he can use them with `patients.with_these_medications()`.

We should be able to write a script to convert these automatically for him.  The script should take, as input, a file containing the full slug of the BNF codelist version, and create new codelists.  The new codelists should:

* be in in the same organisation as the original
* have `-dmd` appended to the codelist slug
* include a link to the original codelist version in the methodology
* have codes generated by [`mappings.bnfdmd.mappers.bnf_to_dmd`](https://github.com/opensafely-core/opencodelists/blob/main/mappings/bnfdmd/mappers.py)

The codelists that need to be converted are in the BNF tab of [this spreadsheet](https://docs.google.com/spreadsheets/d/1gB6MsAeyGyYk_uo8TPcR4RMRkNf09lpdegJWdtZgPws/edit#gid=823327680).

rebkwok:
Closes #1496
Closes #1334 

Adds:
1)  A codelist action to convert a bnf codelist version to a dm+d codelist (an old-style codelist, with `csv_data`).  This creates one codelist version which has under review status by default, with the following properties:
  - has the same owner (user or org) as the original
  - has "-dmd" appended to the name and slug
  - has the same references and description as the original
  - has a link to the BNF version it was created from in the methodology
  - does NOT copy over the signoffs from the original
  - has codes generated by mappings.bnfdmd.mappers.bnf_to_dmd (via `CodelistVersion.dmd_data_for_download`)

2) A management command that takes a CSV file with at least a "URL" column, containing the full url for BNF codelist versions to be converted, and uses the new action to create new dm+d codelists from those versions.  It 
  - handles bad urls (that don't match a CodelistVersion)
  - handles versions for which their codelist already has a converted dm+d codelist
  - generates a report CSV that can be passed on to the user.

3) A url/view for converting a BNF codelist to dm+d from the version page
  - only visible for BNF versions
  - only visible for users who can edit that BNF version
  - uses the new action to create the new codelist and redirects to it
  - creates the new codelist version as under review so the user can publish it themselves when they're happy with it
  - shows an error message if a converted dmd codelist already exists, and includes a link to it - the idea is that if they want a new version of the dm+d codelist based on a new version of the BNF codelist, they can download the dm+d data and upload it to the converted codelist).


inglesp (approved):
This is great, thank you.

Are you able to update Kevin's codelists?

rebkwok:
> Are you able to update Kevin's codelists?

Yes, will do.

inglesp:
> Is this what I think it is?!

I think it might be what you think it is!

PR #1524, opened 2023-02-07
Add a "downloadable" field to codelists api endpoint

### LINKED ISSUE #1523: Only return codelists in the /codelists api endpoint if they have valid downloadable CSV data
OSI uses the /codelists endpoint to retrieve all codelists by organisation (the only consumer of this endpoint at the moment).  It should filter out (or idenify in some way) any codelist versions that do not have valid downloadable CSV data (i.e. CSV data where a valid code column cannot be identified)

rebkwok:
Fixes #1523 

For each version returned in the codelists_get endpoint, it now returns a field indicating whether it is downloadable or not.  

A codelist version is downloadable if it is:
- not draft and
- is a new-style codelist (i.e. created with the main user upload form or builder - `csv_data` is None)
OR
- has a valid "code" header in its csv_data

Possible CSV headers for codes and term are defined per coding system on the CodingSystem instances, based on what's in the current prod database. Note that these will no longer change for code headers, but it's possible someone could upload a CSV with a different header for term/description in future.


iaindillingham (approved):
I understand what's going on here and why, so ✅.

PR #1529, opened 2023-02-07
Validate csv upload more efficiently

### LINKED ISSUE #1527: Timeout when uploading codelist
[This codelist][1] causes OpenCodelists to timeout. It contains 9,449 codes; the largest codelist on OpenCodelists contains 149,626 codes. It takes roughly 32 minutes to upload the codelist to a local instance of OpenCodelists.

The issue was first reported in [this Slack thread][2], with follow-up discussion in [this Slack thread][3].

[1]: https://github.com/opensafely-core/opencodelists/files/10676122/symptoms-pain.csv
[2]: https://bennettoxford.slack.com/archives/C010SJ89SA3/p1675704513633899
[3]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1675763648958949

rebkwok:
Fixes #1527 

With this change to the validation, the codelist mentioned in the issue takes ~4s to upload on a local instance.

### DIFF
--- a/opencodelists/forms.py
+++ b/opencodelists/forms.py
@@ -95,12 +95,8 @@ def validate_csv_data_codes(coding_system, codes):
     # validation for these systems, and just allow upload of the CSV data as it is.
     if not coding_system.has_database:
         return
-
-    unknown_codes_and_ixs = [
-        (ix, code)
-        for ix, code in enumerate(codes)
-        if code not in set(coding_system.lookup_names(codes))
-    ]
+    unknown_codes = set(codes) - set(coding_system.lookup_names(codes))
+    unknown_codes_and_ixs = [(codes.index(code), code) for code in unknown_codes]
 
     if unknown_codes_and_ixs:
         line = unknown_codes_and_ixs[0][0] + 1


iaindillingham (approved):
Mind blown!

opencodelists/forms.py
```diff
@@ -96,7 +96,9 @@ def validate_csv_data_codes(coding_system, codes):
     if not coding_system.has_database:
         return
     unknown_codes = set(codes) - set(coding_system.lookup_names(codes))
-    unknown_codes_and_ixs = [(codes.index(code), code) for code in unknown_codes]
+    unknown_codes_and_ixs = sorted(
+        [(codes.index(code), code) for code in unknown_codes]
+    )
```
iaindillingham, line 101: Does this sort on the first element of the tuple?

iaindillingham (approved):

opencodelists/forms.py
rebkwok: yes; initally I put a key, but it doesn't need one

rebkwok: Ah, but it will be faster if it uses a key, given we don't care about the second element.  I will add one :)

PR #1537, opened 2023-02-17
Validate case-insensitive unique usernames and run validation on save

rebkwok:
Ensure that usernames cannot be duplicates.  A username is still case sensitive, but validation of it is case insensitive. i.e. I can still sign up with the username "Becky", but another user cannot sign up as "becky".

The validation is now explicitly run on first `save`.  Without this, a manually created user in the shell (or elsewhere in the code, outside of a ModelForm) can be created with a duplicate username, or with an invalid slug.  (This doesn't happen in the sign up view, because the `RegisterForm` (a `ModelForm`) calls the `full_clean` method on the new user, which does run the validation.

As the username is the PK, the lack of validation meant that you could create a new user with a duplicate username, and it would overwrite the existing user.  E.g. this test was passing 

```
def test_create_user_case_insensitive_duplicate_username(organisation_user):
    assert organisation_user.username == "bob"
    user = User.objects.create_user(
        username="bob", email="newbob@example.com", name="Bob1", password="test"
    )
    assert user == organisation_user
    assert user.email == "newbob@example.com"
```

We can't run the validation on subsequent saves if the username changes, because the username is the PK, so we don't have any way (that I could find) to exclude the current object from the check for duplicates.  But we don't allow users to change their own usernames, so this shouldn't cause an issue for users.  In addition, because `email` has to be unique, you can only change a username to  a duplicate of another user if you ALSO change their email address, and it seems unlikely we'd do that in a shell for a real user.


evansd (approved):
This looks like a sensible fix. It does seem a bit weird to me that we're using username as a PK in the first place. But it looks like that decision was taken a [long time ago](https://github.com/opensafely-core/opencodelists/commit/763b6130e14d8eb357ada09c083e06c0240b7d87) and I imagine it will be quite a faff to try to change it now.

rebkwok:
> This looks like a sensible fix. It does seem a bit weird to me that we're using username as a PK in the first place. But it looks like that decision was taken a [long time ago](https://github.com/opensafely-core/opencodelists/commit/763b6130e14d8eb357ada09c083e06c0240b7d87) and I imagine it will be quite a faff to try to change it now.

Indeed, that's what I decided too.

PR #1541, opened 2023-02-23
API updates to support OSI

### LINKED ISSUE #1540: API updates to support OSI
To help OSI display useful codelist selection info to users, add the following to the codelists_get endpont:

1. Add an `updated_date` (note we discussed adding a`published_date` for all published versions, but what we actually collect is an `updated_at` date; for published versions this would usually be the published date, but it's the date the codelist was last saved (whether or not there were changes), so it's not reliably the published date. 
2. Display organisation-owned codelists only by default, and add a query param `include-users` to include user codelists

rebkwok:
Fixes #1540 
updates /api/v1/codelist endpont to:

- return organisation codelists only by default
- include user codelists if the "?include-users" query param is used
- if an owner is specified (by calling e.g. /api/v1/codelist/opensafely/), ignore any codelists for which that owner is not the current owner (previously we were getting the one codelist back in the response which used to belong to opensafely and now belongs to a user)
- Add an `updated_date` for each codelist version returned. (Note this isn't "published_date" for published versions, although in general it SHOULD date of publication, because from the interface, there's no way to modify a published version.  However, since it's an autoadd field, it'll get updated on any save, with or without changes, so it could potentially get updated after actual publication, and we have no way that I know of to verify that.)

### DIFF
--- a/codelists/api.py
+++ b/codelists/api.py
@@ -41,6 +41,7 @@ def codelists_get(request, owner=None):
 
         * coding_system_id
         * tag
+        * include-users
 
     Eg:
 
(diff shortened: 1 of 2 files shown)


codelists/api.py
```diff
@@ -77,6 +77,8 @@ def codelists_get(request, owner=None):
 
     records = []
 
+    include_user_codelists = "include-users" in request.GET
```
lucyb, line 80: The comment block above should be updated now that request.GET may contain "include-users" as well.

lucyb (approved):
Looks good. Thank you very much for doing this :)

PR #1542, opened 2023-02-24
Automate Snomed CT coding system updates

rebkwok:
Snomed CT is retreived from TRUD, which has an API, and we can fetch releases in almost exactly the same way that we already do for dm+d.

There are quite a lot of changes in this PR, but they're mostly to move the TRUD fetching and downloading code into something more generic that both dm+d and snomed can use.  The actual code to do the fetching and downloading has changed very little.


inglesp (approved):
All looks sensible, thanks!

PR #1551, opened 2023-03-07
Ignore blank lines in CSV upload

### LINKED ISSUE #1548: Handle empty rows in CSV uploads
When we read in CSV data, empty lines cause an index error here (standard upload from My Codelists page):
https://github.com/opensafely-core/opencodelists/blob/3babcb88cc79c207b6b01229e58484cbc61a9dc4/opencodelists/forms.py#L87

and here (old-style codelist uploads):
https://github.com/opensafely-core/opencodelists/blob/3babcb88cc79c207b6b01229e58484cbc61a9dc4/codelists/forms.py#L136

We should be able to handle (ignore) completely empty lines in a CSV upload.

evansd:
Previously we would blow up with an exception.

Closes #1548

### DIFF
--- a/codelists/forms.py
+++ b/codelists/forms.py
@@ -131,6 +131,9 @@ class CSVValidationMixin:
         number_of_column_errors = []
         codes = []
         for i, row in enumerate(reader, start=1):
+            # Ignore completely blank lines
+            if not row:
+                continue
             if len(row) != num_columns:
                 number_of_column_errors.append(i)
             codes.append(row[code_col_ix])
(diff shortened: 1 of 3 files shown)


rebkwok (approved):
In old-style codelists, we just store the data as it was uploaded, so if there are blank lines in the CSV data, we keep them, and any request to download as CSV will spit out exactly what was put in, blank lines and all.  Do cohort-extractor and databuilder deal with blank lines in codelists OK?

evansd:
Data Builder does, but your question prompted me to check Cohort Extractor and that doesn't so we should fix that. Thanks.

PR #1554, opened 2023-03-20
Use the right coding system for lookups in codelist diffs

### LINKED ISSUE #1553: Codelist diff should look up names in both coding system releases
Codelist diffs can take two codelist versions from different releases.  Currently only the coding system release from the LHS codelist version is used to look up names.  When comparing codelist versions from different releases, this can result in lots of "unknown" codes.

E.g. this diff shows all the codes in the LHS as "unknown"
https://www.opencodelists.org/codelist/opensafely/methotrexate-oral/041a8320/diff/0b17e9f5/

If we switch the order of the codelists in the diff, all the LHS ones have names and all the RHS are unknown
https://www.opencodelists.org/codelist/opensafely/methotrexate-oral/0b17e9f5/diff/041a8320/

The diff should use the RHS's coding system to look up terms that are found in the RHS only.
(Common codes may be a bit trickier for coding systems that have hierarchies, if the hierarchies have changed)

rebkwok:
Use the relevant coding system release to look up terms for codes in codelist versions diffs. (Previously we were using the LHS codelist version's coding system to also do the lookups for codes found only on the RHS codelist version, which meant we ended up with codes being labelled as "unknown").

See [this thread](https://bennettoxford.slack.com/archives/C31D62X5X/p1679299049381149)

Note, however, that although this was a bug, there will still be "unknown" codes in diffs.  We have a lot of codelists with an "unknown" coding system release (because they were created before we introduced the concept of releases).  That means that the coding system we look up isn't necessarily the one the codelist was created with, so there may still be some unknown codes because of that.  Also, older dm+d uploads did no verification of the codes - so if there were invalid codes in the uploaded codelist, they'll also be displayed up as unknown.

Fixes #1553

### DIFF
--- a/coding_systems/versioning/models.py
+++ b/coding_systems/versioning/models.py
@@ -39,6 +39,11 @@ class CodingSystemRelease(models.Model):
     class Meta:
         unique_together = ("coding_system", "release_name", "valid_from")
 
+    def __str__(self):
+        if self.release_name == "unknown":
+            return self.release_name
+        return f"{self.release_name} (valid from {self.valid_from.isoformat()})"
+
     def save(self, *args, **kwargs):
         # CodingSystem methods and the database router depend on the format
         # of database_alias following an expected pattern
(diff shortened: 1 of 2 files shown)


Jongmassey (approved):
Seems to cover the scenario where we know the LHS and RHS coding system + release fine

inglesp (approved):
Great

PR #1556, opened 2023-03-22
Show terms from original CSV data in diff view if unknown in coding system

### LINKED ISSUE #1555: Display terms from old CSV data in codelist diffs when lookup terms are unavailable
The diff view can be used to compare different versions of the same codelist - in particular, it's being used to look at stale dm+d codelists.  dm+d codes can be retired and are therefore not present in newer releases of the coding system.

The diff looks up terms for the codelist version's codes using its respective coding system release. However, for codelist versions that pre-date the coding system releases, the coding system release is just whatever was the latest then, not necessarily the one it was actually created with.  If a code doesn't exist in that coding system release, the term is just shown as "Unknown".  

e.g. https://www.opencodelists.org/codelist/opensafely/methotrexate-oral/041a8320/diff/0b17e9f5/

For codelist versions created by old-style CSV upload (which is all dm+d codelist versions), we can get the terms from the original upload, even if they're not valid in the coding system lookup.  i.e. that codelist version with all the unknowns in the diff is this one: https://www.opencodelists.org/codelist/opensafely/methotrexate-oral/041a8320/#full-list

It would be helpful to users to show the term from the original upload (labelled with something like "not found in coding system data").

rebkwok:
If a code from a codelist version is not found in the coding system data, its term is returned as "Unknown" in the codelist diff view.  However, some codelist versions, including all dm+d ones, are uploaded from (old-style) CSV data, which keeps a record of the literal csv data, rather than coverting it into `CodeObj` objects via the coding system data. As long as we can determine a term column from the original CSV data, we can report the term that was uploaded with the original data.  Some dm+d CSV uploads also include the dm+d type, so if we have that, we can report it too.

Terms for unknown codes are prefixed with "[Unknown]" to inform the user that the term isn't found in the coding system.

Screenshot from 2023-03-22 10-17-07

Fixes #1555

### DIFF
--- a/codelists/views/version_diff.py
+++ b/codelists/views/version_diff.py
@@ -27,16 +27,19 @@ def version_diff(request, clv, other_tag_or_hash):
     lhs_coding_system = clv.coding_system
     rhs_coding_system = other_clv.coding_system
 
+    lhs_csv_data_codes_to_terms = rhs_csv_data_codes_to_terms = None
     if lhs_coding_system.id == "dmd":
-        lhs_codes = get_dmd_codes(clv)
-        rhs_codes = get_dmd_codes(other_clv)
-        if lhs_codes is None or rhs_codes is None:
+        lhs_csv_data_codes_to_terms = get_dmd_codes_and_terms(clv)
+        rhs_csv_data_codes_to_terms = get_dmd_codes_and_terms(other_clv)
+        if lhs_csv_data_codes_to_terms is None or rhs_csv_data_codes_to_terms is None:
             raise HttpResponseBadRequest("Could not identify code columns")
+        lhs_codes = set(lhs_csv_data_codes_to_terms)
+        rhs_codes = set(rhs_csv_data_codes_to_terms)
     else:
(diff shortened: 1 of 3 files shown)


codelists/views/version_diff.py
```diff
@@ -48,27 +50,36 @@ def version_diff(request, clv, other_tag_or_hash):
...
         # dm+d has no hierarchy to look up, just return the codes themselves with their
         # terms
-        summary = [
-            {
-                "code": code,
-                "term": code_to_term[code],
-            }
-            for code in codes
-        ]
+        summary = []
```
Jongmassey, line 80: any particular reason for changing away from a list comp here?

Jongmassey (approved):
looks good!

codelists/views/version_diff.py
rebkwok: No - originally I wrote the new code for getting the term without the inner `get_term` function, and it was more readable without the list comprehension, and then I didn't put it back again.  I have now :)

PR #1558, opened 2023-03-30
Use local just commands in pre-commit

rebkwok:


.pre-commit-config.yaml
```diff
@@ -2,20 +2,14 @@ default_language_version:
...
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
-    hooks:
-    - id: isort
+    - id: check
+      name: check
+      entry: just check
+      language: system
+      types: [python]
+      require_serial: true
```
rebkwok, line 12: The `require_serial` isn't strictly needed since there's just the one local hook, but I'm leaving it in for when we come to changing to ruff

ghickman (approved):

PR #1565, opened 2023-05-24
add dm+d id Excel malformation warning

Jongmassey:
Advise not using Excel to handle dm+d lists

Provide a _safe_(ish) method to do so if really neccesary

### DIFF
--- a/userdocs/creating-a-codelist-from-a-csv-file.md
+++ b/userdocs/creating-a-codelist-from-a-csv-file.md
@@ -42,7 +42,7 @@ When you click _Create_ the codelist will be created and you will be taken to th
 From here, you can edit any metadata.
 You can also edit the codelist.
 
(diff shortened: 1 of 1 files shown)


userdocs/creating-a-codelist-from-a-csv-file.md
```diff
@@ -42,6 +42,8 @@ When you click _Create_ the codelist will be created and you will be taken to th
 From here, you can edit any metadata.
 You can also edit the codelist.
 
+We are aware of an issue whereby Excel can truncate or round dm+d IDs, turning them into invalid IDs. This is due to Excel's interpretation of this column as a number type insufficiently large to contain a dm+d ID. For this reason, opening dm+d codelist files in Excel should be avoided wherever possible. Where it is unavoidable to do so, rather than opening a dm+d codelist csv file directly in Excel (such as through the Open dialogue or from the file explorer), we reccomend opening a blank Excel workbook and using the "Import data from Text/CSV" feature. To avoid the problematic rounding/truncation behaviour described above, specify the data type of the dm+d id column as "Text" during the import process.
```
rebkwok, line 45:
```suggestion
We are aware of an issue whereby Excel can truncate or round dm+d IDs, turning them into invalid IDs. This is due to Excel's interpretation of this column as a number type insufficiently large to contain a dm+d ID. For this reason, opening dm+d codelist files in Excel should be avoided wherever possible. Where it is unavoidable to do so, rather than opening a dm+d codelist csv file directly in Excel (such as through the Open dialogue or from the file explorer), we recommend opening a blank Excel workbook and using the "Import data from Text/CSV" feature. To avoid the problematic rounding/truncation behaviour described above, specify the data type of the dm+d id column as "Text" during the import process.
```

rebkwok (approved):
Sorry I missed this one - one spelling correction suggested

PR #1566, opened 2023-06-05
Management command to create and update a draft for use in the builder

rebkwok:
This allows us to manually create a new draft that can't be generated via the builder UI.

See https://bennettoxford.slack.com/archives/C03FB777L1M/p1684936597607569

We currently can't create a new draft if the coding system has changed since the original version was created, such that the searches on the original version now bring back new codes, which will be given unknown ("?") status.

We already had an `update_draft` management command, which would fix up an **existing** draft which had fallen behind the coding system (i.e. the draft was created in the past, and the coding system had been updated since).  We can use the same mechanism to create a new draft with the current coding system release, and update any unknown codes on it.

Note that we still don't want to do this automatically, because of the potential risk of users creating a new version and not being aware of new codes that have been added to it (and assigned an implicit status) due to an updated coding system release.
The new `create_and_update_draft` management command will output the codes that have had their status updated; this can be be reported to the user so they can check and confirm them.  (Technically we could do this on creation of a draft in the builder, and report on the new/changed/removed codes there, but it requires some more consideration of the risks first)

### DIFF
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -17,6 +17,10 @@ from .search import do_search
 logger = structlog.get_logger()
 
 
+class UnknownCodeError(Exception):
+    ...
+
+
 @transaction.atomic
 def create_old_style_codelist(
     *,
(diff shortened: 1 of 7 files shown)


codelists/actions.py
```diff
@@ -573,7 +575,11 @@ def export_to_builder(*, version, author, coding_system_database_alias):
     # the builder frontend cannot deal with a CodeObj with status ?  if any of its
     # ancestors are included or excluded.  We will have to deal with this soon but for
     # now fail loudly.
-    assert not draft.code_objs.filter(status="?").exists()
+    if not ignore_unknown_statuses:
+        assert not draft.code_objs.filter(
```
evansd, line 579: Rather than an `assert` here, how easy it would it be to raise some kind of error we could bubble up to the user? If the answer is "not easy" or this happens infrequently enough that it's just not worth it then fine. I'm just thinking of things which might make our life easier next time this happens.

evansd (approved):
Thanks for all the helpful explanation here. I wouldn't have been able to fix this myself, but I get what's going on now and it all makes sense.

codelists/actions.py
rebkwok: I think it's probably easy enough, and I expect it to happen fairly often when a user tries to make a new draft from an old-ish codelist version (although I'm not entirely sure how often _that_ happens).  It'd be better than the obscure 500 though

PR #1569, opened 2023-06-21
Create new draft versions with CodeObj instances for unknown codes

### LINKED ISSUE #1563: 'Create new version' creates dumb copy
When I select 'Create new version' of one of my codelists, a new draft is created but nothing can be done with it. Please, see examples below of attempted new versions that are dumb copies. __Any ideas why this might be happening?__

image

image

image

rebkwok:
Fixes #1563 

When we create a new draft from an existing version, it reruns the searches, and if any codes are returned with unknown (?) status, it raises an error (because - according to a comment - the frontend can't deal with codes of unknown status if an ancestor code is included/excluded) and tells the user to contact tech-support.  However, in fact, the frontend can deal with these just fine.

What the frontend can't deal with is new codes that have an ancestor code that is included/excluded, and which have no CodeObj created on the draft. We now check for those in `export_to_builder` and create them with no specific status (so they will be assigned unknown).  We don't do any automatic assigning of statuses based on the parent codes, we just return the draft to the user with unresolved codes, which they will need to deal with before they can save the new draft for review.

This should avoid the need for using the management commands to manually create/update drafts, although they can still be used to automatically assign statuses based on parent codes, which might be useful if someone creates a new draft that has a lot of new codes.

The codelist in #1563 has no associated searches, and was presumably created either by uploading a CSV or by using the API. This means that the codes from the original version just get copied over onto the new draft, and rerunning searches does nothing, so no new code objs are created. The affected codelist had some new child codes which now get created as unknown codes:
Screenshot from 2023-06-21 17-04-39

### DIFF
--- a/codelists/management/commands/update_draft.py
+++ b/codelists/management/commands/update_draft.py
@@ -5,7 +5,7 @@ from django.core.management import BaseCommand
 from django.db import transaction
 from django.utils.text import slugify
 
-from ...actions import cache_hierarchy
+from ...actions import add_new_descendants, cache_hierarchy
 from ...models import CodelistVersion, CodeObj, SearchResult
 from ...search import do_search
 
@@ -65,7 +65,7 @@ class Command(BaseCommand):
             )
             self.update_searches(draft)
             self.delete_removed_codes(draft)
-            self.add_new_descendants(draft)
+            add_new_descendants(version=draft)
             self.update_code_statuses(draft)
             updates = self.diff(draft.codeset, original_codeset)
             cache_hierarchy(version=draft)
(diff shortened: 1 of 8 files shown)


codelists/management/commands/update_draft.py
```diff
@@ -155,6 +156,32 @@ def delete_removed_codes(self, draft):
                 removed_codes=old_codes,
             )
 
+    def add_new_descendants(self, draft):
```
inglesp, line 159: It would be good if this were in codelists.actions, (a) so that we could reuse code, and (b) to keep all code that manipulates state in one place.

inglesp (approved):
Past me was obviously pretty sure that this wasn't going to work, but I can't now work out why!

One question about the commit message -- by:

> If any of those codes are not created as CodeObjs on the draft...

do you mean:

> If the draft does not have a corresponding CodeObj for each of those codes...

?

rebkwok:
> One question about the commit message -- by:
> 
> > If any of those codes are not created as CodeObjs on the draft...
> 
> do you mean:
> 
> > If the draft does not have a corresponding CodeObj for each of those codes...
> 
> ?

Yes...I think so. I think what I meant was, when a new draft is created from an older version, CodeObjs are created for that draft for each code on the original version. But if the coding system has changed, the new draft may be missing some CodeObjs that are descendants of codes that were on the original.

codelists/management/commands/update_draft.py
rebkwok: I put it in here because we previous deliberately kept all the update-draft code in the management command so that when/if we decided to put it into the main app code, it would all be in one place.  But I realise that doesn't actually apply to this bit of code, since it's being used in the main app already.  I will move it

PR #1570, opened 2023-06-27
Make updating drafts more efficient

rebkwok:
Some more efficiency improvements when creating and updating new draft versions

### DIFF
--- a/coding_systems/base/coding_system_base.py
+++ b/coding_systems/base/coding_system_base.py
@@ -107,6 +107,15 @@ class BaseCodingSystem(ABC):
     def codes_by_type(self, codes, hierarchy):  # pragma: no cover
         raise NotImplementedError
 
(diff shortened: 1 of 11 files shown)


coding_systems/base/coding_system_base.py
```diff
@@ -107,6 +107,15 @@ def code_to_term(self, codes):  # pragma: no cover
     def codes_by_type(self, codes, hierarchy):  # pragma: no cover
         raise NotImplementedError
 
+    def match_codes(self, codes):
```
inglesp, line 110: Is `matching_codes` a better name?

inglesp (approved):

coding_systems/base/coding_system_base.py
rebkwok: hmm, yes, maybe

PR #1573, opened 2023-07-31
Make dmd imports more memory efficient

### LINKED ISSUE #1571: Investigate failed dm+d import
The latest dm+d release did not import.

image

We should investigate and fix.

rebkwok:
Hopefully fixes #1571 

dm+d imports have been failing since 10th July.  There's no specific error, and it runs OK locally, but the process is killed at some point during the import on dokku3. When I ran it manually and watched htop, it looks like it's maxing out memory.

This PR is an attempt to decrease memory usage by using generators etree's `iter` methods wherever possible, and avoid turning the entire loaded data files into lists.  I used `tracemelloc.get_traced_memory()` as a basic comparison (wrapped around the `import_coding_system()` call on line 57), and ran the entire import with the old and new code. Reported peak memory for the old code was 130,356,648, and for the new code (this PR) it was 186,501, so hopefully this is enough to fix the imports on dokku3 (but can't tell for sure until it's deployed).  Only ran one time each because it's slooooww.

The database created from both traced runs is identical in size, and all import data tests are passing.

### DIFF
--- a/coding_systems/dmd/import_data.py
+++ b/coding_systems/dmd/import_data.py
@@ -208,7 +208,7 @@ def import_coding_system(release_dir, database_alias):
         import_model(model, elements, database_alias)
 
     # gtin
-    elements = load_elements(filepaths["gtin"])[0]
+    elements = next(load_elements(filepaths["gtin"]))
     for element in elements:
         assert (
             element[0].tag == "AMPPID"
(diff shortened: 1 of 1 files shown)


StevenMaude (approved):
This looks good!

***

I have one slight suggestion. Could you add similar — or even just the same message — that you've nicely documented in your PR comment, in the single commit's message?

The benefits of doing that:

* it's instantly visible in the actual repository itself to someone doing `git show` or similar, without having to look up the related PR
* it potentially has some permanency (assuming branch protection), whereas PR comments could, while unlikely, get edited/deleted
* it's not tied to GitHub's platform; in case this code ever did live somewhere else, where the PR comments may not get migrated

PR #1574, opened 2023-08-10
Update bash scripts in `justfile`

StevenMaude:


justfile
```diff
@@ -78,6 +86,9 @@ _env:
 #
 # ensure dev requirements installed and up to date
 devenv: _env prodenv requirements-dev && install-precommit
+    #!/usr/bin/env bash
```
madwort, line 89: this one is interesting! this line https://github.com/opensafely-core/opencodelists/pull/1574/files#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1L82 has maybe not been working as expected?

madwort (approved):

PR #1611, opened 2023-08-18
Upgrade to Django 4.2

rebkwok:
- Adds the new STORAGES setting
- Add a test to reassure me that the [changes to update_or_create](https://docs.djangoproject.com/en/4.2/releases/4.2/#setting-update-fields-in-model-save-may-now-be-required) in 4.2 aren't a problem
- minor fixes to get rid of new deprecation warnings


inglesp (approved):
Thanks for keeping on top of this!

PR #1633, opened 2023-08-25
Call tech-support in slack for failed automated imports

rebkwok:
A few weeks ago, dm+d automated import started to fail, but we didn't notice for a while because the failure notifications were posted into #tech-noise.  This makes them call tech-support for investigation.

Note that snomed isn't released on a very regular schedule (it's around every 2 months, but sometimes has interim minor releases), so we try to import weekly, but we expect it to fail about 9/10 times, when it doesn't find a new release to import. We can probably fix that in the code, but for now, just notify tech support with the urls to check.

### DIFF
--- a/deploy/bin/import_latest_snomedct.sh
+++ b/deploy/bin/import_latest_snomedct.sh
@@ -35,6 +35,6 @@ if [ $RESULT == 0 ] ; then
 else
   #  Post failure message to slack
   curl -X POST -H 'Content-type: application/json' \
-    --data '{"text":"Latest snomedct release failed to import to OpenCodelists; this is probably because no new release was found."}'\
+    --data '{"text":"Latest snomedct release failed to import to OpenCodelists; this is probably because no new release was found. tech-support please check latest SNOMED-CT release at https://www.opencodelists.org/coding-systems/latest-releases against https://isd.digital.nhs.uk/trud/users/authenticated/filters/0/categories/26/items/101/releases"}'\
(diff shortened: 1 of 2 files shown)


deploy/bin/import_latest_snomedct.sh
```diff
@@ -35,6 +35,6 @@ if [ $RESULT == 0 ] ; then
 else
   #  Post failure message to slack
   curl -X POST -H 'Content-type: application/json' \
-    --data '{"text":"Latest snomedct release failed to import to OpenCodelists; this is probably because no new release was found."}'\
+    --data '{"text":"Latest snomedct release failed to import to OpenCodelists; this is probably because no new release was found. tech-support please check latest SNOMED-CT release at https://www.opencodelists.org/coding-systems/latest-releases against https://isd.digital.nhs.uk/trud/users/authenticated/filters/0/categories/26/items/101/releases"}'\
```
iaindillingham, line 38: Could you add an entry to [the playbook][1] or [the team manual][2] that describes what someone on tech support should check. The NHS page is a login form; should I create an account?

[1]: https://docs.google.com/document/d/1Lln5N8p3hGtD3kLHJHJMWB64aTnOGpQZmqEt7ZE6d98/edit#heading=h.7wwb8hlbbnu2
[2]: https://bennettinstitute-team-manual.pages.dev/tech-group/tech-support/

iaindillingham (approved):

PR #1652, opened 2023-09-07
Improve automated import scripts

rebkwok:
Refactor dmd and snomed import scripts into one
Redirect output from running the dokku import command into a log file
If the import command failed, persist the log file and tell tech-support where to find it.
Check the output in case the failed command was just because the latest release already exists (this happens about 9/10 times for snomed) and don't bother tech-support about those.


deploy/bin/import_latest_release.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# NOTE: this script is run by cron (as the dokku user) weekly
+# For dm+d, it is run every Monday noight, to co-incide with weekly
```
madwort, line 6:
```suggestion
# For dm+d, it is run every Monday night, to co-incide with weekly
```

inglesp (approved):
Am not remotely qualified to review the bash, but this looks pretty sensible.

Issue #1612, opened 2023-08-18
Import latest PRIMIS codelists


rebkwok:
See https://bennettoxford.slack.com/archives/C03FB777L1M/p1691509895377489

rebkwok:
New codelist versions created as v2.0 and currently under review

PR #1661, opened 2023-09-12
Import/update codelists from xlsx file

rebkwok:
This is specifically intended to address #1612 and so is designed to import codelists from the xlsx file provided by PRIMIS . We've apparently done this before, with the same format spreadsheet, so this script means we should be able to do it again in future if/when they update these again.  There's also some attempt to make it more generic for bulk importing other codelists.

This calls the API endpoints on prod, so for peace of mind, it runs by default it runs against a locally running server, and doesn't actually do anything other than report the codelists/versions it will create.  The idea is that if you have a copy of the prod db, and the local coding system dbs for the right releases, you can run the whole thing locally to check everything works OK before running for real.


codelists/scripts/import_codelists_from_xlsx.py
```diff
@@ -0,0 +1,333 @@
...
+    print(user.api_token)
+
+Run with:
+    API_TOKEN=###
+    python codelists/scripts/import_codelists_from_xlsx.py
+      <path/to/xlsx>
+      <path/to/config/json>
+      --host # run on the specified host; defaults to localhost:8000
+      --force-new-version -f  # to force a new version, even if there are no changes
+      --live-run  # run for real; if not set, just reports what would be done
```
inglesp, line 21: This is excellent.

codelists/scripts/import_codelists_from_xlsx.py
```diff
@@ -0,0 +1,333 @@
...
+    arguments = parser.parse_args()
+    config = read_config(arguments.config_file)
+
+    return (
+        arguments.xlsx_file,
+        config,
+        arguments.host,
+        arguments.force_new_version,
+        not arguments.live_run,
+    )
```
inglesp, line 329: Perhaps worth adding a test that this is not run live by default.  How paranoid are we feeling?

inglesp (approved):

PR #1663, opened 2023-09-13
Add note about prod host for import_codelists_from_xlsx script

rebkwok:
API calls to https://opensafely.org are redirected to https://www.opensafely.org which results in 405s from endpoints that only allow POST


inglesp (approved):

rebkwok:
I had forgotten this one was still open! Thanks @inglesp

PR #1672, opened 2023-09-20
Include mapped VMPs in dm+d CSV downloads

rebkwok:
Includes mapped VMPs in dm+d CSV downloads by default (they can be omitted with a query parameter flag ?omit-mapped-vmps)

This means that running `opensafely codelists update` on a study will download CSVs that include mapped VMPs (and  note in the downloaded CSV which VMP they map to).  The study will have codelists with mapped VMPs that it can use, without modifying CSV files that have already downloaded (which would cause OS CLI's codelist checks to fail), and we don't have to create new dm+d versions that circumvent the current coding system validation.

Example download from the opiods study that currently having problems with mapped VMPs:
[opensafely-opioid-containing-medicines-parenteral-excluding-drugs-for-substance-misuse-dmd.csv](https://github.com/opensafely-core/opencodelists/files/12731152/opensafely-opioid-containing-medicines-parenteral-excluding-drugs-for-substance-misuse-dmd.csv)

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -565,20 +565,21 @@ class CodelistVersion(models.Model):
         term_header_index = header_row.index(term_header) if term_header else None
 
         table_rows = self.table[1:]
-        if include_mapped_vmps:
+        if include_mapped_vmps and self.coding_system_id == "dmd":
+            term_header_index = term_header_index or len(self.table[0])
+            # ignore include_mapped_vmps if coding system is anything other than dmd
             codes = [row[code_header_index] for row in table_rows]
-            assert self.coding_system_id == "dmd"
             # add in mapped VMP codes
             previous_vmps_for_this_codelist = vmp_ids_to_previous(codes)
-            mapped_vmps_to_add = set()
             for current_vmp, previous_vmp in previous_vmps_for_this_codelist:
                 assert current_vmp in codes
                 if previous_vmp not in codes:
-                    mapped_vmps_to_add.add(previous_vmp)
-            for mapped_vmp in mapped_vmps_to_add:
-                new_row = ["" for cell in table_rows[0]]
-                new_row[code_header_index] = mapped_vmp
-                table_rows.append(new_row)
+                    new_row = ["" for i in range(len(table_rows[0]) + 1)]
+                    new_row[code_header_index] = previous_vmp
+                    new_row[
+                        term_header_index
+                    ] = f"Mapped previous VMP for {current_vmp}"
+                    table_rows.append(new_row)
 
         # re-write the table data with the new headers, and only the code and term columns
         table_data = [
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -1,3 +1,5 @@
+from collections import Counter
+
 from django.contrib.auth.models import AnonymousUser
 from django.db import models
 from django.urls import reverse
@@ -576,9 +578,27 @@ class CodelistVersion(models.Model):
             )
             # vmp_to_prev_mapping is a simple 1:1 mapping of vmp ID to the immediate previous
             # VMP that it replaced
+
+            # Count if there were any duplicate previous VMPs in the mapping
+            counter = Counter(vmp_to_prev_mapping.values())
+            _, count = counter.most_common(1)[0]
+
             # Since we're mapping later codes as well as previous ones, we also need the
             # reverse mapping of previous VMP to the one that replaced it
             prev_to_vmp_mapping = {v: k for k, v in vmp_to_prev_mapping.items()}
+            # If there were any duplicate previous VMPs, we join them in the reversed mapping
+            # This is only needed for the description of where the mapped code comes from
+            if count > 1:
+                duplicate_mappings = [
+                    vmp for vmp, count in counter.items() if count > 1
+                ]
+                for mapped_vmp in duplicate_mappings:
+                    mapped_to = [
+                        vmp
+                        for vmp, prev in vmp_to_prev_mapping.items()
+                        if prev == mapped_vmp
+                    ]
+                    prev_to_vmp_mapping[mapped_vmp] = ", ".join(mapped_to)
 
             previous_vmps_to_add = set()
             subsequent_vmps_to_add = set()
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -5,7 +5,7 @@ from django.utils.functional import cached_property
 from taggit.managers import TaggableManager
 
 from mappings.bnfdmd.mappers import bnf_to_dmd
-from mappings.dmdvmpprevmap.mappers import vmp_ids_to_previous
+from mappings.dmdvmpprevmap.mappers import vmpprev_full_mappings
 from opencodelists.csv_utils import (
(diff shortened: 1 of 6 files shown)


codelists/models.py
```diff
@@ -562,12 +563,30 @@ def table_with_fixed_headers(self):
...
+            # add in mapped VMP codes
+            previous_vmps_for_this_codelist = vmp_ids_to_previous(codes)
+            for current_vmp, previous_vmp in previous_vmps_for_this_codelist:
+                assert current_vmp in codes
+                if previous_vmp not in codes:
+                    new_row = ["" for i in range(len(table_rows[0]) + 1)]
+                    new_row[code_header_index] = previous_vmp
+                    new_row[
+                        term_header_index
+                    ] = f"Mapped previous VMP for {current_vmp}"
```
rebkwok, line 581: Not sure this is the best wording.  "Superceded by {current_vmp}" maybe?

Jongmassey: The nomenclature in the source file and its metadata is "previous" ( e.g. `VPIDPREV` )

codelists/models.py
```diff
@@ -562,12 +563,30 @@ def table_with_fixed_headers(self):
         # Identify the index for the two columns we want
         code_header_index = header_row.index(code_header)
         term_header_index = header_row.index(term_header) if term_header else None
+
+        table_rows = self.table[1:]
+        if include_mapped_vmps and self.coding_system_id == "dmd":
+            term_header_index = term_header_index or len(self.table[0])
+            # ignore include_mapped_vmps if coding system is anything other than dmd
```
Jongmassey, line 570: nit - should this comment go up a line?

codelists/models.py
```diff
@@ -562,12 +563,30 @@ def table_with_fixed_headers(self):
...
+            codes = [row[code_header_index] for row in table_rows]
+            # add in mapped VMP codes
+            previous_vmps_for_this_codelist = vmp_ids_to_previous(codes)
+            for current_vmp, previous_vmp in previous_vmps_for_this_codelist:
+                assert current_vmp in codes
```
Jongmassey, line 575: Surely we want the mapped codes where the previous code is in the list but not the current one (i.e. the dm+d has been updated since this list's creation) ?

Jongmassey (commented):
Generally looks good but I think I might not fully understand how this addresses the situation where there is a codelist with old/superseded VMP IDs and we want to download a version with the new and old IDs in there?

codelists/models.py
Jongmassey: Yeah, to replicate the cohort-extractor behaviour we need to back to the future

codelists/models.py
rebkwok: Users don't necessarily know or need to know that though, do they?

rebkwok:
@Jongmassey This should now be doing future as well as past mapping (I also updated the example CSV in the description)

codelists/models.py
Jongmassey: No, but (maybe just because I've been immersed in it!) I think that terminology is sufficiently clear and there is merit to matching what other sources call it

Jongmassey: Actually reviewing the code please disregard my comment I like what you've done there!

mappings/dmdvmpprevmap/tests/test_mappers.py
```diff
@@ -41,4 +44,29 @@ def test_codes_to_previous_with_self_previous():
...
+    # return previous for selected codes only
+    Mapping.objects.create(id="1", vpidprev="0")
+    Mapping.objects.create(id="2", vpidprev="1")
+    Mapping.objects.create(id="3", vpidprev="2")
+    Mapping.objects.create(id="4", vpidprev="3")
+    Mapping.objects.create(id="5", vpidprev="6")
+
+    _, vmp_to_previous_tuples = vmpprev_full_mappings(codes=["1", "2"])
+
+    assert sorted(vmp_to_previous_tuples) == [
```
Jongmassey, line 61: This seems to cover all the cases well

Jongmassey (approved):
Looks good to me and should fix the issue in question!

codelists/models.py
```diff
@@ -562,12 +563,60 @@ def table_with_fixed_headers(self):
...
+            def add_row(vmp, description):
+                new_row = ["" for i in range(len(table_rows[0]) + 1)]
+                new_row[code_header_index] = vmp
+                new_row[term_header_index] = description
+                table_rows.append(new_row)
+
+            for vmp in previous_vmps_to_add:
+                # add the code to the table data
+                # include its description as the code it was superceded by
+                add_row(vmp, f"VMP previous to {prev_to_vmp_mapping[vmp]}")
```
inglesp, line 607: Is there any way we could include the name of the previous VMP?

inglesp (approved):

codelists/models.py
```diff
@@ -566,20 +566,50 @@ def table_with_fixed_headers(self, include_mapped_vmps=True):
...
-                    ] = f"Mapped previous VMP for {current_vmp}"
-                    table_rows.append(new_row)
+            vmp_to_prev_mapping, mapped_vmps_for_this_codelist = vmpprev_full_mappings(
+                codes
+            )
+            # vmp_to_prev_mapping is a simple 1:1 mapping of vmp ID to the immediate previous
+            # VMP that it replaced
+            # Since we're mapping later codes as well as previous ones, we also need the
+            # reverse mapping of previous VMP to the one that replaced it
+            prev_to_vmp_mapping = {v: k for k, v in vmp_to_prev_mapping.items()}
```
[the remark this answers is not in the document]
rebkwok: The mapping is invertible, and that assertion will pass right now.  However, I'm not convinced that it always will - it could be possible that a VMP code is replaced by 2 new codes. I'm not sure how the retrospective updating would work in that case, so maybe it never happens, but I think it's worth checking and dealing with it if it does.

codelists/models.py
```diff
@@ -562,12 +563,60 @@ def table_with_fixed_headers(self):
         # Identify the index for the two columns we want
         code_header_index = header_row.index(code_header)
         term_header_index = header_row.index(term_header) if term_header else None
+
+        table_rows = self.table[1:]
+        if include_mapped_vmps and self.coding_system_id == "dmd":
```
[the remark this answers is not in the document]
rebkwok: I took the assert out because it's now True by default for any codelist, no matter what their coding system

codelists/models.py
inglesp: The tests made me think that we're not including the name -- eg "VMP previous to 10514511000001106"

rebkwok: Oh, I see, you mean the term as well as the code? 
It depends; we'd have to look up the term in the right release of dm+d.  We only have the ability to do that for releases we've imported since the time when we introduced coding system releases into OpenCodelists.  There's not an easy way of telling which release the mapped code is from (we just know it's not in the release that this codelist was made with, and we know it is in one that's either older or newer, depending on whether the mapped code is a previous or future one), so we'd have to try them until we found one that contained it, which could be expensive.

PR #1696, opened 2023-10-19
Allow mapping of previous VMPs more than one step from codelist codes

rebkwok:
VMP codes in the dm+d coding system can change; the imported data provides us with a mapping at each release of current VMP to previous VMP, and OpenCodelists maintains a list of those mappings and updates it with each release.

dm+d CSV downloads automatically add in codes for VMPs that were previously (or subsequently) mapped to any code in the codelist.

e.g. Assume we have a codelist `["a", "b", "c"]`
And in the mappings we have:
```
("a", "d"),
("b", "e"),
("e", "f")
```
So VMP "a" replaced VMP "d"; VMP "b" replaced VMP "e", and sometime in the past VMP "e" replaced VMP "f".

When we download this codelist we have to map in any codes that map to any code in the codelist, even if they're more than one step away.  In this case, we need to map in "d" and "e", which are immediately predecessors of codelist codes, but also "f", which isn't an indirect predecessor of "b".

In the CSV download, we also report which code in the current codelist the mapped-in VMP code is related to (and include it as the description for the code). 
The current code already does this, but there was a bug in the lookup for the description if the previous mapped code wasn't a direct predecessor of a codelist code.

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -591,7 +591,12 @@ class CodelistVersion(models.Model):
                 codes
             )
             # vmp_to_prev_mapping is a simple 1:1 mapping of vmp ID to the immediate previous
-            # VMP that it replaced
+            # VMP that it replaced. It ONLY contains mappings where one of the pair is in
+            # the codelist
+
+            # mapped_vmps_for_this_codelist is a full list of (vmp, previous) tuples, where one of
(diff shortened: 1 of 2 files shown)


evansd (approved):
Thanks for the clear and helpful explanation of the problem this is fixing.

PR #1707, opened 2023-10-30
Refactor csv download code for mapped VMPs

rebkwok:
Discovered another small bug in how the mapped VMP descriptions were produced, and realised that the previous code was ridiculously overcomplicated.  The first commit refactors the way the mapped-VMP descriptions are generated, the second one removed the extra code that was only there for the overcomplicated previous version.

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -597,9 +597,7 @@ class CodelistVersion(models.Model):
             codes = [row[code_header_index] for row in table_rows]
 
             # add in mapped VMP codes
-            vmp_to_prev_mapping, mapped_vmps_for_this_codelist = vmpprev_full_mappings(
-                codes
-            )
+            mapped_vmps_for_this_codelist = vmpprev_full_mappings(codes)
 
             # mapped_vmps_for_this_codelist is a full list of (vmp, previous) tuples, where one of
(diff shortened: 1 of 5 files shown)


codelists/models.py
```diff
@@ -601,68 +600,28 @@ def table_with_fixed_headers(self, include_mapped_vmps=True):
...
                 if vmp in codes:
-                    # mapping a code in the codelist to a previous code
-                    # add the previous code to the list
-                    previous_vmps_to_add.add(previous_vmp)
-                    # the previous vmp may not be in the prev_to_vmp_mapping mapping
-                    # yet, if it was more than one step away from the code in the codelist.
-                    if previous_vmp not in prev_to_vmp_mapping:
-                        prev_to_vmp_mapping[previous_vmp] = vmp
+                    # mapping in a previous code
+                    previous_vmps_to_add.setdefault(previous, []).append(vmp)
```
iaindillingham, line 617: I'd have reached for `collections.defaultdict` here. I didn't know about `dict.setdefault` 👍🏻

iaindillingham (approved):
I've only been able to give a cursory review, I'm afraid, because I'm not familiar with the codebase. However, there's less code for the same tests, which strongly suggests everything's in order 🙂

PR #1746, opened 2023-11-15
Cache csv download data and shas

### LINKED ISSUE #1740: OpenCodelists Freshping checks are flapping
It's been going up and down every few minutes from [11.30 on Wed 15 Nov](https://bennettoxford.slack.com/archives/C0270Q313H7/p1700047810165249).

### LINKED ISSUE #1744: Cache CSV download shas
The opensafely CLI, study github actions and jobserver now all call the new codelist check endpoing /api/v1/check. Mostly this is a pretty fast call, but for very large dm+d codelists, and/or studies that have a lot of codelists, especially dm+d ones, it can be slow, and it seems to be the cause of tech-support issue #1740 

We can cache these shas so we don't have to go through the VMP mapping checks every time.  dm+d is release once a week, so at most we need to re-cache weekly, and we may be able to do that as part of the data import too.

rebkwok:
Fixes #1744 
And hopefully also fixes #1740

The opensafely CLI, study github actions and jobserver now all call the new codelist check endpoing /api/v1/check. Mostly this is a pretty fast call, but for very large dm+d codelists, and/or studies that have a lot of codelists, especially dm+d ones, it can be slow, and it seems to be the cause of tech-support issue https://github.com/opensafely-core/opencodelists/issues/1740

We can cache these shas so we don't have to go through the VMP mapping checks every time. dm+d is release once a week, so at most we need to re-cache weekly. There are under 700 dm+d codelist versions that would need to be re-cached, and locally this is quick to do, so it can be done as part of the weekly import.

Since caching the shas involves fetching the CSV data for downloading, we can also cache that.  It's caching with a key that includes the function parameters so that we can cache different calls (e.g. OSI will always pass `fixed_headers=True`)


inglesp (approved):
I think this is ok, but there's a risk of a (harmless, as far as I can tell) race when a codelist is downloaded at the same time as the caches are being built after import.

Do we need to rebuild the caches after import?  Or can we do the caching on demand?

rebkwok:
> I think this is ok, but there's a risk of a (harmless, as far as I can tell) race when a codelist is downloaded at the same time as the caches are being built after import.
> 
> Do we need to rebuild the caches after import? Or can we do the caching on demand?

We don't _need_ to rebuild the caches after import, but I was concerned that if we don't, we'll still sometimes get the behaviour we saw yesterday on OpenCodelists, where the check endpoint made everything go slow.  Given that it takes a minimal time to re-cache 600+ dm+d codelist versions though, it's possible that dm+d isn't actually the problem, it's more studies that have huge numbers of codelists in general.

However, there's also a chance of a race if two users try to download a codelist that needs cached at the same time (also CI runs/job-server checks, since the check endpoint will also re-cache if necessary).  The import runs automatically at about 2am, so there's probably less risk of a race with users if it re-caches then.

I could make the re-cache after import an optional flag, so it does it in the automated 2am import, but not if we re-run an import manually (more likely to be during working hours)

inglesp:
> I could make the re-cache after import an optional flag,

I don't think it's worth adding the complexity for this.

PR #1761, opened 2023-12-08
Include original csv data columns with default dm+d downloads

### LINKED ISSUE #1760: dm+d default downloads need to include category columns
See https://bennettoxford.slack.com/archives/C03FB777L1M/p1701966725580019

By default, dm+d downloads now include standardised column headings ("code", "term") plus a duplicate code column that is labelled with the original column heading, to [avoid breaking studies](#https://github.com/opensafely-core/opencodelists/pull/1727) that used the old-style download (which was just a dump of the CSV data uploaded to create the codelist version in the first place).

However, users sometimes use the csv-upload option to create codelist versions that have extra columns that they need - specifically category columns.  OpenCodelists doesn't care, as long as the upload contains a valid and identifiable column for the codes.  This means that dm+d downloads also need to include any original columns from the CSV data. Since we can't tell which of those might be wanted by a user, we just want to add them all in after mapping in any previous/subsequent VMPs.

rebkwok:
Fixes #1760 

dm+d downloads were previously downloading with just a code and term column by default, and mapping in any previous/subsequent VMPs that they needed to (with a note in the term column to describe where they were mapped from/to). However, in some cases, a study uses other columns from the original uploaded csv data - particularly category columns. OpenCodelists doesn't know or care about category columns, so we just add in any other original columns to the csv download.

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -591,26 +591,69 @@ class CodelistVersion(models.Model):
     def _get_dmd_shas(self, shas, current_csv_data_download):
         """
         Return valid shas for a dm+d codelist CSV download
+
+        This is used to check whether an existing downloaded CSV for this
+        dm+d version (i.e. a download in a study repo) is still up-to-date.
+        dm+d VMPs can change with each weekly dm+d release, so downloads
+        need to include any mapped VMPs, in addition to the original ones in
+        the codelist. Due to historic changes in the way the default download
+        is formatted, there are multiple formats of the download that could
+        be considered valid, as they include all relevant mapped VMPs at the
+        time of this release.  These include:
+        - "old-style" downloads - these were just a copy of the uploaded `csv_data`,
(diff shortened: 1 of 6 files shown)


codelists/models.py
```diff
@@ -649,24 +687,33 @@ def csv_data_shas(self):
...
+        include_mapped_vmps: for dm+d codelists, also map in any previous/subsequent
+        VMPs
         """
         assert self.downloadable
         header_row = [header.lower() for header in self.table[0]]
         # Find the first matching header from the possible code and term column headers for this
         # codelist's coding system.  These are listed in order of assumed most-to-least likely,
         # in case of multiple matching headers
-        code_header = next(
+        original_code_header = next(
```
Jongmassey, line 711: nice - I'd never have thought of using `next()` like that - TIL!

PR #1776, opened 2023-12-18
Run as non-root user

rebkwok:
Trying to do the run-as-non-root user properly this time.

- unrevert the previous commits to run as non-root
- update dockerfile/docker-compose following job-server example
- add smoketest for prod image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -149,6 +149,10 @@ ARG GITREF=unknown
 LABEL org.opencontainers.image.revision=$GITREF
 
 
+ARG USERID=10003
+ARG GROUPID=10003
+USER ${USERID}:${GROUPID}
+
 ##################################################
 #
 # Dev image
@@ -167,3 +171,8 @@ RUN --mount=type=cache,target=/root/.cache \
 
 # Override ENTRYPOINT rather than CMD so we can pass arbitrary commands to the entrypoint script
 ENTRYPOINT ["/app/docker/entrypoints/dev.sh"]
+
+# Run as non root user. Required when building image.
+ARG USERID
+ARG GROUPID
+USER ${USERID}:${GROUPID}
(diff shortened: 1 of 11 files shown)


rebkwok:
Not currently passing the smoketest because of gunicorn errors when trying to run with the prod image
https://github.com/opensafely-core/opencodelists/actions/runs/7252571849/job/19757343759

docker/Dockerfile
```diff
@@ -148,6 +183,10 @@ LABEL org.opencontainers.image.created=$BUILD_DATE
 ARG GITREF=unknown
 LABEL org.opencontainers.image.revision=$GITREF
 
+ARG USERID=10003
+ARG GROUPID=10003
+RUN chown -R ${USERID}:${GROUPID} /app
```
rebkwok, line 188: I had to do this, otherwise I got the "can't write to a readonly database" sqlite error again.

docker/docker-compose.yaml
```diff
@@ -1,33 +1,41 @@
...
+          # this makes the image work for later cache_from: usage
+          - BUILDKIT_INLINE_CACHE=1
+          # env vars supplied by just
+          - BUILD_DATE
+          - GITREF
     # use dockers builitin PID daemon
     init: true
     ports:
-      - "8000:8000"
+      - "8000:7000"
```
bloodearnest, line 25: I have mild preference to map ports through 1:1, not change them if possible, so would prefer `7000:7000`. I think it reduces confusion about what port something is running on if its the same inside and outside docker.

But I may well be missing something.

rebkwok: I initially did that, but then changed it back so that it doesn't change under anyone who's working on opencodelists locally.  But that's probably only me really, so :woman_shrugging:

bloodearnest: Ah, right. To me, the ability to run both the docker version *and* the non-docker version simultaneously feels like a massive footgun. I almost quite like the simple port limit meaning they can not be run simultaneous. Especially if they are using the same db!

rebkwok: I have updated it for docker and non-docker local servers

bloodearnest (approved):

PR #1785, opened 2023-12-22
Ensure prod smoketest can create/read db file

rebkwok:
In actual prod, we mount a volume that is owned by the non-root docker user. In the prod smoketest, we need to make sure that the volume mounted in is also owned by the non-root user. The db doesn't exist at this stage, so it will be created in the test.

### DIFF
--- a/docker/justfile
+++ b/docker/justfile
@@ -53,6 +53,13 @@ test: test-js test-py
 
 # run dev server in docker container
 serve env="dev" *args="":
+    #!/usr/bin/env bash
+    if [ "$env" = "prod" ]; then
+        echo foo
+        SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+        mkdir "$SCRIPT_DIR/storage"
+        sudo chown 10003:10003 "$SCRIPT_DIR/storage"
+    fi
     docker-compose up {{ args }} {{ env }}
 
 
--- a/docker/justfile
+++ b/docker/justfile
@@ -51,15 +51,19 @@ test-js:
 
 test: test-js test-py
 
-# run dev server in docker container
-serve env="dev" *args="":
+_create_storage env:
     #!/usr/bin/env bash
-    if [ "$env" = "prod" ]; then
-        echo foo
-        SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+    set -euo pipefail
+
+    if [ "{{ env }}" = "prod" ]; then
+        SCRIPT_DIR={{justfile_directory()}}
         mkdir "$SCRIPT_DIR/storage"
         sudo chown 10003:10003 "$SCRIPT_DIR/storage"
     fi
+
+
+# run dev server in docker container
+serve env="dev" *args="": (_create_storage env)
     docker-compose up {{ args }} {{ env }}
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ static/dist/js
 staticfiles/
 /assets/dist
 /docker/docker-staticfiles/*
+docker/storage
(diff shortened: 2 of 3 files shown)


docker/justfile
```diff
@@ -53,6 +53,13 @@ test: test-js test-py
 
 # run dev server in docker container
 serve env="dev" *args="":
+    #!/usr/bin/env bash
+    if [ "$env" = "prod" ]; then
```
bloodearnest, line 57:
```suggestion
    if [ "{{ env }}" = "prod" ]; then
```

bloodearnest: just command args are not exposed as bash variables, sadly.

docker/justfile
```diff
@@ -53,6 +53,13 @@ test: test-js test-py
...
+        echo foo
+        SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
bloodearnest, line 59:
```suggestion
        SCRIPT_DIR={{justfile_directory()}}
```
I think this is nicer way to locate the directory?

docker/justfile
```diff
@@ -53,6 +53,13 @@ test: test-js test-py
...
+        mkdir "$SCRIPT_DIR/storage"
+        sudo chown 10003:10003 "$SCRIPT_DIR/storage"
+    fi
     docker-compose up {{ args }} {{ env }}
```
madwort, line 63: some suggestions for justfile style - you have to use bash in order to do the test, but then you have this thing that the behaviour subtly changes if one command fails in a justfile recipe. The just default is I think pretty much `pipefail` (i.e. if one command fails, the whole recipe fails), but when you switch to bash it switches to the bash default (i.e. if one command fails, the recipe carries on). So, at the moment we're pretty much adding  `set -euo pipefail` anytime we have `#!/usr/bin/env bash`, like this:

```bash
serve env="dev" *args="":
    #!/usr/bin/env bash
    set -euo pipefail
    if [ "$env" = "prod" ]; then
    ...etc...
```

The other thing is a purely optional suggestion, which is to use a [private recipe](https://just.systems/man/en/chapter_49.html) & [pass the variables as arguments to the recipe](https://github.com/casey/just#recipe-parameters), which would look something like this:

```suggestion
_create_storage env:
    #!/usr/bin/env bash
    set -euo pipefail

    if [ "$env" = "prod" ]; then
        echo foo
        SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
        mkdir "$SCRIPT_DIR/storage"
        sudo chown 10003:10003 "$SCRIPT_DIR/storage"
    fi

serve env="dev" *args="": (_create_storage env)
    docker-compose up {{ args }} {{ env }}
```

docker/justfile
rebkwok: I mean, it works locally, which suggests that they are

docker/justfile
rebkwok: Did not know this was a thing :+1:

docker/justfile
bloodearnest: Oh! TIL

docker/justfile
madwort: for me (on macos) the original `SCRIPT_DIR` points to `/var/folders/y8/455d21sn4zb237jyyl3q6z3c0000gn/T/justByxnRZ`, whereas @bloodearnest 's suggestion points to what I think is the expected `/Users/tom/Documents/bennett-institute/opencodelists/docker`

docker/justfile
madwort: I'm trying this locally, and for me it silently fails, i.e. doesn't do anything. When I include `set -euo pipefail` (see below) it fails like this:
 
```
env: unbound variable
error: Recipe `_create_storage` failed with exit code 1
```

madwort (approved):

docker/justfile
madwort: 🤔 it might be that you were using `$env` from out of your environment somewhere (e.g. in `.env`), instead of the just recipe argument.

PR #1843, opened 2024-02-07
Convert React app to components, add prop-types

tomodwyer:
...and improve the ESLint config.

Also includes a quick fix for the more modern Plausible script.


iaindillingham (approved):
Thanks for breaking this PR into a nice sequence of commits 👏🏻

PR #1855, opened 2024-02-15
Add initial `.dockerignore`

StevenMaude:
This is somewhat inspired by:

https://github.com/opensafely-core/job-server/blob/13e2eaa97849fc3766b60360ce229b84f8532a80/.dockerignore

and

https://github.com/opensafely-core/repo-template/blob/76493f62166372738dd7b3f75f997bde75293571/.dockerignore

I've chosen to:

* exclude `*.pyc` and `__pycache__/` anywhere in the repository
* be more selective about specifying directories with a trailing slash, where I think directories are the intended target of the ignore rule (without the trailing slash, files or directories can be matched)


StevenMaude:
As @madwort pointed on Slack, we should do something about not including the `.env`.

iaindillingham (approved):

PR #1903, opened 2024-04-02
Remove use of openpyxl

StevenMaude:
This is the only use of openpyxl in the codebase.

We already have pandas as a requirement, so let's use that.

This also probably speeds up the code slightly.


StevenMaude:
Originally, the reason for this was that I thought it this would remove `openpyxl` completely. But it doesn't because `pandas` depends on it for `read_excel()` which we use elsewhere.

The speed up drops the import data time by about 10 seconds on my PC (from about 35 seconds to 25 seconds).

Jongmassey (approved):
Nice, does `openpyxl` need removing from requirements, too?

mappings/bnfdmd/import_data.py
```diff
@@ -1,37 +1,36 @@
+import pandas as pd
 from django.db import transaction
-from openpyxl import load_workbook
```
lucyb, line 2: You mentioned that this was the only occurrence of `openpyxl` in the codebase, so can it be removed from the requirements.*.txt now?

StevenMaude: `openpyxl` might be removable as a direct dependency. But it would be a transitive dependency via `pandas`. `pandas` doesn't have its own code to read Excel files, but wraps various libraries.

So, we could remove `openpyxl` from the requirements and specify the extra requirement via `pandas`:

```
pandas[excel]
```

which would then specify more clearly why we're pulling in `openpyxl`.

However, `pandas[excel]` actually pulls in *all* the [possible Python Excel libraries that pandas uses](https://pandas.pydata.org/docs/getting_started/install.html#excel-files) which maybe isn't what we want — there's an extra four direct dependencies(!) because `pandas` can use one of several Excel libraries.

And, maybe more importantly: looking again at this dependencies question more closely, this particular script is run in production. But we currently only install `pandas` as a dev requirement, which would mean adding `pandas` as a production requirement too.

So in the end, maybe it's better to close and leave this unmerged, unless the code simplification outweighs adding production dependencies.

lucyb: Oh, that's a really good piece of detective work. Thank you. I'd missed that this would move pandas from a dev requirement to a production one too. 

> maybe it's better to close this and leave this unmerged

Yes, I think that's the right decision in this case. Pandas is not a small dependency and I don't think the code simplification is worth the extra dependency (especially for those of us who don't know Pandas!).

StevenMaude: That sounds good to me too; let's close it.

PR #1993, opened 2024-06-21
Add codelist metadata to API response

### LINKED ISSUE #1991: Add Codelist Metadata to API response
See Slack  🧵 [thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1718704121335439).

We've been asked if we could allow HDRUK to retrieve information about individual codelists, including metadata such as description and references. **If this is to be useful to them if needs to be available in the "near future", which I've taken to mean by mid-end July**.

The current api/v1/codelist endpoint returns this information, but does not include the description and references. This metadata is stored on a codelist rather than codelist version so it _should_ be fairly straightforward to include this as well.
Note: This endpoint is currently used by OSI, which does not require this information, but also should be able to ignore it.

## Acceptance criteria
1. Ability to retrieve codelist information, including metadata such as description and references via an API endpoint.
2. OSI is still able to retrieve codelist information as expected (test locally, if required).
3. Let Seb know this work has been done, so that he can inform the developer who wants to have this information.

iaindillingham:
Add codelist metadata to API response

This adds description, methodology, and references properties (codelist metadata) to a response, following a request with the appropriate query parameters. The following example takes roughly 4s to execute locally.

http://localhost:7000/api/v1/codelist/?description&methodology&references

Closes #1991

Acceptance criteria:

- [x] Ability to retrieve codelist information, including metadata such as description and references, via an API endpoint.
- [x] OSI is still able to retrieve codelist information as expected. (Tested locally.)
- [ ] Let Seb know this work has been done, so that he can inform the developer who wants to have this information.

Closes #1991

### DIFF
--- a/codelists/api.py
+++ b/codelists/api.py
@@ -144,6 +144,22 @@ def codelists_get(request, owner=None):
                 }
             )
 
+        if "description" in request.GET:
+            record["description"] = cl.description
+
+        if "methodology" in request.GET:
+            record["methodology"] = cl.methodology
+
+        if "references" in request.GET:
+            record["references"] = []
+            for reference in cl.references.all():
+                record["references"].append(
+                    {
+                        "text": reference.text,
+                        "url": reference.url,
+                    }
+                )
+
         records.append(record)
     return JsonResponse({"codelists": records})
 
--- a/codelists/tests/test_api.py
+++ b/codelists/tests/test_api.py
@@ -10,7 +10,9 @@ from opencodelists.tests.assertions import assert_difference, assert_no_differen
 
 
 def test_codelists_get(client, organisation):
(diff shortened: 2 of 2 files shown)


Jongmassey (approved):
delightfully simple

rebkwok (approved):
It wasn't specifically requested, but i think it'd be worth adding methodology as well.

Is there any value to only providing these extra metadata with a query param (`?include-metadata` or whatever) so that requests that only care about the basic codelist info don't get a dump of all the descriptions etc as well (some of them might be quite lengthy)?

iaindillingham:
Thank you for your comment, @rebkwok. I've added methodology to the response, and controlled each with a separate query parameter. How does the PR look to you, now?

rebkwok (approved):

rebkwok:
LGTM :)

PR #2071, opened 2024-10-01
Spike dmd codelist builder

### LINKED ISSUE #176: Medicines definition builder
In the longer term we need to think about building a medicines  definition/codelist builder in the same way that we build for clinical terms. Briefly out current methodology is 
- we choose BNF codes terms to be in a codelist
- dm+d is the common language of EHRs in english primary care and core of OpenSAFELY meds software
- we utilise OpenPrescribing google BQ to match this to dm+d and generate a codelist via a SQL query

Benefits building a definition builder would
a) allow anyone to build a medicines codelist (i.e. don't need access to BQ or have SQL skills)
b) have a clearer inclusion/exclusion interface
Not a benefit but we should give some thought how it might fit in with OpenPrescribing infrastructure like analyse page, dm+d browser etc and any potential future tools (e.g. hospital meds data).

In order to build we need a hierachy that makes sense to humans and computers! Both of the immediately obvious options have pros and cons.

| Meds coding system | Pros | Cons | 
| ----------------------- |------|-------|
[BNF codes](https://ebmdatalab.net/prescribing-data-bnf-codes) | Well defined categories that are clinically relevant, machine and human readable, a med can only appear in a single place,  | not dm+d our common meds language, comprehensive for primary care meds but not for meds in secondary care |
[NHS Dictionary of medicines and devices](https://ebmdatalab.net/what-is-the-dmd-the-nhs-dictionary-of-medicines-and-devices/) | Common language of opensafely EHRs, meds appear in multiple places | It does not contain a clinical hierarchy |

 I think ideally we would have both options available in a search function recognising limitations of both. There may be other hierarchies that could be useful. Can anyone suggest some?

Jongmassey:
Fixes #176 

In-person hack with @evansd and @inglesp to provide ingredient, VMP, and AMP name based building of dm+d codelists

### DIFF
--- a/coding_systems/dmd/coding_system.py
+++ b/coding_systems/dmd/coding_system.py
@@ -1,5 +1,7 @@
+from django.db.models import Q
+
 from ..base.coding_system_base import BaseCodingSystem
-from .models import AMP, AMPP, VMP, VMPP, VTM
+from .models import AMP, AMPP, VMP, VMPP, VTM, Ing
 
 
 class CodingSystem(BaseCodingSystem):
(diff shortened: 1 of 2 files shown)


Jongmassey:
Having done a bit more digging, the NHSBSA dm+d supplementary file (aka dmdbonus) has VTM to ingredient mapping. This presents us with an opportunity to present the search results by VTM hierarchy more clearly.

This would require the dmd supplementary data to be added to the dm+d coding system.

For example, [the descendent relationships](https://github.com/opensafely-core/opencodelists/blob/c943cf692576ac87725651dfd38527d12f619684/coding_systems/dmd/coding_system.py#L26-L44)  defined here, go from ing-to-vmp, ing-to-amp, vmp-to-amp. This could be simplified to (assuming vtm ingredient as `VTI`)

```python

vtms = VTM.objects.using(self.database_alias).filter(
   Q(vti__ing_id__in=codes) | Q(id__in=codes)
)
vmps = VMP.objects.using(self.database_alias).filter(
   Q(vtm__in=vtms_from_ings) | Q(id__in=codes)
  )
amps = AMP.objects.using(self.database_alias).filter(
    Q(vmp_id__in=[vmp.id for vmp in vmps_from_ings]) | Q(vmp_id__in=codes)
  )

rv = (
   {
      (vti.ing_id, vtm.id)
      for vtm in vtms 
      for vti in vtm.vti_set.all()
    }
    | {(vmp.vtm_id, vmp.id) for vmp in vmps}
    | {(amp.vmp_id, amp.id) for amp in amps}
  )
return rv
```

Jongmassey:
to pick up in a future initiative

mikerkelly:
Related to [#2388](https://github.com/opensafely-core/opencodelists/issues/2388).

PR #2088, opened 2024-10-17
Document how to restore from backup

iaindillingham:
In OpenCodelists, backups are JSON fixtures rather than SQL dumps. To restore from a backup, we must remember to remove the existing database and migrate, and *then* to load the data from the fixture.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -47,7 +47,19 @@ A place to put scripts to be run via [runscript](https://django-extensions.readt
 ## Production database and backups
 
 The production database and backups are located at  `/var/lib/dokku/data/storage/opencodelists` on dokku3 (see also [deployment notes](DEPLOY.md)).
+This database is the core (default) database;
+the coding system databases are located within the `coding_systems` subdirectory.
 
+The backups are created with the dumpdata management command (`deploy/bin/backup.sh`).
+They can be restored with:
+
+```sh
+rm db.sqlite3
+
+python manage.py migrate
+
+python manage.py loaddata core-data-<date>.json
+```
 
 ## Local development
 


DEVELOPERS.md
```diff
@@ -48,6 +48,16 @@ A place to put scripts to be run via [runscript](https://django-extensions.readt
 
 The production database and backups are located at  `/var/lib/dokku/data/storage/opencodelists` on dokku3 (see also [deployment notes](DEPLOY.md)).
 
+The backups are created with the dumpdata management command (`deploy/bin/backup.sh`).
```
rebkwok, line 51: Might be worth noting that these are backups of the core (default) db only, and not the coding system databases (which incidentally, are not backed up at all (but can theoretically be re-created from the coding system data files))

rebkwok (approved):
This looks fine to me as documentation of how to restore a production backup in production. However, if it's used as a method of loading a version of the production db locally, as you did, it's probably worth some notes about how to also get the latest coding system databases (and a caveat that codelist versions use a specific coding system release, which corresponds to a database, and they won't display properly if they depend on a coding system database that you don't have locally)

iaindillingham:
> However, if it's used as a method of loading a version of the production db locally, as you did, it's probably worth some notes about how to also get the latest coding system databases (and a caveat that codelist versions use a specific coding system release, which corresponds to a database, and they won't display properly if they depend on a coding system database that you don't have locally)

Could I check my understanding, @rebkwok? I think this means that to restore the production database locally, I'd need to download all (or nearly all) the coding system databases. Is that correct? Would me not having done so explain #2091?

And if that _is_ correct, then would that mean downloading every (or nearly every) `*.sqlite3` file within `/var/lib/dokku/data/storage/opencodelists/coding_systems/`? (In other words, there aren't fixtures for these databases.)

DEVELOPERS.md
iaindillingham: Noted! Thank you 🙂

iaindillingham:
> This looks fine to me as documentation of how to restore a production backup in production. However...

Thanks, Becky. I'm going to side-step this, for now. REX are starting an OpenCodelists initiative, and I anticipate that we'll spend some time figuring out how best to get a local development environment set up.

PR #2100, opened 2024-10-22
Automate downloading of bnf dmd mappings

### LINKED ISSUE #2099: Improve process for updating BNF SNOMED mapping
The NHS [publishes](https://www.nhsbsa.nhs.uk/prescription-data/understanding-our-data/bnf-snomed-mapping) BNF SNOMED mapping files monthly. The updates are usually published towards the end of the month.

The date of the mapping file given on the website is different from the datestamp in the filename. This is because the filename datestamp is the date which the mapping file was generated, and the date published on the website is the date to which the mappings are up to date. There is usually a 3 month lag.

For example, at the time of writing the most recent update file is the "July 2024" update, but the mapping file is called `BNF Snomed Mapping data 20240923.xlsx`

The current [mapping import process](https://github.com/opensafely-core/opencodelists/blob/main/opencodelists/management/commands/import_mapping_data.py) takes no record of the filename used in the import, and thus it is not possible to ascertain how up to date the BNF-SNOMED mapping on OpenCodelists is.

There has been a recent tech support issue querying why mappings were missing and I have memories of missed mapping updates being a problem in the past.

We should improve this such that:

* Updates to this mapping file are done in a timely fashion
* It is easy to check that the updates have been done

We should consider:
* Surfacing the mapping file update date to users
* Automating the getting of these updates as per we do for other coding systems - [Draft](https://github.com/opensafely-core/opencodelists/pull/2100)
* ~~Setting a Bennett Bot reminder to apply these updates~~ ✅

Jongmassey:
This PR largely makes use of pre-existing mechanisms to automatically download and load the BNF to dm+d mappings from the NHSBSA website.

The structure of the downloader largely mirrors that of those for the dm+d and SNOMED CT downloaders, as does the means of calling it and the scheduling of this call.

The web scraping code is mostly taken from OpenPrescribing with a couple of additions to parse the "valid from" date and to conform to the above described downloader pattern.

Some of this functionality (parsing of extra metadata and the ability to download multiple/specific releases of the mapping other than the latest) is not yet used, and so arguably could be removed until such time as we start showing the mapping metadata to users in the frontend and/or storing multiple versions of the mapping as per the coding systems.

Fixes #2099

### DIFF
--- a/mappings/bnfdmd/data_downloader.py
+++ b/mappings/bnfdmd/data_downloader.py
@@ -1,0 +1,130 @@
+import glob
+import os
+import re
+from datetime import datetime
+from pathlib import Path
+from urllib.parse import unquote, urljoin, urlparse
+
+import requests
+import structlog
+from bs4 import BeautifulSoup
+
+
+logger = structlog.get_logger()
+
+
+# modelled after coding_systems/base/trud_utils.py::TrudDownloader @ 979d995
+
+
+class Downloader:
+    def __init__(self, release_dir):
+        self.url = "https://www.nhsbsa.nhs.uk/prescription-data/understanding-our-data/bnf-snomed-mapping"
+        self.release_dir = release_dir
+
+    def get_releases(self):
+        # adapted from OpenPrescribing:
+        # https://github.com/ebmdatalab/openprescribing/blob/main/openprescribing/dmd/management/commands/fetch_bnf_snomed_mapping.py
+        filename_re = re.compile(
+            r"^BNF Snomed Mapping data (?P<date>20\d{6})\.zip$", re.IGNORECASE
+        )
+        datepat = re.compile(r"\w+ \d{4}")
+        datefmt = "%B %Y"
+
+        rsp = requests.get(self.url)
+        rsp.raise_for_status()
+        doc = BeautifulSoup(rsp.text, "html.parser")
+
+        matches = []
+        for a_tag in doc.find_all("a", href=True):
+            if a_tag.text.startswith("\xa0"):
+                continue
+            url = urljoin(self.url, a_tag["href"])
+            filename = Path(unquote(urlparse(url).path)).name
+            match = filename_re.match(filename)
+            if match:
+                inner_text = a_tag.text
+                while not (
+                    valid_from := datepat.search(inner_text.replace("\xa0", " "))
+                ):
+                    # sometimes "valid from" inner text is across adjacent elements
+                    inner_text += a_tag.next_sibling()[0].text
+                valid_from = datetime.strptime(valid_from.group(), datefmt)
+                matches.append((match.group("date"), url, filename, valid_from))
+
+        if not matches:
+            raise RuntimeError(f"Found no URLs matching {filename_re} at {self.url}")
+
+        # Sort by release date and get the latest
+        matches.sort()
+
+        return matches
+
+    def get_latest_release(self):
+        return self.get_releases()[-1]
+
+    def get_release_metadata(self, release):
+        datestamp, url, filename, valid_from = release
+
+        return {
+            "release": datestamp,
+            "valid_from": valid_from,
+            "url": url,
+            "filename": filename,
+            "release_name": datestamp,
+        }
+
+    def get_latest_release_metadata(self):
+        latest = self.get_latest_release()
+        return self.get_release_metadata(latest)
+
+    def download_release(self, release_name, valid_from):
+        """
+        Download a release that matches the specified release_name and valid_from values
+        """
+        logger.info(
+            "Attempting to download release from NHS BSA",
+            release=release_name,
+            valid_from=valid_from,
+        )
+
+        releases = self.get_releases()
+        release_metadata = (self.get_release_metadata(release) for release in releases)
+
+        match_found = False
+        for metadata in release_metadata:
+            if (
+                metadata.get("release_name") == release_name
+                and metadata.get("valid_from") == valid_from
+            ):
+                match_found = True
+                break
+
+        if not match_found:
+            raise ValueError(
+                f"No matching release found for release {release_name}, valid from {valid_from}"
+            )
+
+        local_download_filepath = Path(self.release_dir) / metadata["filename"]
+        self.get_file(metadata["url"], local_download_filepath)
+
+        return local_download_filepath
+
+    def get_file(self, url, filepath):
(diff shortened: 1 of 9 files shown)


rebkwok:
Just in case you weren't aware already - OP also does bnf-dmd downloading of the same file; it'd be nice if we could make sure they're always in line
https://github.com/ebmdatalab/openprescribing/blob/main/openprescribing/dmd/management/commands/fetch_bnf_snomed_mapping.py

Jongmassey:
Ah cool, similar end by slightly different means. I'll bump this back into triage for team rex to see if we can get it merged

mappings/bnfdmd/data_downloader.py
```diff
@@ -0,0 +1,129 @@
...
+        return local_download_filepath
+
+    def get_file(self, url, filepath):
+        logger.info("Starting download", download_filepath=filepath)
+        with requests.get(url, stream=True) as resp:
+            resp.raise_for_status()
+            with open(filepath, "wb") as f:
+                for chunk in resp.iter_content(chunk_size=8192):
+                    f.write(chunk)
+        logger.info("Download complete", download_filepath=filepath)
```
StevenMaude, line 118: thought: It's not a big problem, but this code *might* be a candidate for factoring out both places it appears.

The code, neither here, nor in `TrudDownloader` never uses the `self` reference, so it could be a class static method, or a shared utility function.

That said, it was previously moved from being a function into the class: https://github.com/opensafely-core/opencodelists/commit/979d995aee8e377a47257323340f3245adae29b0, and it also brings up the question of "where does this live?".

**So having said all that, I'm not sure we need to do much about it, especially as some of the other code is duplicated too.**

But we could make it, and the other duplicate, static methods though, to make it simpler for the future reader.

mappings/bnfdmd/data_downloader.py
```diff
@@ -0,0 +1,129 @@
+import glob
+import os
+import re
+from datetime import datetime
+from pathlib import Path
+from urllib.parse import unquote, urljoin, urlparse
+
+import requests
+import structlog
+from bs4 import BeautifulSoup
```
StevenMaude, line 10: thought: Having just been looking at lots of dependencies, BeautifulSoup introduces an additional two. It's a *little* bit of a smell, because we already have lxml as a dependency, which we could use instead to do the HTML parsing.

I think it is OK for now, because this is how the OpenPrescribing code works (but it could also be refactored there in future).

I'd consider opening a deck scrubbing issue to revisit switching to lxml.

requirements.prod.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with Python 3.12
 # by the following command:
 #
-#    pip-compile --allow-unsafe --generate-hashes --output-file=requirements.prod.txt requirements.prod.in
+#    pip-compile --generate-hashes --output-file=requirements.prod.txt requirements.prod.in
```
StevenMaude, line 5: nit: This has changed, which changes the setuptools-related output in this file.

mappings/bnfdmd/data_downloader.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+logger = structlog.get_logger()
+
+
+# modelled after coding_systems/base/trud_utils.py::TrudDownloader
```
StevenMaude, line 16: nit: I'd consider referencing the commit ID, to make it easier for someone to piece together if the other code changes.

mappings/bnfdmd/data_downloader.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class Downloader:
+    def __init__(self, release_dir):
+        self.url = "https://www.nhsbsa.nhs.uk/prescription-data/understanding-our-data/bnf-snomed-mapping"
+        self.release_dir = release_dir
+
+    def get_releases(self):
+        # adapted from openprescribing/openprescribing/dmd/management/commands/fetch_bnf_snomed_mapping.py
```
StevenMaude, line 25: nit: As above, I'd consider providing a GitHub link to the file in question for the specific commit you borrowed from.

mappings/bnfdmd/data_downloader.py
```diff
@@ -0,0 +1,129 @@
...
+            with open(filepath, "wb") as f:
+                for chunk in resp.iter_content(chunk_size=8192):
+                    f.write(chunk)
+        logger.info("Download complete", download_filepath=filepath)
+
+    def download_latest_release(self):
+        metadata = self.get_latest_release_metadata()
+
+        if glob.glob(os.path.join(self.release_dir, metadata["filename"])):
+            return
```
StevenMaude, line 124: question: Might it be possible that this gets in a broken state by, say:

* the file downloaded is broken (either the request is not completed for whatever reason, or the server gives an empty output)
* the file persists on disk
* and then further attempts to update for that release fail because the broken file is still there?

Fixing it would require someone to intervene, or waiting for the next release filename.

(For comparison, I am not completely sure how atomic the coding system imports are, and therefore how robust they are.)

StevenMaude (approved):
summary: Overall, this is fine. I've added a few thoughts and suggestions.

Thank you for explaining the provenance of the original code, as it made it easier to consider what was particular to the specific case being handled here :sunglasses:

mappings/bnfdmd/data_downloader.py
Jongmassey: My first draft of this used lxml, but I switched to the prior art implementation from OpenPrescribing as it has seen a lot of use in production and thus can be relied on

requirements.prod.txt
Jongmassey: huh, not sure why or how this changed, I'll revert it

mappings/bnfdmd/data_downloader.py
Jongmassey: this was taken from the OP management command.

I think "manual intervention if it breaks" is fine in this case (deleting the broken download).

The alternative is to have the file be re-imported every time this command runs, regardless of whether there's a new update or not.

Since we don't (yet) record the version of the mapping file that's been loaded the only information we have as to the version is the filename itself.

Something to revisit if/when we start versioning the mappings

mappings/bnfdmd/data_downloader.py
Jongmassey: I'll raise a deck-scrubbing issue to refactor this and the TrudDownloader

PR #2120, opened 2024-11-12
Add updated date to codelist version template

### LINKED ISSUE #2106: Show codelist creation date
From [this Slack thread](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1727955808431179):

> It would be very useful if we could see the date a codelist was created. There is the sign off date, but that only applies to the original codelist, i'm more interested in the dates of different versions, to know when it was last updated.

Could be on either or both of the landing page index and in the codelist version view. Landing page index could allow sorting or filtering by dates?

We have [`CodelistVersion.created_at`](https://github.com/opensafely-core/opencodelists/blob/cd038a6670ba17ff68cfd283a669ad245b46d004/codelists/models.py#L221). Versions also have `tag` and `id`.

Some version identifiers seem to include a date, but not all. E.g., [this](https://www.opencodelists.org/codelist/opensafely/ace-inhibitor-medications/2020-05-19/) vs. [that](https://www.opencodelists.org/codelist/openprescribing/ace-inhibitors/2bb0d404/).

Jongmassey:
Fixes #2106 

The codelist versions are rendered in date creation order, but this is not always obvious to users.
Even then, we do not surface the creation of modification date anywhere.

For a user, trying to find a version of a codelist at a particular point in time, this change provides the neccesary information.

image

### DIFF
--- a/templates/codelists/version.html
+++ b/templates/codelists/version.html
@@ -197,24 +197,43 @@
 
     <hr />
 
-    <h6>Versions</h6>
-    <ul class="pl-3">
-      {% for version in versions %}
-      <li>
-        {% if version == clv %}
-        {{ version.tag_or_hash }}
-        {% else %}
-        <a href="{{ version.get_absolute_url }}">{{ version.tag_or_hash }}</a>
-        {% endif %}
-        {% if version.is_under_review %}
-        <span class="badge badge-primary">Review</span>
-        {% elif version.is_draft %}
-        <span class="badge badge-primary">Draft</span>
-        {% endif %}
-        <span>(updated {{ version.updated_at | date:'Y-m-d H:i:s' }})</span>
-      </li>
-      {% endfor %}
-    </ul>
+    <div class="card">
+      <div class="card-header px-2 py-3">
+        <h2 class="h5 mb-0">Versions</h2>
+      </div>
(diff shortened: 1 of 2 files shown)


templates/codelists/version.html
```diff
@@ -211,6 +211,7 @@ <h6>Versions</h6>
         {% elif version.is_draft %}
         <span class="badge badge-primary">Draft</span>
         {% endif %}
+        <span>(updated {{ version.updated_at | date:'Y-m-d' }})
```
StevenMaude, line 214: Unless I'm missing something, maybe this needs the closing tag?

```suggestion
        <span>(updated {{ version.updated_at | date:'Y-m-d' }})</span>
```

StevenMaude (approved):
This mostly looks fine.

But in the example screenshot, both codelists have the same date; I can't obviously tell which version is newer. (I could have if they had different dates, granted.)

Is it worth including the update time too?

templates/codelists/version.html
Jongmassey: Oh wow, don't know how I missed that one!

Jongmassey:
Here's it with the time:

image


I was trying to keep things as small and un-ugly as possible, but as you said there are genuine reasons to show the time so I think it's unavailable.

PR #2124, opened 2024-11-12
Avoid suggesting `rm` when working with a database

StevenMaude:
It could be the production database, and copy-pasting that series of commands could be undesirable.


mikerkelly (approved):
Good idea.

I think this is also useful practice for local development.

I suppose we hope that the DB files are still backed up somewhere but I think we're not even sure of that at the moment, and it would be inconvenient to need to resort to that in any case.

PR #2126, opened 2024-11-13
Produce representative subset of data for development fixtures

### LINKED ISSUE #2117: SPIKE populate local core DB with dummy (or anon) data

Jongmassey:
fixes #2117


Jongmassey:
Not for now, could revisit if we find other approaches inadequate

PR #2129, opened 2024-11-13
Move healthcheck to `app.json`

### LINKED ISSUE #1868: Dokku CHECKS file is deprecated
The same should be done to this repository as opensafely-core/job-server#4152.

EDIT: See https://github.com/opensafely-core/job-server/issues/4489

StevenMaude:
Fixes #1868.

### DIFF
--- a/CHECKS
+++ b/CHECKS
@@ -1,6 +1,0 @@
-# Wait for 30 seconds to allow time for migrations, if necessary
-WAIT=30
-TIMEOUT=60
-ATTEMPTS=10
-
-http://localhost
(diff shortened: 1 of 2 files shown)


StevenMaude:
I could be wrong, but I think from memory: there's possibly a quirk with how Dokku applies this configuration that it may only actually take effect in the subsequent deploy after the one where this is merged.

iaindillingham (approved):
Thanks for this, @StevenMaude. Rather than setting up Dokku in a Codespace, I'm going to merge and release, then merge and release a separate (zero change) PR to check that the health check was successful.

PR #2140, opened 2024-11-15
Set up Vite to reload pages on code changes

tomodwyer:
Add `vite-plugin-live-reload` to simplify reloading when changing styles.

This only takes effect when `npm run dev` is being used.


Jongmassey:
>This only takes effect when npm run dev is being used.

Could you explain how this works?

Jongmassey (approved):

PR #2152, opened 2024-11-19
Rework Docker image compression in workflows

StevenMaude:
As opensafely-core/job-server#4392.

* Use zstd instead of gzip
* Skip compression when uploading the image as an artifact


StevenMaude:
> Skip compression when uploading the image as an artifact

For clarity: this is because the image is already compressed by us, and then the `upload-artifact` action also wants to compress the image, which takes more time but doesn't gain too much.

Jongmassey (approved):
👍

PR #2130, opened 2024-11-13
Upgrade django to 4.2.16

StevenMaude:
Fixes #2107.

This is the latest 4.2 LTS release.

The changes since 4.2.4 are small in scope; mostly regression and security fixes.

PR #2155, opened 2024-11-20
Upgrade to Python 3.13

StevenMaude:


StevenMaude:
Questions before merging:

* Do we want to let the Python 3.12 upgrade settle sometime, before deploying this?
* Are there any issues with any developers getting Python 3.13 on their local installs? Python 3.13 was only released a few weeks ago.

StevenMaude:
Have done the same manual testing [as in this comment](https://github.com/opensafely-core/opencodelists/pull/2130#issuecomment-2476300602).

PR #2164, opened 2024-11-27
Remove use of `subprocess` in test

mikerkelly:
Using `call_command` is better than `subprocess` because it’s simpler, faster, more readable, and doesn’t rely on `python` being available. In most cases, test cases should not use `subprocess` unless testing integration with external tools or the shell.  This change aligns the test with the rest of the module and codebase.

Also made some minor readability tweaks.

### DIFF
--- a/opencodelists/tests/test_import_coding_system_data.py
+++ b/opencodelists/tests/test_import_coding_system_data.py
@@ -1,4 +1,3 @@
-import subprocess
 from datetime import date
 from unittest.mock import patch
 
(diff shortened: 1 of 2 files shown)


opencodelists/tests/test_import_coding_system_data.py
```diff
@@ -90,16 +90,16 @@ def test_calls_import_data_function_coding_system_force_overwrite(
 
 
 @pytest.mark.parametrize(
-    "input_date,error",
+    "input_date,error_regex",
```
iaindillingham, line 93: This is a really nice improvement. Knowing that the string is a regex immediately tells me why the parentheses are escaped.

iaindillingham (approved):
Thank you for writing really clear commit messages. Knowing why you chose to make these changes was really helpful.

PR #2185, opened 2024-12-02
Add custom endpoint for dokku healthchecks

### LINKED ISSUE #2154: Add health check endpoint
At present, a single user-defined health check runs after a new container is started, but before it receives connections. Essentially, the Dokku host `curls` the new container, requesting the root URL (see #2148 and #2156).

https://github.com/opensafely-core/opencodelists/blob/b3da21a5125a19b968d99f17da7fe3d0da8667e3/app.json#L9-L26

A 200 response from the root URL isn't necessarily a good indicator that the new container is healthy. The root URL is also slow to respond (see #1857). job-server has a `HealthCheck` view, which returns a 200 response (`jobserver/views/health_check.py`, <https://jobs.opensafely.org/health-check/>). OpenCodelists should have the same; it should be used by Freshping.

KatieB5:
When we deply OpenCodelists, Dokku starts each container and performs a health check to ensure the container has come up properly before routing traffic to it. These health checks are customisable (see [docs](https://dokku.com/docs/deployment/zero-downtime-deploys/)). Dokku only uses this health check when deploying new containers, it does not use it once the container has been deployed (we do have Freshping checking the site every few minutes though to make sure it's still working okay).

This change moves away from using the homepage as the health check endpoint and to a specific endpoint that will return 200 when the app starts and nothing else. Using the homepage as a health check is slower and makes a number of database connections, whereas this in theory decouples the app health check from the database.

Closes #2154

See https://github.com/opensafely-core/job-server/commit/717946f18

### DIFF
--- a/app.json
+++ b/app.json
@@ -18,7 +18,7 @@
                 ],
                 "initialDelay": 30,
                 "name": "Web health check",
-                "path": "/",
+                "path": "/health-check",
                 "port": 7000,
                 "timeout": 60,
                 "type": "startup"
--- a/opencodelists/tests/views/test_health_check.py
+++ b/opencodelists/tests/views/test_health_check.py
@@ -1,0 +1,3 @@
+def test_health_check_successful(client):
+    response = client.get("/health-check/")
+    assert response.status_code == 200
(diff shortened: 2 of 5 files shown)


iaindillingham (approved):
Excellent work! It's nice to see that you've used a function based view, for consistency with the other views.

PR #2186, opened 2024-12-02
Report rows imported in import mapping command

mikerkelly:
Provide some feedback on how many rows are imported. This is useful as a cross-check against the number of mappings in the source file(s).

Also tidy up the simplicity, readability and documentation of this command.

First commit ("List available mappings in mappings.mappings.") has some accidental changes to `import_mapping_data.py` -- please ignore, this was WIP simplification of the `possible_modules` function I realized could go away altogether. Beware `git add -u`...

### DIFF
--- a/opencodelists/management/commands/import_mapping_data.py
+++ b/opencodelists/management/commands/import_mapping_data.py
@@ -12,8 +12,8 @@ class Command(BaseCommand):
         parser.add_argument(
             "source",
             help=(
-                "File or dir that defines the mapping, see import_data.py in "
-                "the mapping's package to understand how it is used"
+                "File or dir that defines the mapping, see README and/or "
+                "import_data.py in the mapping's package for more info"
             ),
         )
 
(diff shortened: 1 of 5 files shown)


opencodelists/management/commands/import_mapping_data.py
```diff
@@ -7,10 +7,17 @@
...
 class Command(BaseCommand):
     def add_arguments(self, parser):
-        parser.add_argument("dataset", choices=mappings.keys())
-        parser.add_argument("release_dir")
+        parser.description = "Import a coding system mapping"
+        parser.add_argument("mapping", choices=mappings.keys(), help="Mapping name")
+        parser.add_argument(
+            "source",
+            help=(
+                "File or dir that defines the mapping, see import_data.py in "
```
Jongmassey, line 15: nit: maybe the per-mapping README is a friendlier guide than going to the source?

Jongmassey (approved):
This is an excellent improvement, fixing things that have annoyed me for some time! One minor nit

opencodelists/management/commands/import_mapping_data.py
mikerkelly: Good point, thanks. I'll mention both I think, in case there is no README or it's inadequate.

PR #2188, opened 2024-12-02
Add custom error page templates

### LINKED ISSUE #2168: Custom 500 and 404 error pages
If a user receives an error or requests a resource that can't be found, they receive the default error pages, e.g a white page with the text (in the case of a 404):

```
Not Found
The requested resource was not found on this server.
```

We should consider doing the same kind of thing that we do on Job Server, with a link to the home page and a link to contact support (see [404](https://jobs.opensafely.org/404) and [500](https://jobs.opensafely.org/500) pages).

tomodwyer:
Closes #2168

- Adds custom error pages
- Link to the home page and to contact support

### DIFF
--- a/opencodelists/urls.py
+++ b/opencodelists/urls.py
@@ -4,6 +4,7 @@ from django.urls import include, path
 from django.views.generic import RedirectView
 
 from . import views
+from .views.errors import bad_request, page_not_found, permission_denied, server_error
 
 
 users_patterns = [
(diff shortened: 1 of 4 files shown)


opencodelists/urls.py
```diff
@@ -4,6 +4,7 @@
 from django.views.generic import RedirectView
 
 from . import views
+from .views.errors import bad_request, page_not_found, permission_denied, server_error
```
mikerkelly, line 9: To cause the error pages to be used when the errors are triggered:

```suggestion
from django.conf.urls import handler404, handler500, handler403, handler400
handler400 = bad_request
handler403 = permission_denied
handler404 = page_not_found
handler500 = server_error
```
Note that Django debug toolbar will still take precedence if installed so set DEBUG to False if you want to provoke them to see them work locally.

mikerkelly:
```python
from .views.errors import csrf_failure
CSRF_FAILURE_VIEW = csrf_failure
```
In opencodelists/settings.py will cause the CSRF handler to take effect. Then I think this is good to go.

I think this fixes #2168 and should be linked as such.

mikerkelly: Templates and views all look good.

mikerkelly (approved):

PR #2193, opened 2024-12-03
Allow content to fill more of the viewport width

tomodwyer:
## Commits

- **Create full_width_content block**
  - This allows the normal content to fit to the Bootstrap standard container, whilst allowing wider content to extend to a maximum width of `1536px` (as per Tailwind CSS we use elsewhere)
- **Add break-word CSS utility**
  - Allow URLs and long strings without spaces to break over lines in the sidebar (see diff page images below)
- **Improve diff view layout**
  - Add a meta title
  - Use the wide container
  - Allow the left column to use more space at large screen widths
- **Update the codelist builder layout to full width**
  - Add a meta title
  - Use the wide container
  - Allow the left column to use more space at large screen widths
  - Allow right column to overflow in the X axis, rather than overflowing the page
  - Add break-word utility class
- **Update codelist version to full width**
  - Use the wide container
  - Allow the left column to use more space at large screen widths
  - Add break-word utility class
- **Use djhtml to format templates**
  - This is so that the templates use the same spacing for future diffs
  - We can add this to the `format` and `check` steps once the majority of templates have been updated
- **Fix invalid list item nesting in diff view**
  - List items were not correctly indented properly causing lots and lots of empty `<ul>`s to be generated
    - This ends up removing about 30% of the HTML nodes on the page
- **General**
  - Remove `<hr>` tags where possible
  - Fix whitespace issues

### DIFF
--- a/assets/src/scripts/components/CodelistBuilder.jsx
+++ b/assets/src/scripts/components/CodelistBuilder.jsx
@@ -151,7 +151,7 @@ class CodelistBuilder extends React.Component {
     return (
       <>
         <div className="row">
-          <div className="col-md-3 col-lg-2">
+          <div className="col-md-3">
             {this.props.isEditable && (
               <>
                 <this.ManagementForm complete={this.complete()} />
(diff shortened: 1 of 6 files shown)


Jongmassey (approved):
Ruddy marvellous

PR #2197, opened 2024-12-05
Apply password validation in all environments

### LINKED ISSUE #2172: Apply password validation for new accounts in all environments
Currently password validation is only enforced in production environments:

https://github.com/opensafely-core/opencodelists/blob/9fb2cf8bbbf3cb9cef533f09242d2cb9d80b1c63/opencodelists/settings.py#L177-L185

This caught me out when preparing https://github.com/opensafely-core/opencodelists/pull/2171 as it wasn't showing the password validation fields.

Unless there is a good reason I'm not considering, we should remove this if statement and apply the `AUTH_PASSWORD_VALIDATORS` at all times.

KatieB5:
Closes #2172

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -177,15 +177,15 @@ AUTH_USER_MODEL = "opencodelists.User"
 # Password validation
 # https://docs.djangoproject.com/en/3.0/ref/settings/#auth-password-validators
 
-if IN_PRODUCTION:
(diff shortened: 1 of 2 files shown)


iaindillingham (approved):
Well done! 🚀 I'm sorry we took so many wrong turns, but hopefully we learnt more about Django's settings and monkeypatching.

PR #2213, opened 2024-12-09
Remove Python distutils

tomodwyer:
Duplicate of https://github.com/opensafely-core/reports/pull/820

The distutils module has been deprecated for some time and was removed in Python3.12[^1]. A recent update to deadsnakes means that the package is no longer available at all and is breaking the build. Removing the package doesn't appear to break anything and it's likely that everything has already been moved over to setup tools.

[^1]: https://peps.python.org/pep-0632/#migration-advice


Jongmassey (approved):
Seems to work. Might be worth letting the channel know they'll have to rebuild their venvs after this 👍

Issue #2151, opened 2024-11-19
Use sqlite cli to backup core database


lucyb:
We recently noticed a problem with restoring the existing backup of the core database (#2091).

Rather than try to fix it, this is a proposal to replace the current backup with a sqlite CLI functionality instead.

We should also strongly consider making the backups into a different directory to the production sqlite database, for ease of management and to reduce potential for any errors.

See [slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1731940750156009) 🧵. 

This [blog post](https://oldmoe.blog/2024/04/30/backup-strategies-for-sqlite-in-production/) evaluates the options available for backing up a SQLite db.

## `.dump` vs `.backup`

`.dump` was the original suggestion in this ticket. It's designed for data migration and similar to `dumpdata`, producing text output of SQL commands required to reconstruct the DB schema and contents. `.backup` is an alternative SQLite CLI command designed for backups that produces binary output.

`.backup` produces smaller, binary output that is less compressable but more-amenable to de-duplication if we need to move to incremental backups to save space (as described in the linked blog post).

`.backup` is effectively a file copy with the database locked and is faster and more scalable to both backup and restore than `.dump`.

`.dump` is portable to other database systems or versions, `.backup` is not.

Either has things to recommend them but I (Mike) have chosen to use `.backup` as backup is its intended use case, it is faster, and more optimizable for space if we need to optimize that. It's also more aligned with the backup approach envisaged in #2153 for the coding system databases.

## Assumptions
We'll be using the official sqlite cli, so it's less likely to have problems than using `manage.py dumpdata` / `loaddata`. It's also likely to use less memory than [the current](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1731357807835719) `mange.py loaddata` command, which may be significant on our relatively small server.

The current approach doesn't backup some sensitive tables (namely the django session tables), to support its use in local development. I don't believe that will be an issue, as we are planning to move away from using production data (unless for a specified purpose) in local development and because in practice people aren't using this data anyway.

## Acceptance criteria

- [x] Update the [current cron job](https://github.com/opensafely-core/opencodelists/blob/main/deploy/bin/backup.sh) to use sqlite cli `.backup` -- #2214.
- [x] Update line 17 to keep only the last 30 days of the new backup files -- #2214.
- [x] Check that a local restore (NOT production) works as expected -- #2214.
- [x] Document the changes in an ADR. -- #2246 
[...]

PR #2214, opened 2024-12-09
Take backups via `sqlite ".backup"`

mikerkelly:
Partially addresses #2151. ADR about the core backup strategy will be in a separate PR.

Taking backups via the sqlite CLI is simple and clean. We can restore from them.

We haven't been able to locally restore from the `dumpdata` fixtures-based backups. For now, keep the old-format backups just in case we want to try again to restore from them. After 30 days of creating new backups we can update the script to not consider the old format and location.

See [these comments](https://github.com/opensafely-core/opencodelists/issues/2151#issuecomment-2517804854) about the execution time and file system space requirements.

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -66,7 +66,7 @@ Check cron tasks:
 dokku$ dokku cron:list opencodelists
 ```
 
-Backups are saved to `/var/lib/dokku/data/storage/opencodelists` on dokku3.
+Backups are saved to `/var/lib/dokku/data/storage/opencodelists/backup` on dokku3.
 
 ### Manually deploying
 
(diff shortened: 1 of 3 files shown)


Jongmassey (approved):
This looks great, have you considered using `zstd` instead of `gz` here? It's installed on dokku3 and is significantly faster for the same file sizes


```sh

time gzip -f db.sqlite3

real    2m36.764s
user    2m30.439s
sys     0m3.521s

1.1G Dec 11 12:52 db.sqlite3.gz

time zstd db.sqlite3 -o db.sqlite3.zstd
real    0m29.130s
user    0m24.160s
sys     0m10.851s

1.1G Dec 11 12:52 db.sqlite3.zstd

```

mikerkelly:
#2151 for zstd compression.

PR #2220, opened 2024-12-10
Update transitive dependencies of environs

StevenMaude:


StevenMaude:
As mentioned in [this update of environs](https://github.com/opensafely-core/opencodelists/pull/2162#issuecomment-2521087225), environs is used only in `opencodelists/settings.py`.

I tested this locally: environment variables do get loaded.

evansd:
FWIW, we opted not to use this Airlock specifically because its transitive dependencies seemed pretty heavyweight for the value it gave. Not saying you shouldn't use it, but worth bearing in mind if you're in the process of reconsidering this stuff.
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1723201362723169

PR #2232, opened 2024-12-11
Update most of the requests dependency subtree

StevenMaude:
This updates requests and its transitive dependencies, except for urllib3 which may have issues with the versions of the opentelemetry packages.

We'll update urllib3 when the opentelemetry packages are updated.


StevenMaude:
Our own code's uses of requests are routine.

tomodwyer (approved):

PR #2236, opened 2024-12-12
Update transitive dependencies for pandas

StevenMaude:


StevenMaude:
We're only directly using `read_excel` from `pandas` whose API won't change with these updates.

tomodwyer (approved):

PR #2237, opened 2024-12-12
Update some common transitive dependencies

StevenMaude:


StevenMaude:
These are all widely used Python packages; they should have fairly stable APIs.

tomodwyer (approved):

PR #2241, opened 2024-12-12
Switch from Dependabot to update-dependencies-action

### LINKED ISSUE #2165: Dependabot is not updating all Python dependencies
For example, the OpenTelemetry dependencies are old, yet there are no Dependabot pull requests for these packages.

We may want to switch to the opensafely-core/update-dependencies-action for the Python dependencies.

Dependent on https://github.com/opensafely-core/opencodelists/issues/1817

StevenMaude:
Fixes #2165.

This also pins Django to 4.2 LTS, for now, until we resolve issues in upgrading and test a later release.

### DIFF
--- a/.github/workflows/update-python-dependencies.yml
+++ b/.github/workflows/update-python-dependencies.yml
@@ -1,0 +1,26 @@
+name: Update Python dependencies
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  "12 6 * * *"
+
+jobs:
+  update-dependencies:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: "opensafely-core/setup-action@v1"
+      with:
+        python-version: "3.12"
+        install-just: true
+
+    - uses: actions/create-github-app-token@v1
+      id: generate-token
+      with:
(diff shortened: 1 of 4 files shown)


StevenMaude:
We should merge the other dependency PRs and then this one last.

.github/workflows/update-python-dependencies.yml
```diff
@@ -0,0 +1,26 @@
+name: Update Python dependencies
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  "12 6 * * *"
```
Jongmassey, line 6: AFAICT dependabot runs weekly, yet this runs at 0612h every day - is there a reason for this change?

Jongmassey (approved):

.github/workflows/update-python-dependencies.yml
StevenMaude: I think this was taken from an existing repository, so I can change it to weekly.

StevenMaude:
Awaiting merge of #2233.

.github/workflows/update-python-dependencies.yml
StevenMaude: Done.

PR #2242, opened 2024-12-13
Fix location of DB to take backup from

mikerkelly:
Fixes #2214, which was broken. 

Within the container, /app/db.sqlite3 is an apparently empty DB, /storage/db.sqlite3 is the actual app DB of several GB. I'm not sure why the former exists. But this script was backing that up, not the intended DB.

For example, when this ran in production as scheduled the first time the produced backup file is only a few tens of bytes in size.

```
mikerkelly@dokku3:~$ ls /var/lib/dokku/data/storage/opencodelists/backup/db/ -lrth
total 4.0K
-rw-r--r-- 1 opencodelists opencodelists 114 Dec 12 01:00 2024-12-12-db.sqlite3.gz
```

### DIFF
--- a/deploy/bin/backup.sh
+++ b/deploy/bin/backup.sh
@@ -9,13 +9,16 @@ BACKUP_DIR="$STORAGE_ROOT/backup/db"
 mkdir "$BACKUP_DIR" -p
 
 # Take a datestamped backup.
-BACKUP_FILENAME="$BACKUP_DIR/$(date +%F)-db.sqlite3"
-sqlite3 "$STORAGE_ROOT/db.sqlite3" ".backup $BACKUP_FILENAME"
+BACKUP_FILENAME="$(date +%F)-db.sqlite3"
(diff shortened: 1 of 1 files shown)


deploy/bin/backup.sh
```diff
@@ -2,31 +2,29 @@
...
 set -euxo pipefail
 
-# We are changing the backup format and where they are stored. We want to
-# retain 30 days of backups across both locations and formats. Once there
-# are none of the old format remaining, this can be updated to just refer
-# to the new location.
-REPO_ROOT="/app"
-ORIGINAL_BACKUP_DIR="/storage"
-BACKUP_DIR="$ORIGINAL_BACKUP_DIR/backup/db"
+STORAGE_ROOT="/storage"
```
Jongmassey, line 5: I think this should use the `DATABASE_DIR` [env  var](https://github.com/opensafely-core/opencodelists/blob/fe4bdb09469441e23a1c552fe47305012a8b9d45/DEPLOY.md?plain=1#L34) instead

Issue #2225, opened 2024-12-10
SPIKE Investigate the use of Digital Ocean Block Storage for the database backups


lucyb:
**EDIT: This spike is now to only look at using block storage for the database backups**

We think that using Digital Ocean [Block Storage](https://www.digitalocean.com/products/block-storage) will be a cost-effective way of increasing storage capacity, while also providing us with faster database access and improved backups.

In particular, we've identified the following problems:
* We can't store backups of the Coding Systems databases on Dokku3 due to a lack of space (#2153)
* Backups of the core application database are currently stored on the droplet itself.
* Restoring data from a droplet is likely to be annoying, and will be time consuming to test and to document (#2128)
* Droplet backups are are limited. We have one per week, with four retained. We would like something more frequent for databases and something that's easier to access.

There are potentially other solutions to this, such as migrating to a managed Postgres service or storing our backups on Digital Ocean [Spaces object storage](https://www.digitalocean.com/products/spaces) (i.e their S3 equivalent). So this spike should demonstrate that Block Storage is the better solution (or not).

Questions to answer
1. How much will Block Storage cost
2. How is Block Storage used with a Droplet?
3. Where will we store a copy of the Block Storage configuration?
4. How will we migrate over to using Block Storage?
5. How will we test that the migration has been successful and demonstrate there are no performance problems? (Edit: I think Sentry cron monitoring will be enough)
6. How does Block Storage compare to Object Storage?

bloodearnest:
> while also providing us with faster database access

I would be very surprised if this was true. Block storage is essentially a networked file system, albeit at a very different way than NFS. Unless something has drastically changed in cloud land here, I would expect it to be measurably slower.  Possibly at an acceptable level, especially for reads and if we tune sqlite things like mmap and cache sizes to keep stuff in the fs page cache.

So I would encourage performance testing to be done early (perhaps enabling sqlite otel telemetry if not already?) as it may provide a good early failure signal.  For example, there were recently discussions about how N+1 queries are not nearly as much of a problem with sqlite as they are with postgres - that might be less the case with the volume mounted on Block storage.

FWIW, the model I've seen used with sqlite in the cloud is to have the sqlite db on fast local storage, and use backups or litestream off host for durability.

Some other ideas:

- Have we considered mounting a BS volume for "local" backups of all all dbs?
[...]

Issue #2153, opened 2024-11-19
Backup coding system databases


lucyb:
As far as I'm aware, we don't currently backup our coding system databases. If we lost these, they might be time consuming and tricky to restore. We have one database per coding system release and generally consider them to be immutable. In total they take up 38G (at [the time of writing](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1731949109093479?thread_ts=1731940750.156009&cid=C069SADHP1Q)) and will get larger over time.

We only have 155GB of disk space on dokku3 and are currently using half of it. So, it's probably not a good idea to backup the databases to the same machine. We should consider backing them up to Digital Ocean [Spaces Object Storage](https://www.digitalocean.com/products/spaces) and also storing the downloaded files there.

We should manage the backups in a similar way to #2151, by using a cron job. However, since the databases aren't written to we should be able to copy the files directly rather than use the sqlite cli.

lucyb:
I'm removing this from the board, as I believe this requirement will be fulfilled by using Digital Ocean Block Storage for the OpenCodelists databases. However, given that we don't know we're going to move over to that yet I'm keeping the ticket open as a reminder.

Issue #2128, opened 2024-11-13
Document restoring core and coding system release DBs from droplet backup


iaindillingham:
Document the process of restoring the coding system release DBs from a droplet backup. Doing so will mean actually creating a new droplet, and restoring a droplet backup into it, to verify that the process works. #2116 defines the terms _coding system release DBs_, _droplet backup_, and _droplet_.

lucyb:
I'm closing this as I believe it will be made redundant by https://github.com/opensafely-core/opencodelists/issues/2225 and a move over to Digital Ocean Block Storage.

mikerkelly:
This was reopened as if we rely on the droplet storage as a fallback restore option as mentioned in #2246 it would be beneficial to verify that we can actually do that and understand what it would involve.

PR #2246, opened 2024-12-15
Document backup strategy in a draft ADR

### LINKED ISSUE #2151: Use sqlite cli to backup core database

mikerkelly:
Partially fixes #2151.

This documents the current context and decisions after changes from #2151. I'm proposing that we keep this in draft until backup-related issues in the Onboarding OpenCodelists initiative are resolved or deferred. Then the outcomes can be combined, which will be easier to follow in future than multiple updates in a short period of time. E.g., #2225, #2128, #2153.

### DIFF
--- a/docs/adr/0001-backup-strategy.md
+++ b/docs/adr/0001-backup-strategy.md
@@ -1,0 +1,56 @@
+# Backup strategy
+
+Date: 2024-12
+
+## Status
+
+Draft
+
+I'm proposing that we keep this in draft until backup-related issues in the
+Onboarding OpenCodelists initiative are resolved or deferred. Then the outcomes
+can be combined into one ADR, which will be easier to follow in future.
+
+## Context
+
+OpenCodelists runs on a DigitalOcean droplet with Dokku-managed docker
+containers serving Django apps. There are multiple SQLite databases. The
+storage file system is local. Digital Ocean offer Block and Object remote
+storage solutions but we don't currently use those.
+
+The droplet is backed up weekly, with four backups retained.
+
+User-generated codelist content is stored in the core database along with
+information on users and organizations. This is the only source of truth (SOT)
+for the user-generated content, which takes some time to develop if using the
+codelist builder UI. So codelist data loss is costly.
+
+Other databases hold representations of the various coding systems and mappings
+against which a codelist may be developed, displayed, and interpreted. The SOT
+is the source data files provided by a third-party. These are written to once
+then stored read-only. In theory, we can re-run our code against source to
+re-generate the databases, reducing the cost of data loss. In practice, it may
+be difficult to do this in some cases due to the original code or source data
+being unavailable. Some of the source data is stored locally, but not all.
+
+All other material required to deploy the site is under source control.
+
+## Decision
+
+We will back up the core database nightly using scheduled SQLite CLI commands.
+The backup will be stored in the same file system. The backup will be
+compressed. We will retain these backups for 30 days.
+
+We will not back up the coding system databases other than through the
+continuing weekly droplet backups.
+
+## Consequences
+
+The core database can be straightforwardly restored from the local nightly
+backups, limiting core data loss to one day if the droplet file system is not lost.
+
+The coding system and mapping databases can only be restored from the droplet
+backups or recreated from source. Their read-only permissions partially mitigate
+the risk of accidental deletion.
+
+Restoring from the droplet backups could lead to up to a week of data loss and
+it's unclear how difficult it would be to execute restoration.
--- a/docs/adr/README
+++ b/docs/adr/README
@@ -1,0 +1,30 @@
+We record architecturally significant decisions in Architecture Decision
+Records (ADRs). Refer to [this article by Michael
+Nygard](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions)
+for an introduction and guidelines.
+
+ADRs are stored as markdown-format text files in this directory, each with a
+leading 4-digit incrementing identifier and a meaningful name (e.g.,
+`0001-record-name.md`).
+
+Determining what is "architecturally significant" is subjective. Nygard defines it as
+"decisions that affect the structure, non-functional characteristics,
+dependencies, interfaces, or construction techniques". Examples include:
+
+- Decisions about the overall purposes of the system and its interfaces with
+  other elements of its ecosystem.
+- Key technology, framework, and system design decisions.
+- Decisions about how the code is deployed and managed in production.
+- Decisions impacting scalability, security, or maintainability.
+- Adopting a design pattern that influences multiple components.
+
+Purely functional or business-logic decisions are unlikely to be architectural.
+
+If the context, rationale, and consequences of a decision can be easily
+documented in a single location and are still easily discoverable, it may be
+preferable to avoid creating an ADR. Use ADRs when the decision has long-term
+implications or needs greater visibility for future reference.
+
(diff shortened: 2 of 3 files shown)


Jongmassey:
It's a bit meta but in other projects where we use ADRs we record the architectural decision to use ADRs as adr0000 [eg](https://github.com/opensafely-core/job-server/blob/main/docs/adr/0000-record-architecture-decisions.md), do we want to do this here, too?

mikerkelly:
> we record the architectural decision to use ADRs as adr0000 [eg](https://github.com/opensafely-core/job-server/blob/main/docs/adr/0000-record-architecture-decisions.md), do we want to do this here, too?

It's a nice idea, but I don't like that it's not a good example of an ADR. It doesn't include the kind of detail a real one would. And a documentation choice is almost by definition not a decision about _architecture_ but rather _process_.

I think I'll add a README in the directory and update DEVELOPERS.md to point at that.

docs/adr/README
lucyb, line 1: I really like the write-up of what a decision is and specifically how we are using ADRs here. I wonder if some of this could be made into guidance for the wider BI. That said, I think a README in the same directory as the ADRs is more likely to be read than something in the team manual.

docs/adr/0001-backup-strategy.md
```diff
@@ -0,0 +1,56 @@
...
+The backup will be stored in the same file system. The backup will be
+compressed. We will retain these backups for 30 days.
+
+We will not back up the coding system databases other than through the
+continuing weekly droplet backups.
+
+## Consequences
+
+The core database can be straightforwardly restored from the local nightly
+backups, limiting core data loss to one day if the droplet file system is not lost.
```
lucyb, line 49: It may be difficult to restore the core database from a droplet backup. As sqlite is a single file, I believe that if the droplet backup is taken at the same time as a process is writing to it, then we can't guarantee consistency of the data. It might be worth checking that and in practice, I think it's quite unlikely given the time the droplet backup is likely to be taken. However, that does make the locally stored backups more important (those could be restored from a droplet backup if needed)

docs/adr/0001-backup-strategy.md
```diff
@@ -0,0 +1,56 @@
...
+
+The coding system and mapping databases can only be restored from the droplet
+backups or recreated from source. Their read-only permissions partially mitigate
```
lucyb, line 52: I'm not sure that all of the coding system and mapping databases have read-only file permissions. I also don't think that we're creating a read-only database connection.

docs/adr/0001-backup-strategy.md
```diff
@@ -0,0 +1,56 @@
...
+
+We will back up the core database nightly using scheduled SQLite CLI commands.
+The backup will be stored in the same file system. The backup will be
+compressed. We will retain these backups for 30 days.
+
+We will not back up the coding system databases other than through the
+continuing weekly droplet backups.
+
+## Consequences
+
```
lucyb, line 47: Are there any consequences to compressing the core database backup? I assume we're doing it to save disk space?

lucyb (approved):
Thank you for writing this up. I think it's likely to be extremely helpful. I've made some comments, but am happy to leave it to you to address them and merge it.

docs/adr/0001-backup-strategy.md
bloodearnest: The sqlite `.backup` command the ADR suggests takes care of this, by taking a full db lock before doing the copy.

Additionally, it handles the case where there are multiple files related to the db, such as when WAL mode is enabled and there is an additional WAL file.

mikerkelly: [SQLite, in any `journal_mode` ](https://www.sqlite.org/pragma.html#pragma_journal_mode)other than `memory` or `off`, writes transactions to additional files before they are committed. The core OpenCodelists database uses `wal` mode.

[The SQLIte backup API](https://www.sqlite.org/backup.html) ensures that when completed "the backup database contains a consistent and up-to-date snapshot of the original." The nightly backup process described will use this approach. If writes are performed on the source database during the backup process they will slightly delay the process but be included in the backup database file.

If the DigitalOcean snapshot for the backup is taken while a transaction is being written the eventual droplet backup may contain a corrupt or inconsistent database. I don't know if the droplet backup process does or could do something to ensure this cannot happen, such as entering "maintenance mode" before taking the backup. If not, we might have to restore the latest daily backup after restoring the droplet to get the database in a usable state. And it might be prudent to do this in case of usable but inconsistent data.

Note that this would increase the maximum data loss to 7 days plus the lag between taking the last daily backup and taking the weekly snapshot. If nothing else, I suggest we should make that lag short.

mikerkelly: It seems sensible to use `wal` in the core database as reads may sometimes be much more common than writes and it helps address concurrency concerns. I checked the mode by downloading and restoring a backup locally and querying `pragma journal_mode;` giving `wal`. However I'm a little concerned that I cannot see where this is configured in the repo. I've only looked for a couple of minutes. If it's not captured in code or in manual deployment instructions, it should be.

PR #2254, opened 2024-12-17
BNF to dm+d downloader hotfixes

Jongmassey:
Some fixes from testing in prod, plus a unit test for the BNF to dm+d  `import_release` which is called by the scheduled import data management command


mikerkelly:
I can see why the directory being missing was easy to miss locally, but I'm a little concerned that a function call signature being wrong wasn't noticed. Could our automated testing of this functionality be better?

mikerkelly (approved):

mikerkelly:
Testing looks much improved, thanks.

We talked [in Slack](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1734453445053289?thread_ts=1734452051.347749&cid=C080XJJ0KDJ) about how to approach this. Possibly an integration test calling the command without mocking would also be useful, but not critical. The file system could be mocked to reduce test execution time but that's much bigger than this PR.

PR #2262, opened 2024-12-18
Unit test and fix bnfdmd downloader

Jongmassey:
download_latest_release() is the entry point for
the Downloader class that is called by the
import_latest_data management command.

This adds unit tests of that entry point given
that the latest mapping file has and has not
already been downloaded.

In case of the latter, the aforementioned
management command expects a ValueError to be
raised, this adds a test for that condition and a
fix such that the test passes.

[Sentry link for originating issue](https://ebm-datalab.sentry.io/issues/6157385078/?alert_rule_id=1877862&alert_type=issue&notification_uuid=cc036dbc-02c1-4ad9-bac6-8d0272d6e44c&project=5248013&referrer=slack)

### DIFF
--- a/mappings/bnfdmd/tests/conftest.py
+++ b/mappings/bnfdmd/tests/conftest.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 from unittest.mock import patch
+from urllib.parse import quote
 
 import pytest
 import responses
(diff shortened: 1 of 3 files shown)


mappings/bnfdmd/tests/conftest.py
```diff
@@ -12,31 +13,43 @@
...
+            body=f,
+            status=200,
+        )
+        yield mocked_responses
 
 
 @pytest.fixture
 def mocked_responses():
     with responses.RequestsMock() as rsps:
+        rsps.add(
```
mikerkelly, line 37: I'm not sure I understand why this fixture needed to change in this commit. I think `test_get_latest_release_with_existing` is the only client and will pass without this change.

This is also a different meaning than `mocked_responses` in other `conftest.py` where it means something like "responses are just empty mocks". So if this specific fixture needs to return certain content can its name reflect that?

mikerkelly:
`test_data_downloder.py` should be `mappings/bnfdmd/tests/test_data_downloader.py`.

Fix and tests look good. I wondered about the purpose of the change to `mocked_responses` in conftest.py, could this be explained in the commit message if needed or reverted if not?

mikerkelly (approved):

PR #2271, opened 2024-12-27
Remove `environs`

### LINKED ISSUE #2222: Remove environs
See Dave's comment in this [PR](https://github.com/opensafely-core/opencodelists/pull/2220#issuecomment-2531829119).

Airlock specifically didn't use environs because its transitive dependencies seemed pretty heavyweight for the value it gives. We're only using it in settings.py, so it seems unlikely we're getting the full value from it and that we can easily remove it.

Jongmassey:
fixes #2222

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -15,9 +15,9 @@ import re
 import sys
 from pathlib import Path
 
+import dj_database_url
 import sentry_sdk
 from django.contrib.messages import constants as messages
-from environs import Env
 from furl import furl
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import ignore_logger
@@ -30,29 +30,50 @@ __import__("sqlean")
 sys.modules["sqlite3"] = sys.modules.pop("sqlean")
 
 
-# Read env file/vars
-env = Env()
-env.read_env()
+_missing_env_var_hint = """\
+If you are running commands locally outside of `just` then you should
+make sure that your `.env` file is being loaded into the environment,
+which you can do in Bash using:
+
+    set -a; source .env; set +a
+
+If you are seeing this error when running via `just` (which should
+automatically load variables from `.env`) then you should check that
+`.env` contains all the variables listed in `dotenv-sample` (which may
+have been updated since `.env` was first created).
+
+If you are seeing this error in production then you haven't configured
+things properly.
+"""
+
+
+def get_env_var(name):
+    try:
+        return os.environ[name]
+    except KeyError:
+        raise RuntimeError(
+            f"Missing environment variable: {name}\n\n{_missing_env_var_hint}"
+        )
 
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = Path(__file__).resolve().parent.parent
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = env.str("SECRET_KEY")
+SECRET_KEY = get_env_var("SECRET_KEY")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = env.bool("DEBUG", False)
+DEBUG = os.environ.get("DEBUG", default=False) == "True"
 
-DEBUG_TOOLBAR = env.bool("DJANGO_DEBUG_TOOLBAR", default=False)
+DEBUG_TOOLBAR = os.environ.get("DJANGO_DEBUG_TOOLBAR", default=False) == "True"
 
-BASE_URLS = env.list("BASE_URLS", [])
+BASE_URLS = os.environ.get("BASE_URLS", default="").split(",")
 # note localhost is required on production for dokku checks
 BASE_URLS += ["http://localhost:7000"]
 
-ALLOWED_HOSTS = [furl(base_url).host for base_url in BASE_URLS]
+ALLOWED_HOSTS = [furl(base_url).host for base_url in BASE_URLS if base_url]
 
-IN_PRODUCTION = env.bool("IN_PRODUCTION", False)
+IN_PRODUCTION = os.environ.get("IN_PRODUCTION", default=False) == "True"
 
 if IN_PRODUCTION:
     # This setting causes infinite redirects
@@ -153,13 +174,14 @@ WSGI_APPLICATION = "opencodelists.wsgi.application"
 # see coding_systems.versioning.models.update_coding_system_database_connections (called
 # from coding_systems.versioning.apps)
 DATABASES = {
-    "default": env.dj_db_url("DATABASE_URL", "sqlite:///db.sqlite3"),
+    "default": dj_database_url.parse(
+        os.environ.get("DATABASE_URL", default="sqlite:///db.sqlite3")
+    ),
     "OPTIONS": {"timeout": 45},
 }
 
-DATABASE_DIR = Path(
-    env("DATABASE_DIR", BASE_DIR)
-)  # location of sqlite files e.g. /storage/
+DATABASE_DIR = Path(os.environ.get("DATABASE_DIR", default=BASE_DIR))
+# location of sqlite files e.g. /storage/
 DATABASE_DUMP_DIR = DATABASE_DIR / "sql_dump"
 CODING_SYSTEMS_DATABASE_DIR = DATABASE_DIR / "coding_systems"
 DATABASE_ROUTERS = ["opencodelists.db_utils.CodingSystemReleaseRouter"]
@@ -215,15 +237,17 @@ STORAGES = {
 
 # https://docs.djangoproject.com/en/5.0/howto/static-files/
 # Note: these *must* be strings. If they are paths, we cannot cleanly extract them in ./scripts/collect-me-maybe.sh
-BUILT_ASSETS = env.path("BUILT_ASSETS", default=BASE_DIR / "assets" / "dist")
+BUILT_ASSETS = Path(
+    os.environ.get("BUILT_ASSETS", default=BASE_DIR / "assets" / "dist")
+)
 STATICFILES_DIRS = [
     str(BASE_DIR / "static"),
     str(BUILT_ASSETS),
 ]
-STATIC_ROOT = env.path("STATIC_ROOT", default=BASE_DIR / "staticfiles")
+STATIC_ROOT = Path(os.environ.get("STATIC_ROOT", default=BASE_DIR / "staticfiles"))
 STATIC_URL = "/static/"
 
-ASSETS_DEV_MODE = env.bool("ASSETS_DEV_MODE", default=False)
+ASSETS_DEV_MODE = os.environ.get("ASSETS_DEV_MODE", default=False) == "True"
 
 DJANGO_VITE = {
     "default": {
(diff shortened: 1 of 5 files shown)


opencodelists/settings.py
```diff
@@ -30,29 +30,58 @@
...
+have been updated since `.env` was first created).
+
+If you are seeing this error in production then you haven't configured
+things properly.
+"""
+
+
+def get_env_var(name, **kwargs):
+    try:
+        env = os.environ[name]
```
mikerkelly, line 52: I'm not sure about `env` as a variable name. `env_var_value`?

opencodelists/settings.py
```diff
@@ -30,29 +30,58 @@
...
+        if "default" in kwargs:
+            t = type(kwargs["default"])
+            if t is bool:
+                return env == "True"
+            return t(env)
+        return env
+    except KeyError:
+        if "default" in kwargs:
+            return kwargs["default"]
+        raise RuntimeError(
```
mikerkelly, line 62: Arguably `KeyError` is a better fit than `RuntimeError` if something is missing from the `os.environ` dict. I don't feel strongly though.

mikerkelly (commented):
Could you please talk me through the `docker/docker-compose.yaml` change a bit more? I think in production we aren't using a `.env` file or `just`  -- those are dev tools. Is that right? And if so, does changing this docker-related file affect production at all? I don't think we should be doing that. I think in production the environment variables are configured by a different mechanism -- `dokku config:set opencodelists` etc. in `DEPLOY.md`.

I notice the second commit ([6da1873](https://github.com/opensafely-core/opencodelists/pull/2271/commits/4dceb7f62f418bb455f131fbe2d95e30ee8ad6ce)) doesn't pass all the checks. Somehow [the hint is being displayed when run in CI](https://github.com/opensafely-core/opencodelists/actions/runs/12517138207/job/34917415578#logs). Should this be happening? I'm not sure I understood why the other commits pass the checks though. I think if a repo with suitable env configuration this wouldn't happen so maybe it doesn't matter too much. Could some commits be merged to be atomic?

`get_env_var` seems to be doing three things: a parallel implementation of `os.getenv`; coercing `"True"` to `True`; and displaying the hint if a value is missing with no provided default.

I'm not sure handling defaults etc. ourselves here is gaining us anything over using `os.getenv`.

I think the coercing doesn't belong with the others and combing them is hard to follow. I think the variables that depend on the string `"True"` ought to be explicit about that rather than doing that logic in this function whose name doesn't indicate that.

I think the code as it is now is complicated enough that it ought to have a docstring and some tests. Maybe we can avoid that by simplifying a bit, something like (not tried):

```python
def get_env_var(key, default=None):
    env_var_value = os.getenv(key)
    if not env_var_value:
	raise KeyError(f"Missing environment variable: {key}\n\n{_missing_env_var_hint}")
    return env_var_value

...

DEBUG = get_env_var("DEBUG", default=False) == "True"
DEBUG_TOOLBAR = get_env_var("DJANGO_DEBUG_TOOLBAR", default=False) == "True"
IN_PRODUCTION = get_env_var("IN_PRODUCTION", default=False) == "True"
ASSETS_DEV_MODE = get_env_var("ASSETS_DEV_MODE", default=False) == "True"
```

Jongmassey:
>I notice the second commit ([6da1873](https://github.com/opensafely-core/opencodelists/pull/2271/commits/4dceb7f62f418bb455f131fbe2d95e30ee8ad6ce)) doesn't pass all the checks. Somehow [the hint is being displayed when run in CI](https://github.com/opensafely-core/opencodelists/actions/runs/12517138207/job/34917415578#logs). Should this be happening? I'm not sure I understood why the other commits pass the checks though. 

Yep, this relates to an unexpected source of pain which I though was related to a cryptic message in a code comment in airlock about order of precedence of env vars when running a docker container; _however_ it wasn't actually that but was in fact the case that until the third commit in this PR, the env vars were being loaded from `.env` when running in a docker container.  The change to `docker-compose.yaml` in  a724d3b70cf4f2737fe980f93e61116a7f15bdde  fixes this. This previously wasn't necessary  as `environs` did this loading. Airlock [has similar configuration](https://github.com/opensafely-core/airlock/blob/6f747f192df94d75e551a2e290fe81edf3aaec11/docker/docker-compose.yml#L33) in its docker-compose

opencodelists/settings.py
Jongmassey: I don't have feelings either way, this was lifted straight from the airlock code

Jongmassey:
>Could you please talk me through the docker/docker-compose.yaml change a bit more? I think in production we aren't using a .env file or just -- those are dev tools. Is that right? And if so, does changing this docker-related file affect production at all? I don't think we should be doing that. 

Right...

>I think in production the environment variables are configured by a different mechanism -- dokku config:set opencodelists etc. in DEPLOY.md.

They are, but also prior to this change, `opencodelists/settings.py` [uses](https://github.com/opensafely-core/opencodelists/blob/003f6941335710a1db3b3b218b001bcc9330fcc7/opencodelists/settings.py#L34-L35) environs to load these out of the `.env` file, this happens regardless of whether run in prod or dev; directly or via Docker. There isn't a `.env` file in the repo, but instead a [`dotenv-sample`](https://github.com/opensafely-core/opencodelists/blob/main/dotenv-sample) which is copied by the `_env` `just` recipe to a `.env` file. Since this recipe isn't called in prod, there isn't an `.env` file present in prod, and thus this configuration won't many any difference.

```sh
Jongmassey@dokku3:~$ dokku enter opencodelists
groups: cannot find name for group ID 10003
I have no name!@28bd6776c3df:/app$ ls .env
ls: cannot access '.env': No such file or directory
```

There are also some env vars [defined](https://github.com/opensafely-core/opencodelists/blob/003f6941335710a1db3b3b218b001bcc9330fcc7/docker/docker-compose.yaml#L33-L37) in the `prod` service in `docker-compose.yaml` but these are dummy values just used when checking the prod image builds correctly. They are overwritten by the `dokku config:set` operations you mention. Even if, for some reason, a `.env` file made it to production, these would also be overridden similarly.


So, since the pre-existing configuration supported provision of env vars for CI in docker from a `.env` file via `environs` removing `environs` means we've got to replace it with something else, in this case by specifying the `.env` file in `docker-compose.yaml`, as does [airlock](https://github.com/opensafely-core/airlock/blob/6148179ac2b0c669b0388fbaf2ce31ddeeaa3324/docker/docker-compose.yml#L33).

Jongmassey:
>get_env_var seems to be doing three things: a parallel implementation of os.getenv; coercing "True" to True; and displaying the hint if a value is missing with no provided default.
>
>I'm not sure handling defaults etc. ourselves here is gaining us anything over using os.getenv.
>
>I think the coercing doesn't belong with the others and the combined logic is harder to follow. I think the variables that depend on the string "True" ought to be explicit about that rather than doing that logic in this function whose name doesn't indicate that.

The "parallel implementation of os.getenv" and "hinting" aspects are taken from [the airlock implementation](https://github.com/opensafely-core/airlock/blob/6f747f192df94d75e551a2e290fe81edf3aaec11/airlock/settings.py#L39-L45). However, I found a couple of things odd in airlock's `settings.py`, mostly to do with defaults and type coercion:

Despite having a function called `get_env_var` it also calls `os.environ.get` directly where a default value is required. It seemed seemed sensible to me to roll that default-provision behaviour into `get_env_var`, although acknowledge that in those cases it will never hit the missing value case.

It also does the `"True"`=>`True` dance in [a couple of places](https://github.com/opensafely-core/airlock/blob/6f747f192df94d75e551a2e290fe81edf3aaec11/airlock/settings.py#L63-L66) and so also it seemed sensible to make `get_env_var` the one place for getting env vars in the desired destination type. In my testing, I found that the variables previously got by `env.bool()` really did need to be bools and not strings and that changing the things that consume them to be happy with strings rather than bools would be impractical.


In summary, it's more of an attempt to be a drop-in replacement for `env.[something]()` and thus has some more complexity than just calling `os.getenv()`

mikerkelly:
Thanks for the explanation of the CI fail + docker details, Jon.

I think before this ticket in prod the `read_env` call would have found an `.env` file if present, but it wouldn't have overridden any pre-existing env vars (those coming from dokku's configuration). Now, I think the prod app won't try to access `.env` at all, which seems even better. Applications should take the environment they're given. I'd not properly appreciated that the `docker-compose.yaml` `prod` service is only used for building the prod image, nothing to do with running it. So that's all good.

mikerkelly:
On the approach to the getting the env vars in settings.py I suppose I'm saying I still prefer the Airlock approach. But that isn't a very strongly held opinion and if you disagree I'd be keen to understand why.

I think functions should ideally do one thing as concisely as possible and therefore be easy to read and understand through good naming, structure, and minimal-but-good docs. I don't see that a multi-function "one-stop-shop" function for "all of the env var accesses in this module" is better than the specific cases that want the hint, default, or parsing behaviour invoking those in each place that they're needed. That seems more concise, precise, explicit, and readable, to me.

> The "parallel implementation of os.getenv" and "hinting" aspects are taken from AIrlock

I saw that that's where the hinting idea comes from and I can marginally see the dev experience benefit to providing a hint where there is no default (although I would like the function name or docs to indicate it's only useful where there is no default, as that's not so obvious). (Marginal against the cost of baking in some assumptions about the application's possible environments, which may end up out-of-date or just wrong if copied somewhere else.)

The reference to `os.getenv` I made was around the default-provisioning behaviour that the Airlock implementation doesn't do -- it just wraps `os.environ[name]`.

> Despite having a function called `get_env_var` it also calls `os.environ.get` directly where a default value is required. It seemed seemed sensible to me to roll that default-provision behaviour into `get_env_var`, although acknowledge that in those cases it will never hit the missing value case.

I don't think I understand the intended benefit to writing our own default-provision behaviour when either `os.environ.get` or `os.getenv` already handle those cases well. It complicates the function and is one more thing to understand when reading the file and one more thing to maintain.

> It also does the `"True"`=>`True` dance in [a couple of places](https://github.com/opensafely-core/airlock/blob/6f747f192df94d75e551a2e290fe81edf3aaec11/airlock/settings.py#L63-L66) and so also it seemed sensible to make `get_env_var` the one place for getting env vars in the desired destination type.

I think it's better to do that conversion explicitly and directly where those bool variables are set than indirectly in `get_env_var` that is also doing other things.

I don't think there's a problem with writing that conversion in multiple places in a settings file (which is not much like reusable code components). A single-purpose function to do the conversion would be reasonable but I think it's marginal whether that would be more or less readable. But mixing it into a multi-purpose function doesn't seem like a good idea to me.

>In my testing, I found that the variables previously got by `env.bool()` really did need to be bools and not strings and that changing the things that consume them to be happy with strings rather than bools would be impractical.

That seems right to me.

Jongmassey:
I've rewritten the changes to `settings.py` such that `get_env_var()` doesn't do any defaulting or type casting and instead rely on `os.environs.get()` and some inline casting/comparisons. Having done this, there are only two instances remaining where it is actually called but given it proved useful in finding the issue with running in docker I still think it merits inclusion.

Care to re-look @mikerkelly  ?

mikerkelly (approved):
Thanks, Jon. Looks good to me.

PR #2275, opened 2024-12-30
Add Sentry monitoring for cron jobs

### LINKED ISSUE #2257: Sentry alerts for cron job issues
# Why are we doing this?

[Sentry](https://ebm-datalab.sentry.io/crons/?project=5248013&statsPeriod=90d) can perform cron monitoring. The deployed container runs at least one scheduled cron job to perform regular tasks such as taking backups. It would be beneficial if we were notified if these failed or had other issues, so that we can investigate and fix them.

OpenCodelists Sentry alerts are in place for its current cron jobs.

# What are we doing?

You can add [Sentry crons here](https://ebm-datalab.sentry.io/crons/?project=5248013&statsPeriod=90d) and see how it is configured in [Job Server here](https://ebm-datalab.sentry.io/crons/?project=5443358&statsPeriod=90d).

[app.json](https://github.com/opensafely-core/opencodelists/blob/main/app.json) configures the backup cron. Check if there are any others.

---

[Defining delivery tasks guidance](https://docs.google.com/document/d/1F7yBNYnMoCW3liWVvjPXFHnrnCU4qeAOhlKBVD1-8lg/edit?tab=t.0#heading=h.qkl5vvsdkx98)

Jongmassey:
fixes #2257


deploy/bin/run_cron.sh
```diff
@@ -0,0 +1,64 @@
...
+
+JOB_NAME="$1"
+if ! test -z "$2";
+then
+    JOB_ARG="$2"
+fi
+
+# detect is "self-managed" or "dokku-managed" cron job
+# by working out if we're in or outside the container
+BINPATH="/var/lib/dokku/data/storage/opencodelists/"
```
mikerkelly, line 15: Can this just include deploy/bin/ then the "outside container" path doesn't need to append it?

mikerkelly: Or put another way I'm not sure why we can assume /var/lib/dokku/data/storage/opencodelists/ will exist if we're outside the container, but we can't assume /var/lib/dokku/data/storage/opencodelists/deploy/bin will exist in that case?

mikerkelly: Also, this `BINPATH` is not a path to /bin/ it's the root of the storage... so the name is a bit off / confusing.

deploy/bin/run_cron.sh
```diff
@@ -0,0 +1,64 @@
...
+        SENTRY_DSN=$(dokku config:get opencodelists SENTRY_DSN)
+
+        # combine job name and arg for reporting to Sentry
+        JOB_NAME+="_$JOB_ARG"
+    else
+        echo "deploy/bin not deployed"
+        exit 1
+    fi
+else
+    # running inside container
```
mikerkelly, line 33: We might not be if the app storage isn't deployed at all... so should we also test this `BIN_PATH` does exist?

deploy/bin/run_cron.sh
```diff
@@ -0,0 +1,64 @@
...
+    JOB_ARG="$2"
+fi
+
+# detect is "self-managed" or "dokku-managed" cron job
+# by working out if we're in or outside the container
+BINPATH="/var/lib/dokku/data/storage/opencodelists/"
+if test -d "$BINPATH";
+then
+    # running outside container
+    BIN_PATH+="deploy/bin/"
```
mikerkelly, line 19: `BINPATH` vs `BIN_PATH` mismatch?

deploy/bin/run_cron.sh
```diff
@@ -0,0 +1,64 @@
...
+    # running inside container
+    BIN_PATH="/app/deploy/bin"
+    CRONTAB=$(grep -A 2 "$BIN_PATH/run_cron.sh $JOB_NAME" "/app/app.json" | grep -oP "$CRONTAB_PATTERN")
+    BIN_PATH+="/$JOB_NAME"
+    # sentry_dsn env var is available inside container
+fi
+
+# modify the DSN to point it at the cron API
+SENTRY_CRONS=$(sed -E "s/([0-9]+$)/api\/\1/g" <<< "$SENTRY_DSN")
+SENTRY_CRONS+="/cron/$JOB_NAME"
```
mikerkelly, line 42: From looking at `DEPLOY.md` and https://docs.sentry.io/product/crons/getting-started/http/ I think I can see that this is correct, getting the second number from `SENTRY_DNS` and putting it after `/api/`, but I sure don't understand what the name `SENTRY_DNS` means, and this sed might be a bit tricky to interpret so can we say what this is doing and why, possibly with a reference to the Sentry docs?

I would also like `DEPLOY.md` to explain where `SENTRY_DNS` come from and what it means, and possibly re-name it, because I can't tell without digging into how sentry works, and I think I ought to be able to at least know where to look based on that document. But that's not really for this PR, but maybe I could raise an issue about it. (same for `MAILGUN_API_KEY`)

deploy/bin/run_cron.sh
```diff
@@ -0,0 +1,64 @@
...
+
+function sentry_cron_start() {
+curl -X POST "${SENTRY_CRONS}" \
+    --header 'Content-Type: application/json' \
+    --data-raw "{\"monitor_config\": {\"schedule\": {\"type\": \"crontab\", \"value\": \"$CRONTAB\"}}, \"status\": \"in_progress\"}"
+}
```
mikerkelly, line 48: I'm unsure why the #2279 approach has a simplified `in progress` curl compared to this PR?

mikerkelly:
This is in draft and you asked in Slack for an in-flight review. What's left to do?

The simplified approach you point to in #2279 looks like it would have worked okay but I agree that DRY is preferable for the actual HTTP API interactions as there's a bit of processing and logic to do, which is more manageable and extensible when centralized. I really like automating the deployment steps and getting the schedule from config.

How have you tested what you've done so far? I guess for bash scripts automated testing will be a bit tricky?

I like the cron jobs docs added to `DEPLOY.md`. Can this also include a link to the sentry page and docs and briefly explain how we will do the plumbing in the Sentry web UI once this is deployed? I think that would be part of a full redeploy and could save someone some learning time if we record it now, while we're doing it the first time. I think this is using the HTTP Sentry API as [documented here](https://docs.sentry.io/product/crons/getting-started/http/)? 

How does this compare to our approach in Job Server? I think we use the SDK/CLI approach there but to be honest I haven't understood how it works there from looking for a few minutes.

Jongmassey:
>I think we use the SDK/CLI approach there but to be honest I haven't understood how it works there from looking for a few minutes.

We do use the Python SDK since Job Server has its own [job scheduling system](https://github.com/opensafely-core/job-server/tree/main/jobserver/jobs) (why yes, that is another overload for _job_) which is all Python and so we can use the (far nicer) Python SDK.

In this case it's all bash scripts so making HTTP requests with CURL is the most universally compatible approach. If we were, for example, to write a small python wrapper for the SDK we couldn't necessarily rely on the SDK being installed in every python environment these bash scripts get called from.

deploy/bin/run_cron.sh
Jongmassey: This approach sends the _monitor config_ (in this case, just the schedule in crontab, extracted from either the `*_crontab` file(s) or the `app.json`).

This is to avoid having to manually configure (and remember to reconfigure in case the running config changes) the Sentry cron monitor in the Sentry web UI.

We can(/should) send more configuration via this endpoint [as we do in job server](https://github.com/opensafely-core/job-server/blob/1fbe102a5db36d73d6906feb6ac09b2732dca20e/services/sentry.py#L37-L53)

Jongmassey:
>How have you tested what you've done so far? I guess for bash scripts automated testing will be a bit tricky?

It's tricky for sure. I've tested by calling `run_crons` with a dummy monitor/script name and then deleting the resultant monitor from the Sentry web UI as not to cause notification noise when it's not run again on a schedule. 

From poking at the web UI, it's possible to create a monitor with a failure threshold of 720 missed "check-ins", a missed "check in" threshold of 525,600 mins and a check-in timeout of 40,320 mins so having a test that hits a designated test monitor with these configured values should hopefully reduce/avoid notification noise for the team.

deploy/bin/run_cron.sh
```diff
@@ -0,0 +1,64 @@
...
+else
+    # running inside container
+    BIN_PATH="/app/deploy/bin"
+    CRONTAB=$(grep -A 2 "$BIN_PATH/run_cron.sh $JOB_NAME" "/app/app.json" | grep -oP "$CRONTAB_PATTERN")
+    BIN_PATH+="/$JOB_NAME"
+    # sentry_dsn env var is available inside container
+fi
+
+# modify the DSN to point it at the cron API
+SENTRY_CRONS=$(sed -E "s/([0-9]+$)/api\/\1/g" <<< "$SENTRY_DSN")
```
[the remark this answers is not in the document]
Jongmassey: Yep, the name I used from the example in the sentry docs but your suggestion is clearer

deploy/bin/run_cron.sh
Jongmassey: oops, good catch thanks

deploy/bin/run_cron.sh
Jongmassey: DSN not DNS! Data source name, not domain name system. This appears to be an annoyance of the HTTP API, the Python Crons SDK accepts the DSN format which the other bits of the monitoring SDK accept (and the project settings page of the OpenCodelists project on sentry gives you, and is configured in the dokku/env var accessed here).

The [http API for crons](https://docs.sentry.io/product/crons/getting-started/http/) wants a URL that's got all the same components as the DSN, but in a different order. I haven't the foggiest why. It's very annoying. It could be set as a separate env var, or this DSN can be turned into a the correct URL (as I have done here). Neither are ideal

deploy/bin/run_cron.sh
Jongmassey: Yes, I agree that that your approach is a clearer way to build the bin path.


re: 

>I'm not sure why we can assume /var/lib/dokku/data/storage/opencodelists/ will exist if we're outside the container, but we can't assume /var/lib/dokku/data/storage/opencodelists/deploy/bin will exist in that case?

there's no `/var/lib/dokku` inside a dokku-manged docker container, unless you're doing something _interesting_ with nested dokku that I'm not even sure is possible and we certainly aren't doing in this case

deploy/bin/run_cron.sh
Jongmassey: Yes, I think it's good to do this from a belt-and-braces approach. I've added a new commit that checks this and also reports to sentry if there are errors in finding the bin dir (in either case)

deploy/bin/import_latest_release.sh
```diff
@@ -16,7 +16,7 @@ set -euo pipefail
 # not dokku-managed
 
 # This script should be copied to /var/lib/dokku/data/storage/opencodelists/import_latest_release.sh
-# on dokku3 and run using the cronfile at opencodelists/deploy/bin/import_latest_dmd_cron
+# on dokku3 and run using the cronfiles in opencodelists/deploy/bin/
```
mikerkelly: It will just be one cronfile at least for now, right? So can this explicitly refer to `deploy/bin/cronfile`

deploy/bin/sentry_cron_functions.sh
```diff
@@ -0,0 +1,46 @@
...
+function extract_crontab() {
+    JOB_IDENTIFIER="$1"
+    CRONTAB_SOURCE="$2"
+    if [[ "$CRONTAB_SOURCE" == *json ]];
+    then
+        A=2
+    else
+        A=1
+    fi
+    CRONTAB=$(grep -A "$A" "$JOB_IDENTIFIER" "$CRONTAB_SOURCE" | grep -oP "$CRONTAB_PATTERN")
```
mikerkelly, line 14: Could change `A` => `LINES_AFTER_MATCH`.
Could add a comment explaining why these are different.

deploy/bin/sentry_cron_functions.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "$CRONTAB"
+}
+
+# Sentry DSN as used in other bits of Sentry monitoring in this project
+# is not the same as the API endpoint for Cron Monitoring.
+# This function modifies it such that it looks like what's described
+# at https://docs.sentry.io/product/crons/getting-started/http/
```
mikerkelly, line 21: Nice.

deploy/bin/run_cron.sh
mikerkelly: Ah, right. TIL. Thanks!

mikerkelly (approved):
Looks good now, good job. Made some minor comments on docs.

deploy/bin/sentry_cron_functions.sh
Jongmassey: yes, that'd be much more parseable

deploy/bin/import_latest_release.sh
Jongmassey: This comment is removed in a later commit that auto-deploys both the script file and the cronfile

Issue #2268, opened 2024-12-22
Improve SQLite configuration: `IMMEDIATE` transaction mode and `PRAGMA`s


mikerkelly:
# Why are we doing this?

There are [regular Sentry alerts](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d) for several Issues where requests fail with `OperationalError: database is locked` errors. The root cause is that our level of concurrency is currently too high for SQLite to reliably handle as currently configured. This Issue is one possible approach to improving the situation. Refer to the parent issue #2251 for more detail.

# How will we know when it's done?

There are fewer instances of `OperationalError: database is locked` errors, [as monitored through Sentry](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d).

# What are we doing?

[This article](https://kerkour.com/sqlite-for-servers), kindly pointed out in #2251 by Simon, argues for making several configuration changes from the default for SQLite in a server context, as opposed to the client context the defaults are optimized for (such as a data store for an app on a mobile phone). These include using transaction mode `IMMEDIATE` and adjusting various pragmas. These can improve performance or consistency in various ways detailed in the article. `IMMEDIATE` mode in particular may simply and directly resolve almost all of the database locked errors.

This was tried in Airlock and seems to work well. https://github.com/opensafely-core/airlock/blob/main/airlock/settings.py#L168

`IMMEDIATE` mode with SQLIte is a Django 5.1 feature. So #2115 blocks that part.

This might allow us to reduce the Django timeout (ref #2252).

Interestingly the core prod DB seems to be in WAL mode even though I can't see anything in the repo that would lead to that. That is probably the right thing to do for our context as reads may sometimes be much more common than writes and it helps address concurrency concerns. This issue would also incidentally tidy up that inconsistency.

mikerkelly:
We're hoping to do #2115 this week in what's left of this Onboarding Opencodelists initiative. This ticket deferred until the resolution of that one.

bloodearnest:
[...]

Issue #2251, opened 2024-12-17
Improve concurrency requirements to avoid `database is locked` errors


mikerkelly:
# Why are we doing this?

There are [regular Sentry alerts](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d) for several Issues where requests fail with `OperationalError: database is locked` errors. The root cause is that our level of concurrency is currently too high for SQLite to  reliably handle as currently configured. The database is locked while code writes to the database. This can result in failure after a long timeout of requests making updates, which is poor UX and could lead to their updates getting lost.

We know from [Honeycomb](https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/board/4ngq7kuV7z/OpenCodelists-Django-app) several of our views can respond slowly, taking even 10s of seconds. Possibly they are holding a database lock for an extended period. Interacting with the builder to build a codelist or search for code can both write to the database, leading to updates that need the lock occurring frequently. Possibly a queue of updates builds up and some time out before processing and the request fails.

# How will we know when it's done?

There are fewer instances of `OperationalError: database is locked` errors, as [monitored through Sentry](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&groupStatsPeriod=auto&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&page=0&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d).

This Issue might have a number of sub-issues.

# What are we doing?

According to [https://docs.djangoproject.com/en/4.2/ref/databases/#database-is-locked-errors](https://docs.djangoproject.com/en/4.2/ref/databases/#database-is-locked-errors):

>SQLite is meant to be a lightweight database, and thus can’t support a high level of concurrency. `OperationalError: database is locked` errors indicate that your application is experiencing more concurrency than `sqlite` can handle in default configuration. This error means that one thread or process has an exclusive lock on the database connection and another thread timed out waiting for the lock the be released.  
>
>Python’s SQLite wrapper has a default timeout value that determines how long the second thread is allowed to wait on the lock before it times out and raises the `OperationalError: database is locked` error.
[...]

PR #2276, opened 2024-12-31
Increase db timeout to 90s

### LINKED ISSUE #2252: Increase database timeout
# Why are we doing this?

There are [regular Sentry alerts](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d) for several Issues where requests fail with `OperationalError: database is locked` errors. The root cause is that our level of concurrency is currently too high for SQLite to reliably handle as currently configured. This Issue is one possible approach to improving the situation. Refer to the parent issue #2251 for more detail.

There are fewer instances of `OperationalError: database is locked` errors, [as monitored through Sentry](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d).

# What are we doing?

Increasing the database timeout. We already increased the timeout to 45 seconds in `opencodelists/settings.py`. Try increasing it further to 60 or 90 seconds and monitor how this improves the situation through [Sentry](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d). Refer to the parent ticket #2251 and [the Django docs](https://docs.djangoproject.com/en/4.2/ref/databases/#database-is-locked-errors) for more information on why this might help.

Jongmassey:
closes #2252

### DIFF
--- a/opencodelists/settings.py
+++ b/opencodelists/settings.py
@@ -177,7 +177,7 @@ DATABASES = {
     "default": dj_database_url.parse(
         os.environ.get("DATABASE_URL", default="sqlite:///db.sqlite3")
     ),
-    "OPTIONS": {"timeout": 45},
+    "OPTIONS": {"timeout": 90},
 }
 
 DATABASE_DIR = Path(os.environ.get("DATABASE_DIR", default=BASE_DIR))


bloodearnest:
Do we know what the current HTTP request timeouts are?

The defaults for nginx and gunicorn are 60s and 30s respectively.

Naively, I would expect increasing the sqlite timeout to more that that to not solve the problem for the user.

Additionally, i suspect the lack of IMMEDIATE_MODE, which means many queries will not wait for the timeout anyway, means this won't have the desired effect?

mikerkelly:
> Do we know what the current HTTP request timeouts are?

Good point. I think we're okay as they're higher than the 90s used in this PR.

nginx:
```dokku3:~$ dokku nginx:show-config opencodelists | grep timeout
  keepalive_timeout   70;
    proxy_read_timeout 600;
```

I think `proxy_read_timeout` is the relevant one for the slow requests we're concerned about as they are reverse-proxied on to gunicorn. Static files use the default of 60s. I don't think the keep alive timeout is relevant here.

[gunicorn](https://github.com/opensafely-core/opencodelists/blob/6161c99783827727a27c27b3e2ac779bb2bfd52c/deploy/gunicorn/conf.py#L12):
```
timeout = 120
```

> Additionally, i suspect the lack of IMMEDIATE_MODE, which means many queries will not wait for the timeout anyway, means this won't have the desired effect?

Quite possibly, I don't think we've properly characterised what the issue(s) at play are. As you brought up in #2251 changing SQLite configuration may be a better fix but I don't think we know that for sure yet. That's #2268 but it hasn't been triaged yet as it was raised just before Christmas.

This PR is very cheap and _may_ help, in some cases, depending what the issue actually is.

If the issue gets fully resolved we might consider reducing this timeout again.

mikerkelly:
As a point of comparison for later monitoring, there were about 175 errors [regular Sentry alerts](https://ebm-datalab.sentry.io/issues/?alert_rule_id=15682573&alert_type=issue&notification_uuid=a390aab4-1a0c-42b5-8c74-fccec8dfc608&project=5248013&query=is%3Aunresolved%20%22database%20is%20locked%22&referrer=issue-list&statsPeriod=90d) for several Issues where requests fail with `OperationalError: database is locked` errors in October-December 2024.

mikerkelly (approved):

PR #2288, opened 2025-01-07
Separate deploy workflow from main CI tests

### LINKED ISSUE #2260: Improve the deploy workflow concurrency
The deploy has the same lack of correct concurrency lock, that [job-server's CI](https://github.com/opensafely-core/job-server/commit/dc1bc43778f9ab74729abf6a3b9019cfc7162f45) did. It's locking on the deploy step, not the whole workflow, which means two deploys at around the same time could run out of order, if the second run reaches deploy before the first run.

StevenMaude:
Fixes #2260.

This prevents the case where two CI workflow runs can be triggered in quick succession, and the first triggered run may actually, on occasion, end up as the second deployment.

See opensafely-core/job-server@dc1bc43778f9ab74729abf6a3b9019cfc7162f45

There is more refactoring of the workflows that could be done here, but leaving this as a single, simpler change.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,79 @@
+---
+name: Deploy
+
+env:
+    IMAGE_NAME: opencodelists
+    PUBLIC_IMAGE_NAME: ghcr.io/opensafely-core/opencodelists
+    REGISTRY: ghcr.io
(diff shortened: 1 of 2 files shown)


StevenMaude:
I tested this in a fork, and the deploy works up to the point where credentials are needed, which aren't available in the fork.

tomodwyer (approved):

Issue #2249, opened 2024-12-16
Use TypeScript for the Codelist Builder app


tomodwyer:
The React app for the Codelist Builder can be converted to TypeScript to help developers understand the types passed around the app. This is especially useful as the app uses class based components, which are no longer the recommended way to build React apps.

This app also uses PropTypes which are [deprecated in React 19](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops), so TypeScript can replace the functionality of PropTypes.

The outcome of this issue should be zero change in functionality, but better types and comments for developers.

PR #2296, opened 2025-01-14
Move modal component and logic to TreeRow component

tomodwyer:
The MoreInfo Modal is tied to each row in the Tree, yet the component lives in the top-level class component. This PR moves the logic and component in to the tree row, where the button lives. This means the button component can be removed, the logic can be simplified, and the CodelistBuilder becomes less complex to debug.

Linked to #2249


Jongmassey (approved):
running through creating, saving, discarding seems to work fine 👍

PR #2306, opened 2025-01-16
Update the styling of the "Created:" tooltip

tomodwyer:
Update the styling of the whole version section including:

- Enabling the Bootstrap tooltip
- Using an abbreviation tag for the tooltip
- Improving the styles to remove unnecessary HTML tags
- Show which version is currently active


Jongmassey (approved):
LGTM - your change to an ordered list made me double-check the [ordering of the versions](https://github.com/opensafely-core/opencodelists/blob/086090c3f14323895e598c0bb7b5fd93295c7179/codelists/models.py#L176) and all is well, less potential for confusion now with using `created_at` as well as it will always be id-sequenced

PR #2320, opened 2025-01-23
Enhance otel spans with an identifier for logged in users

### LINKED ISSUE #2255: Enhance OTel spans with an identifier for logged-in users
# Why are we doing this?

[Honeycomb](https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/board/4ngq7kuV7z/OpenCodelists-Django-app) allows us to view OTel data exposed by the site. This allows us to monitor performance. This will be enhanced if we can view traces and spans by who is making them. This may help us identify users facing particular issues, for example.

- OpenCodelists Honeycomb queries can use logged in user as a search term.
- We have better developer documentation of how we OTel, Honeycomb, and how to develop against them.

---

[Defining delivery tasks guidance](https://docs.google.com/document/d/1F7yBNYnMoCW3liWVvjPXFHnrnCU4qeAOhlKBVD1-8lg/edit?tab=t.0#heading=h.qkl5vvsdkx98)

KatieB5:
Fixes #2255 

Add username of opencodelist users to OTel spans.

**Why was this work done?**
We use [OpenTelemetry](https://opentelemetry.io/docs/what-is-opentelemetry/) to instrument opencodelists such that when users make requests, we can capture data about these requests in [traces and spans](https://opentelemetry.io/docs/concepts/signals/traces/#spans) emitted by our code. We send this data to [Honeycomb](https://docs.honeycomb.io/get-started/start-building/application/traces/) in order to view and interrogate it, helping us to understand the internal state of our system. Previously, the traces and spans emitted from opencodelists did not include information to identify, or query by, logged-in users. Adding the username of opencodelist users as a span attribute allows us to do this, and potentially help users facing particular issues.

**What has been done?**
Initially, attempts were made to use an [environment variable](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/django/django.html#request-attributes) provided by OTel which extracts attributes from Django’s request object and adds them as span attributes. However, this did not work: spans created when a request is made do not yet have access to the ‘user’ attribute on Django’s request object, as Django’s [authentication middleware](https://docs.djangoproject.com/en/5.1/topics/auth/default/#authentication-in-web-requests) (which adds the user attribute to the request object) hasn’t yet run. Thus a different approach was required. 

OTel provides [request and response hooks](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/django/django.html#request-and-response-hooks), used with [Django instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/django/django.html#id1) to instrument the app. Request and response hooks enable access to the Django request object immediately after span creation and right before a span finishes, respectively. The current approach uses a response_hook() to ensure the authentication middleware has added the 'user' attribute to the Django request object, before attempting to add this to a span. To facilitate testing of the response_hook, and to more closely mirror how tracing is set up in other OpenSafely applications e.g. airlock, this functionality was moved to services/tracing.py. 

Additionally, OTel instrumentation requires specific environment variables to be set, so dotenv-sample and DEPLOY.md have been updated to reflect this. 

**Has this been tested?**
Yes:
1. Pytest fixtures set up a global tracing infrastructure and instrumentation for the purposes of testing, and two ‘end-to-end’ tests then exercise the prod response_hook() functionality.
2. Honeycomb has a development environment; local dev OTel data was sent to this environment by generating an API key in the Honeycomb dev env (note: only Team Organisers have this authority), adding 'user' as a tracing field on the dataset definition, updating the OTEL_EXPORTER_OTLP_HEADERS env var with the API key, spinning up a gunicorn server locally and making requests in local opencodelists (see Honeycomb dashboard images below). NOTE: adding 'user' as a tracing field on the dataset definition will need to be done in the Honeycomb production environment once the changes in this PR are deployed.

Logged in user (see bottom right of image for ‘user’ field):
Screenshot from 2025-01-23 15-25-20

Non-logged in user:
Screenshot from 2025-01-23 15-18-31

Query by logged in user:
Screenshot from 2025-01-23 16-46-20

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,6 +1,14 @@
 import pytest
 from django.db import connections
 from django.test import TestCase
+from opentelemetry import trace
+from opentelemetry.instrumentation.django import DjangoInstrumentor
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+
+from services.tracing import response_hook
 
 
 # TestCase.databases sets the databases that tests have access to; this allows all tests
@@ -71,3 +79,57 @@ def reset_connections():
     databases_added_during_tests = set(connections.databases.keys()) - initial_databases
     for db in databases_added_during_tests:
         del connections.databases[db]
+
+
+################################################################################################################
+# Setup global tracing infrastructure for testing purposes.
+
+# The test setup is designed to be as close to the prod environment as possible,
+# with only necessary modifications specific to testing.
+
+# Create a resource for testing purposes with a unique 'service.name' value to
+# minimise risk of mixing test and prod telemetry data.
+# https://opentelemetry-python.readthedocs.io/en/latest/sdk/resources.html
+resource = Resource.create(attributes={"service.name": "opencodelists-test"})
+
+# Configure a global default OTel TracerProvider for our tests:
+# The TracerProvider enables access to Tracers, Tracers create Spans.
+# TracerProvider configuration is done once, globally because the OTel Tracing API
+# should be used to configure and set/register a global default TracerProvider once,
+# at the start of a process (in this case, the pytest session; in prod, a gunicorn
+# worker process)). This means that during the pytest test session any test
+# that uses Django's test client will generate telemetry data (spans)
+# that we can assert against.
+# https://opentelemetry.io/docs/specs/otel/trace/api/#tracerprovider
+provider = TracerProvider(resource=resource)
+trace.set_tracer_provider(provider)
+
+# In prod, we use the OTLPSpanExporter which exports our telemetry data to Honeycomb.
+# In testing, we use InMemorySpanExporter() to collect spans locally, and assert
+# against them during tests.
+testing_span_exporter = InMemorySpanExporter()
+
+# In prod, we use the BatchSpanProcessor as this is asynchronous, and queues
+# and retries sending telemetry. In testing, we use SimpleSpanProcessor, which is
+# synchronous and easy to inspect the output of within a test.
+provider.add_span_processor(SimpleSpanProcessor(testing_span_exporter))
+
+# Mirror production instrumentation:
+# Use DjangoInstumentor() to instrument OpenCodelists test data, with prod response_hook().
+DjangoInstrumentor().instrument(response_hook=response_hook)
+################################################################################################################
+
+
+@pytest.fixture(autouse=True)
+def clear_testing_span_exporter():
+    """Clear OTel spans from the test exporter after each test"""
+    yield
+    testing_span_exporter.clear()
+
+
+@pytest.fixture()
+def span_exporter():
+    """Provide an empty test exporter to export OTel spans for each test.
+    Interrogate it with, for example get_finished_spans()."""
+    assert not testing_span_exporter.get_finished_spans()
+    yield testing_span_exporter
--- a/conftest.py
+++ b/conftest.py
@@ -94,7 +94,7 @@ resource = Resource.create(attributes={"service.name": "opencodelists-test"})
 
 # Configure a global default OTel TracerProvider for our tests:
 # The TracerProvider enables access to Tracers, Tracers create Spans.
-# TracerProvider configuration is done once, globally because the OTel Tracing API
+# TracerProvider configuration is done once, globally, because the OTel Tracing API
 # should be used to configure and set/register a global default TracerProvider once,
 # at the start of a process (in this case, the pytest session; in prod, a gunicorn
 # worker process)). This means that during the pytest test session any test
@@ -106,7 +106,7 @@ trace.set_tracer_provider(provider)
 
 # In prod, we use the OTLPSpanExporter which exports our telemetry data to Honeycomb.
 # In testing, we use InMemorySpanExporter() to collect spans locally, and assert
-# against them during tests.
+# against them during tests using the span_exporter fixture.
 testing_span_exporter = InMemorySpanExporter()
 
 # In prod, we use the BatchSpanProcessor as this is asynchronous, and queues
@@ -129,7 +129,7 @@ def clear_testing_span_exporter():
 
 @pytest.fixture()
 def span_exporter():
-    """Provide an empty test exporter to export OTel spans for each test.
-    Interrogate it with, for example get_finished_spans()."""
+    """Provide access to the global span exporter for interrogation.
+    Interrogate it with, for example, get_finished_spans()."""
     assert not testing_span_exporter.get_finished_spans()
     yield testing_span_exporter
--- a/services/tracing.py
+++ b/services/tracing.py
@@ -1,0 +1,49 @@
+"""Functions to configure OTel tracing and instrumentation
+We use OpenTelemetry (OTel) to instrument the app and capture telemetry
+data about requests in traces and spans emitted by our code. We send
+this data to Honeycomb to view and interrogate it, helping us to
+understand the internal state of our system and assist our users.
+
+In production, the functions below are invoked in
+deploy/gunicorn/conf.py. OTel requires environment variables to be set;
+these are described in dotenv-sample and DEPLOY.md."""
+
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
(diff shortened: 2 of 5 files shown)


dotenv-sample
```diff
@@ -4,7 +4,7 @@ BASE_URLS=http://localhost:7000,http://127.0.0.1:7000
 SECRET_KEY=dummy-secret
 DJANGO_DEBUG_TOOLBAR=False
 SENTRY_DSN=
-OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io"
+OTEL_EXPORTER_OTLP_ENDPOINT="https://api.honeycomb.io/v1/traces"
```
Jongmassey, line 7: Any idea what changed about the traces we send which means that the endpoint had to change?

Jongmassey (approved):

Jongmassey:
NB this will require updating the dokku config after deployment to point at the new endpoint

conftest.py
```diff
@@ -1,6 +1,16 @@
+# import os
+
```
mikerkelly, line 2: Stale debugging code?

conftest.py
```diff
@@ -71,3 +81,34 @@ def reset_connections():
...
+
+# Set up global tracer provider which can be set only once.
+@pytest.fixture(scope="module")
+def provider():
+    resource = Resource.create(attributes={"service.name": "opencodelists"})
+    return TracerProvider(resource=resource)
+
+
+@pytest.fixture(scope="function")
```
mikerkelly, line 93: This is the default so (our) code doesn't tend to specify it. I don't think that means it needs to change, though.

services/README.md
```diff
@@ -0,0 +1,14 @@
+# Services
```
mikerkelly, line 1: I would prefer this info being in the `services/tracing.py` module itself, so it is closer to where it is used. Weakly-held opinion.

In general I would expect a directory-level readme to either give a very brief manifest of the directory contents with more detail in the files themselves, or relate to wider topics that perhaps aren't best covered by file-level docs such as an overview of the purpose of the directory, its interaction with other components, or the relationship between files.

services/tracing.py
```diff
@@ -0,0 +1,27 @@
...
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+
+def set_up_tracing():
+    resource = Resource.create(attributes={"service.name": "opencodelists"})
+
+    trace.set_tracer_provider(TracerProvider(resource=resource))
+    span_processor = BatchSpanProcessor(
+        OTLPSpanExporter(endpoint="https://api.honeycomb.io/v1/traces")
```
mikerkelly, line 13: This is hardcoded, should it be using the environment variable `OTEL_EXPORTER_OLTP_ENDPOINT`?

services/README.md
KatieB5: Thanks Mike! Have updated the documentation as discussed.

services/tracing.py
KatieB5: Thanks Mike, this is now resolved :)

conftest.py
```diff
@@ -71,3 +81,34 @@ def reset_connections():
...
+    # In production, we use the BatchSpanProcessor as this is
+    # asynchronous, and queues and retries sending telemetry.
+    # We also use the OTLPSpanExporter which exports our
+    # telemetry data to Honeycomb.
+
+    # In testing, we use SimpleSpanProcessor, which is
+    # synchronous and easy to inspect the output of within a test.
+    # We use InMemorySpanExporter() to collect spans locally,
+    # and assert against them during tests.
+    test_exporter = InMemorySpanExporter()
```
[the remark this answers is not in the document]
KatieB5: Thanks very much for this Simon, your explanation and suggestions were really helpful. Mike and I discussed this further and I've reworked the testing set up accordingly.

dotenv-sample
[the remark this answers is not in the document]
KatieB5: Thanks Simon! This has now been updated.

services/tracing.py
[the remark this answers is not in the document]
KatieB5: Thanks Simon! This has now been updated.

dotenv-sample
KatieB5: Thank you very much both! This has now been updated.

conftest.py
```diff
@@ -71,3 +79,57 @@ def reset_connections():
...
+# with only necessary modifications specific to testing.
+
+# Create a resource for testing purposes with a unique 'service.name' value to
+# minimise risk of mixing test and prod telemetry data.
+# https://opentelemetry-python.readthedocs.io/en/latest/sdk/resources.html
+resource = Resource.create(attributes={"service.name": "opencodelists-test"})
+
+# Configure a global default OTel TracerProvider for our tests:
+# The TracerProvider enables access to Tracers, Tracers create Spans.
+# TracerProvider configuration is done once, globally because the OTel Tracing API
```
mikerkelly, line 97:
```suggestion
# TracerProvider configuration is done once, globally, because the OTel Tracing API
```

conftest.py
```diff
@@ -71,3 +79,57 @@ def reset_connections():
...
+# worker process)). This means that during the pytest test session any test
+# that uses Django's test client will generate telemetry data (spans)
+# that we can assert against.
+# https://opentelemetry.io/docs/specs/otel/trace/api/#tracerprovider
+provider = TracerProvider(resource=resource)
+trace.set_tracer_provider(provider)
+
+# In prod, we use the OTLPSpanExporter which exports our telemetry data to Honeycomb.
+# In testing, we use InMemorySpanExporter() to collect spans locally, and assert
+# against them during tests.
```
mikerkelly, line 109:
```suggestion
# against them during tests using the span_exporter fixture.
```

Pointer to how to actually do this, which has further guidance.

conftest.py
```diff
@@ -71,3 +79,57 @@ def reset_connections():
...
+@pytest.fixture(autouse=True)
+def clear_testing_span_exporter():
+    """Clear OTel spans from the test exporter after each test"""
+    yield
+    testing_span_exporter.clear()
+
+
+@pytest.fixture()
+def span_exporter():
+    """Provide an empty test exporter to export OTel spans for each test.
```
mikerkelly, line 132:
```suggestion
    """Provide access to the span exporter for interrogation.
```

This function isn't actually the one that does the clearing, so it shouldn't say it does that.

conftest.py
```diff
@@ -71,3 +79,57 @@ def reset_connections():
...
+    Interrogate it with, for example get_finished_spans()."""
```
mikerkelly, line 133:
```suggestion
    Interrogate it with, for example, get_finished_spans()."""
```

services/tracing.py
```diff
@@ -0,0 +1,49 @@
...
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+
+def set_up_tracing():
+    """Setup tracing infrastructure for production.
+    The resource represents the entity producing telemetry data.
```
mikerkelly, line 20:
```suggestion
    """Setup tracing infrastructure for production.
    
    The resource represents the entity producing telemetry data.
```
Clearer to have an empty line between the summary line and the further detail.

services/tracing.py
```diff
@@ -0,0 +1,49 @@
+"""Functions to configure OTel tracing and instrumentation
+We use OpenTelemetry (OTel) to instrument the app and capture telemetry
```
mikerkelly, line 2:
```suggestion
"""Functions to configure OTel tracing and instrumentation.

We use OpenTelemetry (OTel) to instrument the app and capture telemetry
```
As below.

services/tracing.py
```diff
@@ -0,0 +1,49 @@
...
+    resource = Resource.create(attributes={"service.name": "opencodelists"})
+
+    trace.set_tracer_provider(TracerProvider(resource=resource))
+    span_processor = BatchSpanProcessor(OTLPSpanExporter())
+    trace.get_tracer_provider().add_span_processor(span_processor)
+
+
+def response_hook(span, request, response):
+    """Add user data to spans so that we can query this in Honeycomb.
+    OTel's request_hook and response_hook run at span creation for a request
```
mikerkelly, line 42:
```suggestion
    """Add user data to spans so that we can query this in Honeycomb.
    
    OTel's request_hook and response_hook run at span creation for a request
```
As above.

services/tracing.py
mikerkelly: Also, getting increasingly pedantic, we should be consistent about whether those first lines end with a full stop. (For me, it's easier to read if they do.)

PR #2331, opened 2025-02-04
Replace use of pytest-subtests with `parametrize`

### LINKED ISSUE #2206: Remove use of pytest-subtests if possible
We only use it in a few tests.

It may be possible to drop this dependency and use `pytest.mark.parametrize` instead for these tests. (That said, I only looked in passing, while reviewing #2177.) The benefit is one fewer dependency to update, and for developers to understand.

StevenMaude:
Fixes #2206.

### DIFF
--- a/codelists/tests/test_codeset.py
+++ b/codelists/tests/test_codeset.py
@@ -19,11 +19,11 @@ from .helpers import build_hierarchy, hierarchies
 # g   h   i   j
 
 
-def id_fn(value):
+def _get_description(value):
     return value["description"]
 
 
-@pytest.mark.parametrize("example", examples, ids=id_fn)
+@pytest.mark.parametrize("example", examples, ids=_get_description)
 def test_from_codes(example):
     hierarchy = build_hierarchy()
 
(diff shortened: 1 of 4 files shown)


codelists/tests/test_codeset.py
```diff
@@ -18,56 +19,56 @@
 # g   h   i   j
 
 
-def test_from_codes(subtests):
+def id_fn(value):
+    return value["description"]
+
+
+@pytest.mark.parametrize("example", examples, ids=id_fn)
```
mikerkelly, line 26: This is only used in this module as a data structure indexer. Maybe it could be an anonymous function (even though you use it a few... times), or a _pseudoprivate function. If the function continues to exist I'm not sure how to interpret the function or parameter name, can they be improved?

Maybe either:

```suggestion
@pytest.mark.parametrize("example", examples, ids=lambda example: example["description"]))
```

or

```suggestion
def _get_description(example):
    return example["description"]


@pytest.mark.parametrize("example", examples, ids=_get_description)
```

mikerkelly (approved):
Thanks for taking the time to spot and action this opportunity to reduce dependencies, LGTM, one minor comment.

codelists/tests/test_codeset.py
StevenMaude: I actually think I used a more descriptive function when I started, then copied the pytest docs for `id_fn`. (That [documentation actually uses `idfn`](https://github.com/pytest-dev/pytest/blob/6530f42e976647f0a78ce5aa7418017323616b3b/doc/en/example/parametrize.rst#L120-L123).)

So I've gone for the named function over a `lambda`.

Issue #177, opened 2020-09-08
Medicines Definition Builder: Snomed hierachy link to dm+d


brianmackenna:
In #176 we discuss overall plan. Although [NHS dm+d](https://ebmdatalab.net/what-is-the-dmd-the-nhs-dictionary-of-medicines-and-devices/) exists as a standalone comprehensive reference source it actually is the local UK extension of SnoMed. We can utilise the hierarchies of snomed to support categorisation and building codelists. As we know Snomed is quite extensive and complex, some of the categories don't actually make sense to me however using the [East Kent browser](https://dmd.ekhuft.nhs.uk/) and the Australian Shrimp browser it looks like it will be very useful.

Using Atorvastatin 20mg chewable tabs and taking the ancestors displayed on EK browser here are some examples and uses of the SnoMed hierarchy. 

Snomed CT ancestor | Brian’s expectation
-- | --
Enzyme inhibitor-containing product | Unexpected but useful. Enzyme inhibition is not a black and white issue - this classification is more suited to decision support type content as opposed to blac&white ontology.
Nitrogen and nitrogen compound-containing product | I had no clue atorvastatin has nitrogen in it and not a clue of relevance clinically. It is accurate! There will be chemists, different flavours of nerd pharmacists and pharmacologists who might make use of searching or studies where the principle chemical components are of interest as opposed to mechanism of action/clinical use.
HMG-CoA reductase inhibitor-containing product | Expected - this is technical term for statins, named after mechanism of action.
Azole-containing product | Similar to N but in this case I can’t see an azole in the chemical structure. (note to check accuracy when have time but I imagine it is fine)
Atorvastatin | Expected

inglesp:
The termbrowser knows that 19722511000001105 (Atorvastatin 20mg chewable tablets sugar free) is a 108600003 (Product containing atorvastatin) because 108600003 is the dm+d code for the [Atorvastatin VTM](https://openprescribing.net/dmd/vtm/108600003/).

@brianmackenna have you done any work to see whether all dm+d VTMs are in SNOMED?

brianmackenna:
No

brianmackenna:
These SNOMED _[Drug Extension User Support Group](https://confluence.ihtsdotools.org/display/USRG/Drug+Extension+User+Support+Group)_ has some useful documents and presentations that might be useful for when we get to this. 

It would also be good to consider #525 if we do a comprehensive job on this issue.

Jongmassey:
[...]

PR #2396, opened 2025-04-02
Spike dmd codelist builder

### LINKED ISSUE #2388: Spike: Basic dm+d support in codelist builder (draft)
# Why are we doing this?

Oft requested
Codelist rot

Be able to make a dm+d codelist in builder

# What are we doing?

Define required parts of coding system interface (hierarchy, code types, other required methods) for builder

Jongmassey:
Demonstration of what a first fix of #2388 could look like.

Currently this shows medicinal products in a hierarchy of VTM->VMP->AMP, but searches across Ingredient, VTM, VMP, AMP (by code and name). The VTM-Ing ingredient relationship is of a many-to-many form such that that neither hierarchy of Ing->VTM->VMP->AMP nor VTM->Ing->VMP->AMP make sense. Additionally, there is no direct link between VTM and Ing, only via VMP and VPI.

For example, VTM 776002003 ("Fluorouracil") is related to VMPs that have a single ingredient of one of two forms of Fluorouracil: 387172005 ("Fluorouracil") or 441807008 ("Fluorouracil sodium"), which implies that the VTM is the "parent" object of these two active forms of this drug.

However, Ing 372687004 ("Amoxicillin") is present in VMPs that belong to multiple VTMs: 774586009 ("Amoxicillin") and 774587000 ("Co-Amoxiclav") which implies that the Ing is a parent of both those VTMs.

_[...and these VTMs also have relationships to VMPs with other amoxicillin salts, both as a single ingredient and in combination with clavulanic acid which further reveals that this is not resolvable to a single hierarchy...]_

Thus, searching on Ing but omitting it from the hierarchy is a reasonable compromise - the downside being that a user trying to make a codelist of a very specific ingredient form rather than the general would have to manually select the codes for inclusion and exclusion. I feel like this is a rarer case than users wanting to differentiate between different therapeutic combinations that may contain a particular ingredient (e.g. only non-potentiated Amoxicillins,  paracetamol in combination with codeine but not codeine-only or paracetamol-only products).

Strength and Route and two concepts that are not included in this spike (raised by @rw251 and @sebbacon  I believe) but are useful classification axes for medication codelist building ("all nasal sprays", "low-dose statins"). We would have to do some analysis to see if these would be able to fit in a simple hierarchy. My instinct tells me Route should be OK - I'm not aware of any products that have multiple routes of administration; but each ingredient in a VMP has its own strength so this may be more complicated. It may be that there aren't many strength options for a VMPs of a particular Route and VTM so the potential amount of user-clicking to achieve their desired inclusions and exclusions mightn't be _too_ bad.

### DIFF
--- a/codelists/models.py
+++ b/codelists/models.py
@@ -495,7 +495,7 @@ class CodelistVersion(models.Model):
             return self._new_style_codes()
 
     def _old_style_codes(self):
-        if self.coding_system_id in ["bnf", "ctv3", "icd10", "snomedct"]:
+        if self.coding_system_id in ["bnf", "ctv3", "icd10", "snomedct", "dmd"]:
             headers, *rows = self.table
             headers = [header.lower().strip() for header in headers]
             # (non-dmd) old style codelists are now required to contain a column named "code"
--- a/codelists/tests/views/test_version_diff.py
+++ b/codelists/tests/views/test_version_diff.py
@@ -61,12 +61,17 @@ def test_get_dmd_diff_codes_differ(
     assert rsp.context["lhs_only_codes"] == {"10514511000001106"}
     assert rsp.context["rhs_only_codes"] == {"123"}
     assert rsp.context["rhs_only_summary"] == [
-        {"code": "123", "term": "[Unknown] Test refill (VMP)"}
+        {
+            "code": "123",
+            "term": "[Unknown] Test refill (VMP)",
+            "descendants": [],
+        }
     ]
     assert rsp.context["common_summary"] == [
         {
             "code": "10525011000001107",
             "term": "Adrenaline (base) 220micrograms/dose inhaler refill (VMP)",
+            "descendants": [],
         }
     ]
 
@@ -119,12 +124,17 @@ def test_get_dmd_diff_alternative_column_names(
     assert rsp.context["lhs_only_codes"] == {"10514511000001106"}
     assert rsp.context["rhs_only_codes"] == {"123"}
     assert rsp.context["rhs_only_summary"] == [
-        {"code": "123", "term": expected_unknown_term}
+        {
+            "code": "123",
+            "term": expected_unknown_term,
+            "descendants": [],
+        }
     ]
     assert rsp.context["common_summary"] == [
         {
             "code": "10525011000001107",
             "term": "Adrenaline (base) 220micrograms/dose inhaler refill (VMP)",
+            "descendants": [],
         }
     ]
 
--- a/coding_systems/dmd/coding_system.py
+++ b/coding_systems/dmd/coding_system.py
@@ -1,12 +1,10 @@
 from django.db.models import Q
 
-from django.db.models import Q
-
-from ..base.coding_system_base import BaseCodingSystem
+from ..base.coding_system_base import BuilderCompatibleCodingSystem
 from .models import AMP, AMPP, VMP, VMPP, VTM, Ing
 
 
-class CodingSystem(BaseCodingSystem):
+class CodingSystem(BuilderCompatibleCodingSystem):
     id = "dmd"
     name = "Dictionary of Medicines and Devices"
     short_name = "dm+d"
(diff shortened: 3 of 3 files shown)


brianmackenna:
>Strength and Route and two concepts that are not included in this spike (raised by @rw251 and @sebbacon I believe) but are useful classification axes for medication codelist building ("all nasal sprays", "low-dose statins").

I'm not convinced that those classification axes are very important for an average user.

My gut feel is that this is the most important thing we need for building https://github.com/opensafely-core/opencodelists/issues/177 natively in snomed/dm+d. The lack of clinical classifiers within dm+d itself means a standalone dm+d builder will be limited and probably only useful for more experienced users.

Jongmassey:
One class of entity in the Dictionary of Medicines and **Devices** is a device, which doesn't have a VTM on account of not having any therapeutic ingredients.

I've not been able to find any explicit "device" [codelists](https://github.com/opensafely/codelist-development/issues?q=is%3Aissue%20device%20) in the codelist development repo.

There are devices in the BNF, in chapters 16 and above (although I think there's no 16 or 17). 
Looking to see if we have any BNF codelists already that go above 15 we do have at least some ([e.g.](https://www.opencodelists.org/codelist/nhs-devon/south-and-west-devon-red-drugs/642c9d0f/#tree)).

Jongmassey:
aha, that's not quite right! There is a "pancreatin" VTM but it's not linked to those products for some reason. 

It displays when the "unresolved" link is clicked but not otherwise. That's some odd behaviour we will have to address in future

image

Jongmassey:
Pancreatin was a (un)fortunate example to pick at random and reveals some of the new considerations and challenges that come with making dm+d codelists (which are in many ways akin to those when making SNOMED-CT lists).

Pancreatin is a "[mixture of several digestive enzymes produced by the exocrine cells of the pancreas. It is composed of amylase, lipase, and protease](https://en.wikipedia.org/wiki/Pancreatic_enzymes_(medication)#:~:text=Pancreatin%20is%20a%20mixture%20of,pancreatectomy%2C%20pancreatitis%20and%20cystic%20fibrosis.)".

It is included in the dm+d as a single ingredient in its own right:

```
dmd_2025-420_20250421.sqlite3> select * from dmd_ing where lower(nm)='pancreatin'
+-----------+------------+------------------+---------+------------+
| isid      | isiddt     | isidprev         | invalid | nm         |
+-----------+------------+------------------+---------+------------+
| 387049005 | 2005-08-02 | 3572811000001107 | 0       | Pancreatin |
+-----------+------------+------------------+---------+------------+
```
and as a VTM :

```
dmd_2025-420_20250421.sqlite3> select * from dmd_vtm where lower(nm)='pancreatin'
+-----------+---------+------------+----------+-----------+------------+
| vtmid     | invalid | nm         | abbrevnm | vtmidprev | vtmiddt    |
+-----------+---------+------------+----------+-----------+------------+
| 777054003 | 1       | Pancreatin | <null>   | 12236006  | 2023-02-06 |
+-----------+---------+------------+----------+-----------+------------+
```

albeit with no corresponding VMPs:

```
dmd_2025-420_20250421.sqlite3> select * from dmd_vmp where vtmid = 777054003
0 rows in set
```

Using the current term search strategy of this spike (ingredients and vtms by name, any vmps related to those plus by name, any amps related to those plus by name) it returns `Generic Creon 750mg gastro-resistant granules sachets (VMP) (4551411000001100)` by virtue of it having a relationship to the above pancreatin ingredient, and `Generic Creon Micro Pancreatin 60.12mg gastro-resistant granules (VMP)(8062111000001107)` by virtue of it having pancreatin in the name. So far so good (notwithstanding the slight VTM weirdness).


I decided to compare this to a Pseudo BNF codelist converted to dm+d to check the workings.

The Pseudo BNF codelist was built by including 1.9.4 and its descendents:
image


When converted to dm+d there were some surprising differences:
image


The VMPs in the converted list that are not present in the "native dm+d" list do not have an ingredient of "pancreatin" but instead have amylase, lipase, and protease seperately. A user developing this codelist would have to know to check this sort of analogue of synonymy whereby ingredients may exist as a combined ingredient under a single name or as individual ingredients.

What's also interesting is that a bunch of the AMPs are excluded - because the mapping file is out of date with respect to the main dm+d. This is a great example of how using mapping/converting codelists contributes to codelist rot due to the delays in publication of the mapping file! This is particularly concerning in this case because some of the VMPs in question show up [on OpenPrescribing](https://openprescribing.net/dmd/vmp/8062111000001107/) as "Never Valid To Prescribe As A VMP" which seems to suggest that all prescribing for these will be as AMPs which change more frequently in the dm+d and so are more susceptible to codelist rot.

PR #2415, opened 2025-04-08
Added debug config for vscode

rw251:
Added some debugging shortcuts for vscode users. One click setup for:
1. Launching opencodelists in debug mode
2. Debugging the current test file
3. Debugging the current PlayWright test file (in anticipation of PlayWright tests)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,31 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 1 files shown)


lucyb (approved):
In the past, we've had various discussions about including editor specific config in repos and were reluctant to do so. However, vscode is currently used by a high % of the team and this config isn't something user specific (like key-bindings) but project specific. So I think this change is a perfectly reasonable decision.

Issue #2166, opened 2024-11-27
Codelist builder search returns 500 error with unexpected characters


lucyb:
When constructing a search that contains a `*` or a `%`, or ends with a `-`, then the user receives a 500 error.

The expected behaviour should be to show the banner saying `There are no results for "..."`. This appears for other search terms that are not found in the database.

```
IntegrityError at /builder/532297c2/search/
UNIQUE constraint failed: codelists_search.version_id, codelists_search.slug
```

```
2024-11-27T14:42:27.918036Z [error    ] Internal Server Error: /builder/532297c2/search/ [django.request] ip=127.0.0.1 request_id=7eb67c68-f8ff-44a8-ab02-23fdcece696b user_id=lucyb
Traceback (most recent call last):
  File "/home/lucy/datalab/ebmdatalab/opencodelists/.venv/lib/python3.12/site-packages/django/db/models/query.py", line 916, in get_or_create
    return self.get(**kwargs), False
           ^^^^^^^^^^^^^^^^^^
  File "/home/lucy/datalab/ebmdatalab/opencodelists/.venv/lib/python3.12/site-packages/django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(
codelists.models.Search.DoesNotExist: Search matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/lucy/datalab/ebmdatalab/opencodelists/.venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lucy/datalab/ebmdatalab/opencodelists/.venv/lib/python3.12/site-packages/django/db/backends/sqlite3/base.py", line 328, in execute
    return super().execute(query, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sqlean.dbapi2.IntegrityError: UNIQUE constraint failed: codelists_search.version_id, codelists_search.slug

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/lucy/datalab/ebmdatalab/opencodelists/.venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2416, opened 2025-04-08
Fix bug with non-alphanumeric character searches

rw251:
Fixes this ticket (#2166 ) which describes a bug when searching for possible wildcards such as `*` and `%`. 

The actual issue is not with wildcards per se, but with the following flow:
- add a search term e.g. `aspirin`
- add a second search term which is different, but with the same slug as per the `slugify()` django function. e.g. `aspirin*`
- this then causes the `UNIQUE` constraint on the slug field to throw an error
- NB - you could add these in reverse order, in which case the system would be happy with `aspirin*`, but then throw a 500 error on the subsequent search for `aspirin

I haven't added any client-side checks because the back-end allows searching for these non-alphanumeric characters.


StevenMaude:
I tried this branch locally.

The following case works fine (which is added to the unit tests):

1. Create a new BNF draft. (Also works with say, SNOMED CT; there's nothing special about BNF.)
1. Search for the term `aspirin`: get results.
1. Search for the term `aspirin*`: get no error and the results are those of the `aspirin` search.

Searching the other way round has slightly quirky behaviour (which is possibly preferable to the error, but might warrant a bit of thinking about how to avoid confusion):

1. Create a new BNF draft. (Also works with say, SNOMED CT; there's nothing special about BNF.)
1. Search for the term `aspirin*`: get no results.
1. Search for the term `aspirin`: get no error, but no results; the results are those of the `aspirin*` search.

A user can fix this lack of results by deleting the `aspirin*` search, and searching again for `aspirin`. But that may not be immediately obvious.
 
If you do delete the broken `aspirin*` search, it says `Search for "aspirin" deleted` which is also not strictly what's happened.

***

This behaviour happens because for the entered search term `aspirin`, the `slug` is `aspirin`, but the `term` in the code is still `aspirin*` and [used in this line](https://github.com/opensafely-core/opencodelists/pull/2416/files#diff-8e27c63062d99d9976345797219999347958d3ba923fddd2c2ed6bd69aa95d3cR31). So the `aspirin*` search gets bound to the `aspirin` slug.

StevenMaude:
suggestion: It would be great to have helpful information like that in the opening message in this PR thread in the relevant commit message. It's well written and describes the problem.

(Keeping the context with the code saves any future developer working on this code from having to dig out and consult the PR thread from GitHub.)

rw251:
As per [this thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1744270179964489) this PR has been updated to:
- change the search url to use `search.id` instead of `search.slug`
- change the slug to either the exact search term (removing leading/trailing spaces and to lower case), or the `code:xxxx` string

builder/tests/test_views.py
```diff
@@ -40,10 +40,15 @@ def test_version_from_scratch(client, version_from_scratch):
 
 
 def test_search(client, draft_with_some_searches):
-    rsp = client.get(draft_with_some_searches.get_builder_search_url("arthritis"))
+    slug = "arthritis"
+    for search in draft_with_some_searches.searches.all():
+        if search.term == slug:
+            search_id = search.id
```
mikerkelly, line 46:
```suggestion
    search_id = draft_with_some_searches.searches.filter(term=slug).first().id
```

Think using a queryset filter rather than iteration is more idiomatic and marginally more readable.

builder/actions.py
```diff
@@ -13,13 +13,23 @@
...
+    Returns the search slug which is either the slugified search term,
+    or the code prefixed with "code:". Also validates that only one of
+    term and code is set.
+    """
     assert bool(term) != bool(code)
     if term is not None:
         slug = slugify(term)
     else:
         slug = f"code:{code}"
+    return slug
```
mikerkelly, line 27: My initial reaction was to say I slightly prefer this function being a `classmethod` of the `Search` model, as it's only relevant to that class. But I'm not sure how well that fits with the `actions.py` framework in this repo where any business logic that CRUDs instances is encapsulated in this module. Maybe it's worth a wider discussion as I'm not sure I've understood  the motivations behind approach. I added it the cabinet for discussion in planning next week if that works. Also fine to merge as-is.

mikerkelly (approved):
Thanks Richard, this is looking good. We discussed in [Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1744270179964489) and you tweaked the approach to now:

- Add the search instance's `id` to the URL along with the `slug` for human readability
- Remove the uniqueness constraint on `version`/`slug`.
- Update code referencing searches by `slug` to use `id` instead.

This keeps the human-readable slug in the URL while avoiding uniqueness violations when two search terms have identical slugs (differing only by whitespace or non-alphanumeric characters).

Would it be valuable to write a test for creating a search without non-alphanumeric characters first? This could potentially catch cases like the ones Steve highlighted reviewing an earlier iteration of this PR.

I think it would be useful for the changes to the test cases corresponding to the URL pattern change to be in the same commit. This makes the changes self-contained/atomic. Otherwise you have a commit where if you roll back to it the tests don't pass.

Couple of other minor comments inline.

builder/tests/test_views.py
rw251: That's a lot better thanks. I'm still learning the django model queries.

rw251:
>Would it be valuable to write a test for creating a search without non-alphanumeric characters first?

rw251:
>Would it be valuable to write a test for creating a search without non-alphanumeric characters first?

I suspect we won't hit that situation again, but useful for regression purposes so I've added 2 tests for each scenario. The existing test didn't test the logic as it spoofed the returned codes. The 2 new tests actually test the logic by doing the POST request - similar to how we're currently testing other behaviour of the "new_search" method.

> I think it would be useful for the changes to the test cases corresponding to the URL pattern change to be in the same commit. This makes the changes self-contained/atomic. Otherwise you have a commit where if you roll back to it the tests don't pass.

I'm torn on this. I think the approach of telling a story with the commits in order to help a reviewer is valid - even though that can lead to cases where the system is broken after one commit, but working after another.  Also I think it's fine for a commit to have failing tests e.g the first commit here which is a failing test to demonstrate the error we're trying to fix. In this instance I split the "fix the problem" commit and the "fix the tests that relied on the old way" to help with review, but I can see the argument for combining them.

PR #2418, opened 2025-04-08
Run the Python tests with migrations in CI

### LINKED ISSUE #2392: Enable migrations in CI
# Why are we doing this?

We aren't running migrations during testing. This means that if we have a migration that doesn't work properly, we might not find out until it's deployed into production.

One way of ensuring we are running migrations is to enable them during the testing in CI. Running migrations while running unit tests locally using `just test` will slow down the tests, so we might want to avoid that. 

...

# What are we doing?

...

---

StevenMaude:
Fixes #2392.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,6 +59,10 @@ jobs:
           just docker-build prod
           just docker-build dev
 
+      - name: Check migrations
+        run: |
+          just docker-check-migrations
+
       - name: Run unit tests on docker dev image
         run: |
           # build docker and run tests with migrations
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,8 +61,8 @@ jobs:
 
       - name: Run unit tests on docker dev image
         run: |
-          # build docker and run test
-          just docker-test-py
+          # build docker and run tests with migrations
+          just docker-test-py --migrations
 
       - name: Run smoke test on prod
         run: |
--- a/docker/justfile
+++ b/docker/justfile
@@ -33,6 +33,8 @@ check-js:
 check-py: build
     docker compose run --rm dev /app/docker/entrypoints/check.sh
 
+check-migrations *args="": build
+    docker compose run --rm dev python manage.py makemigrations --check {{ args }}
 
 # run tests in docker container
 test-py *args="":
--- a/justfile
+++ b/justfile
@@ -295,3 +295,8 @@ docker-exec *args="bash": _env
 # run tests in docker container
 docker-smoke-test host="http://localhost:7000": _env
     {{ just_executable() }} docker/smoke-test {{ host }}
+
+
+# check migrations in the dev docker container
+docker-check-migrations *args="":
+    {{ just_executable() }} docker/check-migrations {{ args }}


.github/workflows/main.yml
```diff
@@ -59,10 +59,14 @@ jobs:
           just docker-build prod
           just docker-build dev
 
+      - name: Check migrations
+        run: |
+          just docker-check-migrations
```
StevenMaude, line 64: note: This also runs the migrations, so, in principle, the `--migrations` below is not strictly necessary, but maybe better to keep it in case this line ever gets removed.

Jongmassey: https://github.com/opensafely-core/opencodelists/blob/b167ac128618d388ec2b20dd0ac49da26d21b70c/.github/workflows/main.yml#L66-L69 this one? Just checking. It seems to make sense to me

Jongmassey:
I can't seem to run this locally:

```sh
$ just docker-check-migrations
/usr/bin/just docker/check-migrations 
docker compose run --rm dev python manage.py makemigrations --check 
Traceback (most recent call last):
  File "/app/./manage.py", line 22, in <module>
    main()
  File "/app/./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 416, in execute
    django.setup()
  File "/opt/venv/lib/python3.12/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/opt/venv/lib/python3.12/site-packages/django/apps/registry.py", line 91, in populate
    app_config = AppConfig.create(entry)
                 ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/apps/config.py", line 193, in create
    import_module(entry)
  File "/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'slippers'

error: Recipe `check-migrations` failed on line 37 with exit code 1
error: Recipe `docker-check-migrations` failed on line 302 with exit code 1
```

StevenMaude:
@Jongmassey: It still works here.

It's an odd error you're seeing though: the dependency installation shouldn't be related to this change.

What if you try something like: `just docker-run python manage.py shell`?

Or if you try a `docker compose down` in `./docker` and then rebuild, maybe?

docker/justfile
```diff
@@ -33,6 +33,8 @@ check-js:
 check-py: build
     docker compose run --rm dev /app/docker/entrypoints/check.sh
 
+check-migrations *args="":
```
Jongmassey, line 36: should this call the `build` recipe like `check-py` does? The errors I were seeing were due to not having run `build` manually in advance.

Jongmassey (approved):
This doesn't work if either the `DATABASE_URL` or `DATABASE_DIR` env vars are set to paths outside of the project root, but almost none of the docker recipes work if they are so that'll be a separate issue I'll write up!

Issue #2403, opened 2025-04-07
Spike: Approach to fixtures


mikerkelly:
# Why are we doing this?

`opencodelists` has a rich set of fixtures but we've identified several issues:

- Test cases are not independent. Whether tests pass or not depends which tests they are run with, due to how the fixtures are created. Some tests don't pass in isolation. See [here](https://github.com/opensafely-core/opencodelists/issues/2301) and [here](https://github.com/opensafely-core/opencodelists/issues/2115#issuecomment-2587843892).
- The SNOMED fixtures reproduce a small subsection of the real SNOMED heirarchy from a particular version, with real concepts with complicated vocabulary terms and full codes. This could be simpler and easier to grasp if we invented a small tree with simple invented terms and codes.
- The fixtures define a "universe" of model instances that represent a valid state of the system. These are always set up in full whenever any fixture is requested. Possibly these persisting in memory or the database is related to the tests not being independent. The fixtures expect that tests don't create instances directly, but this is not enforced (and it seems reasonable that tests would want to do that). `pytest` fixtures usually have a hierarchy where fixtures can depend on other fixtures, but it instantiates exactly what's needed, not other objects. So this works against the framework. It may also be slow.

# How will we know when it's done?

We have considered and possibly prototyped approaches to the above issues, and raised new ones for future work.

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

mikerkelly:
The most pressing issue was the first bullet, tests failing in isolation, which poses a problem for developers working on the current initiative. Nearly all tests were unable to run in isolation. Investigating this was complex due to the multiple databases, the way the 'universe' of objects is created, and the interactions between various fixtures. This investigation took up the majority of the time spent on the spike.

I’ve created a PR that resolves 99% of the issues and raised an issue for an additional 2 test cases that still need fixing. Collectively, these changes have increased the number of test cases executable in isolation from 3% to 99%. See the next comment for details.

I can document my understanding of the current fixtures and address the remaining ticket points after I return from annual leave during the week commencing 12th May.

mikerkelly:
# Investigation - tests non running in isolation

[...]

PR #2421, opened 2025-04-09
Add some simple Playwright tests

### LINKED ISSUE #2408: Add basic Playwright test
# Why are we doing this?

We want to be able to confidently make changes to the builder, and other aspects of creating a codelist, and we believe that having a set of automated end-to-end tests will help with that. 

We will have a very basic end-to-end test that runs as part of CI, which goes through the "happy path" of creating a codelist.

# What are we doing?

* Adding playwright into OpenCodelists (we should look at how it's been implemented in Airlock to avoid reinventing too many wheels).
* Creating a basic first test, using some very simple dummy data, which goes through the "happy path" of creating a codelist.
* Allowing this to be run locally as part of development
* Adding this to CI and ensuring it runs in a reasonable time, so that it doesn't significantly slow deployments.
* Sharing how this works with the rest of the team!
* Adding an ADR to record any decisions and trade offs of the approach taken.

---

StevenMaude:
Fixes #2408.

Much of the starting point for this is based on how browser tests were implemented in Airlock (credit primarily to Becky, I think).

### State testing

* Testing of states with multiple searches is more complicated:
  * I've discussed [some of this below](https://github.com/opensafely-core/opencodelists/pull/2421#issuecomment-2867651284) and in commit messages
  * we'll probably need to use the actual backend code to do this thoroughly; see #2541
  * otherwise, we're restricted to doing what I've done for now, which is explicitly marking what to do for each entry for a search; this allows us to compute the final state
* I've not yet implemented checking of the "Searches" tab because it's a tiny bit more complicated: see #2546.

### Structure of the tests

* These all mostly have the same setup pattern that possibly lends itself to having a class-based structure that abstracts almost all the details. That could well be appropriate, but I haven't done this yet, as I'm not sure how generalisable any future tests might be.
* The structure is generally:
  1. specify coding system, codelist name
  1. create a codelist
  1. specify a series of searches and associated selections
  1. perform those searches and selections
  1. save the codelist for review
  1. verify the codes in the review table
  1. verify the downloaded CSV
* You could envisage reducing this down to specifying:
  * coding system
  * codelist name
  * searches and selections
  * whether or not to validate the codelist
* See #2542.
* It's perhaps a bit more complicated to use pytest fixtures in such a class, because of the way we create the Playwright `page`, using another login fixture to do that. Maybe it's possible?

### Fixtures

* Below this immediate point is a lot of technical discussion on how the fixtures do and could work. For now, it might be fine to use the current non-ideal approach of using the existing fixtures; see #2543 for future work.
* We're using the existing fixtures, and modifying their behaviour slightly so that their state is reset between tests for the functional tests. This is not ideal. It may be acceptable enough in the short-term to get the benefit of adding browser tests, with a view to possibly reworking this setup very soon.
  * If all the fixtures get built each time, some of this state reset is unnecessary extra work. The page load and the six simple codelist related tests take about 30 seconds to run in CI, including installing Playwright, which isn't bad though.
  * This also has quirks for using some of the fixtures where there's a mismatch between `session` and `function` scopes; the `universe` fixture having a `session` scope.
* On the `main` branch, the `universe` fixtures currently end up always being imported via `conftest.py` in the root directory, which makes an entire separation and isolation a little bit more awkward.
* We could consider using entirely self-contained fixtures, via pytest's [`confcutdir`](https://github.com/pytest-dev/pytest/blob/33b26e429817935af8eb0202aa1e5ef0476e1cad/doc/en/reference/reference.rst?plain=1#L2203) option, and using the `conftest.py` in `opencodelists/tests/functional` only.
  * This would allow a start from a clean state, and would better isolate the functional tests. One problem here is that *everything* gets imported from the test "universe" in the top-level (that is, most "global") `conftest.py`. However, we actually need some of the setup from the top-level `conftest.py` to configure the databases, so there would be necessary duplication if we start afresh.
  * But maybe the current workaround is acceptable, pending any other future fixture improvements (#2403), rather than creating a new collection of fixtures.
  * In some cases, those fixtures will mirror the existing test fixtures, so avoiding that duplication might also be beneficial to simplifying the tests overall. We've had to do this already to get an organisation user, because of the issue above with the `universe` fixture `session` scope.

### Coding systems not yet tested

* There are no ICD10 builder tests yet, because those fixtures work differently from the BNF and SNOMED CT fixtures, and that needs some extra work. (If you try and use the existing `icd10_data` fixture, you get a 500 error when creating a codelist.)
* See #2557.


StevenMaude:
Also maybe relevant: [live-server documentation](https://pytest-django.readthedocs.io/en/latest/helpers.html#live-server) which mentions interactions with databases.

StevenMaude:
Naturally, I find what the likely cause of the problem is by one random internet search, landing me on the [pytest-django documentation](https://pytest-django.readthedocs.io/en/latest/database.html#populate-the-test-database-if-you-use-transactional-or-live-server)[^1]; it's the use of pytest-django's `live_server`:

> In case you use transactional tests (you use the [`pytest.mark.django_db()`](https://pytest-django.readthedocs.io/en/latest/helpers.html#pytest.mark.django_db) marker with `transaction=True`, or the [`transactional_db`](https://pytest-django.readthedocs.io/en/latest/helpers.html#std-fixture-transactional_db) fixture), you need to repopulate your database every time a test starts, because the database is cleared between tests.
>
> The [`live_server`](https://pytest-django.readthedocs.io/en/latest/helpers.html#std-fixture-live_server) fixture uses [`transactional_db`](https://pytest-django.readthedocs.io/en/latest/helpers.html#std-fixture-transactional_db), so you also need to populate the test database this way when using it.

[^1]: [Don’t go to Stack Overflow, don’t ask the LLM, don’t guess, just go straight to the source.](https://endler.dev/2025/best-programmers/)

rw251:
I had a quick (ish) look. I could get it working by changing the test definition to:

```python
def test_codelist_builder(login_context, snomedct_data, live_server):
```
I.e. by calling the session scoped `snomedct_data` fixture. I thought maybe playwright-pytest was running in a different session, but if I set a breakpoint in the `snomedct_data` fixture it's only called once, so not really sure what's going on.

StevenMaude:
Thinking about separating out these functional tests from the unit tests, which also makes sense in that we want the unit tests to stay fast.

Here are some work-in-progress notes that may receive further edits.

# The problem

* pytest-django's `live_server` is necessary to run the functional tests.
* `live-server` clears the database between tests.
* We create database state via the top-level `conftest.py` as a one-time setup.
* We need at least:
  * a test user
  * more problematic: some minimal coding system database import so that we can actually create a minimal codelist against a database

# How do the database fixtures work now?

Note that there are several `conftest.py` files with differing scope.

What we're particularly interested in are the database fixtures.

Briefly, the database fixtures appear to work via:

* `opencodelists/tests/fixtures.py`
  * `universe()` is a session fixture that calls `build_fixtures()`
* In the top-level `conftest.py`
  * There's an import of everything from `opencodelists.tests.fixtures` that imports everything here, and therefore makes all these fixtures available to all tests.
  * `django_db_modify_db_settings` configures some database settings.

# Ideas dump

* Use the existing `conftest.py`; effectively using the `universe` fixture
  * Has lots of things we probably don't need for the functional tests.
  * We should try and minimise what we have imported for the tests.
* Populate with a smaller `conftest.py`.
  * Can we override the existing behaviour we have at the top-level? Can we override the `universe` fixture and provide a minimal state, maybe using the same fixtures in the `opencodelists/tests/fixtures.py`?
* Work around restrictions in live server?
  * [Stop the database from being destroyed?](https://stackoverflow.com/questions/52561816/pytest-django-add-fixtures-to-live-server-fixture)
  * This is probably bad because it persists state we don't need, and doesn't give a clean database for each test, reducing isolation.
* If we can't easily override the use of the existing fixtures, do we need to load the state into different databases entirely, and use those?
* Could overriding `django_db_setup` or `django_db_modify_db_settings` help?
  * How does that interact with the existing setup?
 
## TODO
 
- [ ] Separate out functional tests into marked functional tests and run these separately
- [ ] Remove pytest-order because I don't think it'll be necessary when we fix the database state

StevenMaude:
I've pushed the old state of this as of d0629c83 to `steve/playwright-testing-old`. I'll be rebasing and squashing commits here.

StevenMaude:
I think the issue with the dependency warning is due to where Playwright gets installed, compare from the Docker build:

```
Chromium Headless Shell 134.0.6998.35 (playwright build v1161) downloaded to /root/.cache/ms-playwright/chromium_headless_shell-1161
…
```

to when the tests are run:

```
Chromium Headless Shell 134.0.6998.35 (playwright build v1161) downloaded to /app/.playwright-browsers/chromium_headless_shell-1161
```

I don't quite understand why the user-based install doesn't see the existing installed dependencies though.

StevenMaude:
A quirk with the workaround: if you access other fixtures other than the coding systems, you may get a failure due to mismatched scope access ("function" for fixtures, and "session" for `universe`), but if you set "session", then you'll need to mark every functional test as using databases.

rw251:
I haven't reviewed this, but just responding to your request for feedback.

1. Regarding fixture approach - I agree with this:
> ... maybe the current workaround is acceptable, pending any other future fixture improvements ([Spike: Approach to fixtures #2403](https://github.com/opensafely-core/opencodelists/issues/2403)), rather than creating a new collection of fixtures.

2. The tests run on my machine (but it's linux so not that useful for you)
3. Some thoughts on potential test improvements:
  - maybe check that "Save for review" is disabled until all items are marked
  - Instead of "Save draft" (which as far as I can see does nothing except display a reassuring message) we should probably go straight to "Save for review" - which makes it more of an end 2 end test - and then :
	- check the codes we expect are on the "Full list" tab (and that the codes we excluded are not)
	- check the search term is on the "Searches" tab
	- then proceed to publish - and maybe more checks around that
  - Instead of clicking the nth +/- you could create a list of expected search results, and for each mark it as whether we want to include it or not (and also whether to expand it or not for the nested ones). You could then do something like this: 
```python
expected_concepts = [
  {"code": "3723001", "expand": False, "include": True},
  {"code": "439656005", "expand": True, "include": False},
  {"code": "202855006", "expand": False, "include": True},
]

for concept in expected_concepts:
    concept_locator = page.locator(f"[data-code='{concept['code']}']")
    if concept["expand"]:
        concept_locator.get_by_role("button", name="⊞").click()
    button_icon = "+" if concept["include"] else "−"
    concept_locator.get_by_role("button", name=button_icon).click()
```
This has the advantage that when checking the expected results in the saved codelist you can just loop through the list again and check the included things are included and the excluded things are excluded.

StevenMaude:
This is getting closer, but still needs some tidying up.

What we could cover more:

* multiple searches properly
* other exclusions/inclusion states
* versions, and changing the exclusion/inclusion states

StevenMaude:
Thinking about this a bit more today… here are some notes, but may apply to future work and note here, in which case I'll move into the specific issue. In lieu of a better place for now, here they are anyway.

# Obtaining the final state from the specified selections

There are some basic end-to-end tests here using a single search.

We specify the state of the concepts in the builder as included or excluded.

With this, we have explicitly stated the selection state of each code in the search. Because of this, we can determine what the final state should be for each code in this simple case.

That is, we:

* specify the states (included/excluded) of the concept selections in the builder, in order they should be applied
* iterate through those states
* take the later state corresponding to a code as being the final state

# Obtaining the final state in more complex scenarios

One obvious candidate for more complex tests in future might involve using different searches.

Actually automating the steps with Playwright isn't much more complicated. We currently have function calls that perform the task of searching or applying concept selections. We can repeat those.

What I anticipate is more complicated is computing that expected final overall state.

I haven't tried out the builder enough to know, but here are examples of more complicated cases.

## Example 1

* add one search; select some concepts, maybe including some concepts are implicitly included
* return to the same search; remove part of that included hierarchy

## Example 2

* add one search; select some concepts
* add another search; select some concepts that overlap with the first and previously unselected

## Example 3

* add one search; select some part of the hierarchy and its descendents
* add another search that shows the same part of the hierarchy; remove some of the included descendents

## Example 4

* add one search; select some concepts
* add another search; select some concepts
* delete the first search (I think this is the behaviour that Jon talked about when we met in person); what's the behaviour here? how does it appear in the review page?

## Example 5

* add one search; select some concepts
* delete the search; what's the behaviour here?

# Possible ways to handle this
 
From having thought about it more than looking at the code or testing things out (so I may be wrong), I think we could choose one of the following:

* **Don't compute the final state at all**: just write out the expected state by hand.
  * This means a correct result is defined in the test, and not by the backend. This would work, but be extra verbosity to the tests.
* **Do what we do now**: specify an explicit inclusion state for each code in each search, and then expand the full hierarchy if it needs expanding.
  * From this, maybe we could compute the final state by collating those selections and there aren't other quirks?
  * This means you have to expand each part of the hierarchy and include/exclude each code, which does not cover all ways a user might create a codelist. Maybe a user implicitly includes codes by selecting something higher up the hierarchy?
* **Use the appropriate backend code to figure out which codes are included in the end, per search if presumably possible.**
  * That is:
    * create the appropriate codelist/codeset object that records this data
    * add the codes (and searches?) to it/them
    * and then use their methods to evaluate the expected results we should see in the frontend 
  * This way we assert that the frontend code gives the same result as the backend, and feels most correct to do.
* **Write more complicated state handling to figure out the state** in the same way as the backend code, and handle the general case.
  * Probably seems an unnecessary duplication of work, if the backend does this for us.

# Other thoughts

It may also be that the behaviour in some of these cases is not yet as we would like it.

StevenMaude:
Also, it's possible that the tests lend themselves to having a class to create them, since the:

* existing tests that we add here have a fairly common pattern
  * setup some common variables (name of test, coding system, searches, concept selections)
  * build a codelist: create codelist, do a search, add codes
  * verify the codelist
* functions we're creating to abstract parts of this pattern often take several arguments from the original test function, and would be simpler to pass `self`

I've left this for now, pending seeing how other tests we're interested in might pan out and how generalisable a pattern this is, but I think this is a candidate for refactoring in future.

One possible slight complication there is around pytest fixtures, which I've found a little clunkier to work with when being used inside classes.

DEVELOPERS.md
```diff
@@ -199,6 +199,18 @@ just test-py
...
+just test-functional
+
+# run Playwright functional tests with a visible browser
+# This is useful in combination with script pausing when writing tests:
+# https://playwright.dev/python/docs/api/class-page#page-pause
+#
+# Playwright also has other debugging options configurable
+# via the PWDEBUG environment variables, see:
+# https://playwright.dev/python/docs/debug
+just test-functional --headed
```
rw251, line 212: Might be worth giving both commands that I think developers will likely use e.g. 
1. For watching a specific test `just test-functional <path_to_test> --headed --slowmo=500`. 
2. For debugging a specific test `PWDEBUG=1 just test_functional <path_to_test>`

opencodelists/tests/functional/test_build_codelist.py
```diff
@@ -0,0 +1,583 @@
...
+    """Takes a Playwright page, a codelist name, and heading text to match
+    and validates that the codelist name exists under that heading.
+
+    It does not currently go for an exact match because organisation users get
+    "owned by organisation" appended to the codelist name.
+
+    We could improve this matching in future."""
+
+    page.get_by_role("link", name="My codelists").click()
+    # TODO: can we remove the duplication here?
```
rw251, line 161: Do you mean the duplication of `page.get_by_role("link", name=codelist_name)`? In which case you can just do `codelist_link = page.get_by_role("link", name=codelist_name)` - but maybe that's not what you meant.

StevenMaude: Specifically, I think my question was: is it OK/idiomatic to assign such a thing you'll then check with `expect`, before you do that check?

rw251: Assigning locators is fine. There's a section in [the docs](https://playwright.dev/python/docs/locators#locating-elements) about it with examples. Also the locator reevaluates each time the assigned variable is used.

docs/adr/0003-browser-testing.md
```diff
@@ -0,0 +1,40 @@
...
+those. It is possible that the existing test fixture system is redesigned in
+future to simplify it, and that simplification may help setup of these
+tests.
+
+These tests will be run separately from the existing unit tests. This is
+partly a design choice — to keep the existing test suite fast — and
+partly a consequence of the fixture setup — pytest-django's
+`live_server` fixture resets database state between tests.
+
+## Consequences
```
mikerkelly, line 34: Can we state that the consequence of reusing the universe fixture is the need for the workaround to change the fixture scope for these new tests, and why?

docs/adr/0003-browser-testing.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Status
+
+Draft
+
+## Context
+
+While we have unit tests for parts of our code, both frontend and
+backend, user flows throughout the application end-to-end, integrating
+these are not tested.
```
mikerkelly, line 13: I think the first paragraph is a bit garbled and although I get the gist I'm not sure I follow it grammatically.

```suggestion
While we have unit tests covering parts of our frontend and
backend code, we lack end-to-end tests for user flows across
the application.
```
- Should we say that we don't currently enforce any Coverage metric, so the tests we do have are definitely not comprehensive?
- Can we give some examples of what a "user flow" is? Registering, viewing an existing codelist, creating a minimal codelist with the builder?
- Can we say anything about why Playwright was chosen over alternatives like Selenium or Cypress?

docs/adr/0003-browser-testing.md
```diff
@@ -0,0 +1,40 @@
...
+
+It would be useful to have such tests to validate that key user journeys
+continue to function as expected.
```
mikerkelly, line 16:
```suggestion
It would be useful to have such tests to validate that key user journeys
continue to function as expected when making changes to the code.
```

docs/adr/0003-browser-testing.md
```diff
@@ -0,0 +1,40 @@
...
+## Decision
+
+We will add automated browser tests that are run using Playwright and
+pytest-django's `live_server` fixture.
+
+We will use the existing `universe` fixtures for the time being to setup
+the coding system data where appropriate, rather than reimplementing
+those. It is possible that the existing test fixture system is redesigned in
+future to simplify it, and that simplification may help setup of these
+tests.
```
mikerkelly, line 27: Can we very briefly state what this is and why that causes a complication/workaround?

docs/adr/0003-browser-testing.md
mikerkelly: Worth stating the positive consequence (why we are doing this): we will have greater assurance there aren't unintended consequences to changes we make?

docs/adr/0003-browser-testing.md
```diff
@@ -0,0 +1,40 @@
...
+
+These tests will be run separately from the existing unit tests. This is
+partly a design choice — to keep the existing test suite fast — and
+partly a consequence of the fixture setup — pytest-django's
+`live_server` fixture resets database state between tests.
+
+## Consequences
+
+The tests will need to be maintained, and new tests added if we add
```
mikerkelly, line 36: Also adding additional tests to cover additional user journeys these initial tests don't cover? Should it be something we pursue stand-alone or just in the context of other tickets? How can we encourage that activity?

docker/justfile
```diff
@@ -45,12 +45,18 @@ test-py *args="":
     --cov=mappings \
     --cov=opencodelists \
     --cov-report html \
-    --cov-report term-missing:skip-covered {{ args }}
+    --cov-report term-missing:skip-covered \
+    -m "not functional" {{ args }}
```
mikerkelly, line 49: I'm not sure what the name of the '-m' flag stands for? Maybe a longer name would be clearer? I haven't understood how pytest options/config work, sorry if I'm missing something.

mikerkelly:
I think the workaround for adjusting the scope of the opencodelists/tests/fixtures.py fixtures is fine for now. I hope that aligns with whatever https://github.com/opensafely-core/opencodelists/issues/2523 does -- data and model instances persisted in the memory of the module, and only saved to the database at a function-scope.

I like the general approach of this PR but haven't done a detailed review. Left a few minor comments.

docker/justfile
mikerkelly: Ah, thanks. I hadn't appreciated the workaround is slightly overloading this in-built pytest configuration. I think that seems fine overall.

opencodelists/tests/functional/test_build_codelist.py
```diff
@@ -0,0 +1,583 @@
...
+    #    concept_html_list_items = page.locator("div#search-results").locator(
+    #        "details ul li"
+    #    )
+    #    for list_item in concept_html_list_items.all():
+    #        # TODO: extract the colour and use that to determine included
+    #        # TODO: if included, extract name and code from text content
+    #        print(list_item.text_content())
+
+
+def publish_codelist(page, codelist_name):
```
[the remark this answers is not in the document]
StevenMaude: I've done this and parametrized the test to go through two paths: "always navigate to the required page" and "navigate to required page when it differs from the one the browser is on".

DEVELOPERS.md
StevenMaude: Done.

opencodelists/tests/functional/test_build_codelist.py
```diff
@@ -0,0 +1,632 @@
...
+def setup_playwright_page(login_context, url):
+    """Takes a login context, and returns a Playwright page."""
+
+    page = login_context.new_page()
+    page.goto(url)
+    expect(page.get_by_role("button", name="My account")).to_be_visible()
+    return page
+
+
+def create_codelist(navigator, coding_system, initial_page_navigation=True):
```
StevenMaude, line 178: thought: This is more of a general comment, rather than a suggestion that maybe we need to change anything yet.

Functions like `create_codelist` and the other imperative actions all take a `Navigator`, and feel like they could be almost methods on that class.

Another alternative might be that they are instead methods on a test factory class.

I'm not sure what the right structure is here, so they're left as functions for now.

I don't think that matters for now, but it's worth bearing in mind. I like the current separation that the `Navigator` only goes to pages, rather than does other actions.

For example, if we made it so the `Navigator` performs actions, then I'm not sure if it's confusing or not that, say, a `SearchAction` has an `apply` method that also performs actions with the `Navigator`?

opencodelists/tests/functional/test_build_codelist.py
```diff
@@ -161,9 +162,8 @@ def validate_codelist_exists_on_site(page, codelist_name, heading_text):
     We could improve this matching in future."""
 
     page.get_by_role("link", name="My codelists").click()
-    # TODO: can we remove the duplication here?
-    # Note that this partially matches the name, unless exact=True.
-    expect(page.get_by_role("link", name=codelist_name)).to_be_visible()
+    codelist_link = page.get_by_role("link", name=codelist_name)
+    expect(codelist_link).to_be_visible()
     expect(
         page.get_by_role("link", name=codelist_name)
```
rw251: You missed replacing this line with `codelist_link`

opencodelists/tests/functional/test_build_codelist.py
```diff
@@ -118,6 +118,15 @@ def apply(self, page):
             search_action.handle_concept_selections(page)
 
 
+class CodelistHeading(StrEnum):
+    """Represents the heading for a codelist on the My Codelists page."""
+
+    USER_OWNED = "Your codelists"
```
rw251, line 163: This will change after rebasing to main because my recent work on that page has changed 2 headings. "Your codelists" is now "Your published codelists", and "Your drafts" is now "Your draft codelists"

rw251 (approved):

Issue #2420, opened 2025-04-09
Django 5.2 Upgrade


Jongmassey:
# Why are we doing this?

We are currently running Django 5.1, for which the end of extended support is 2025-12.
Django 5.2 was [released](https://docs.djangoproject.com/en/5.2/releases/5.2/) 2025-04-02.

Our dependency updating system is [starting to suggest](https://github.com/opensafely-core/opencodelists/pull/2399/files) upgrading to 5.2.

There are some major changes in 5.2 which might cause us problems.

We neither have 100% test coverage, nor good automated end-to-end tests. 

# How will we know when it's done?

When OpenCodelists runs successfully on Django 5.2 — both automated and manual testing of all features is successful — in production

# What are we doing?

- [x] Pin django to `>=5.1,<5.2` such that other dependency updates may proceed. #2423 
- [ ] Test the upgrade to 5.2
- [ ] Un-pin django to allow updates to proceed

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

StevenMaude:
We do at least have some end-to-end testing now, which gives us a little more confidence in upgrades. This may still need some manual testing too.

PR #2423, opened 2025-04-10
Pin django to v5.1.x

Jongmassey:
`django` is already at 5.1.8 in requirements.[prod|dev].txt so trying to update just this package has no effect, but definitely does not bump to 5.2


StevenMaude (approved):
thought: This looks fine; you could maybe link to the existing issue #2420 either in the commit comment or a comment here, so someone looking at the requirements knows the reason for it's pinned.

Jongmassey:
Seems sensible, comment added!

PR #2430, opened 2025-04-14
Populate development dbs with test fixture data

### LINKED ISSUE #2313: Populate development DB(s) with test fixture data
Following from discussion in #2117 it seems like using the test fixture data as dummy data for development is an attractive alternative to using a copy of the production database.

The OpenCodelists tests contain a number of fixtures (e.g. [here](https://github.com/opensafely-core/opencodelists/blob/main/opencodelists/tests/fixtures.py) but NB there exist fixtures in other apps) which are small, internally consistent, comprehensive (in terms of coding systems and site features), and devoid of user PII.

The comprehensiveness of these test fixtures ensures adequate test coverage, but also makes them potentially useful for interactively and iteratively developing new features/fixing bugs.

Being PyTest fixtures, these are loaded into in-memory sqlite databases which are created at the beginning of a test session and destroyed at the end. To be useful as development data, a process to load them into persistent on-disk sqlite databases would be required.

Directly calling the test fixtures outside of a PyTest test is not permitted so some other way around this may be required (see [this comment](https://github.com/opensafely-core/opencodelists/issues/2117#issuecomment-2517454652) for relevant prior art in ehrQL).

This work may or may not be affected to changes to our test database configuration required for upgrading to Django 5.1.x in #2115

KatieB5:
Closes #2313.

### What has been done?
A `setup_local_dev_databases` management command deletes, rebuilds and populates databases (dbs). This is called from a `just` recipe which deletes and rebuilds the core db. Both are safe by default. The result is a lightweight local development setup that uses our test data. 

You can use the `just` recipe in **safe-mode**: `just build-dbs-for-local-development`
To actually **delete local dbs** if you have them, pass the `nuclear` argument: `just build-dbs-for-local-development nuclear`

#### A note on design choices
Originally, deletion of the core db occurred inside the management command. However, this led to unexpected behaviour. When a Django management command is invoked, it triggers `django.setup()` and the [app initialization process](https://docs.djangoproject.com/en/5.1/ref/applications/#initialization-process). During initialization, Django loads `settings.py`and expects the core database to exist at the path in `DATABASES`. If it's missing, SQLite creates a `db.sqlite3` file automatically. If the file is deleted after initialization, when migrations are run later in the same process, they don’t trigger a reinitialization of the app and re-check for the presence of the database file, so a new core db isn't recreated. 

Addressing this would require reinitializing the Django app from within the command — a potentially complex rabbit hole. Instead, the deletion step was moved out of the management command and into the `just` recipe.

### DIFF
--- a/justfile
+++ b/justfile
@@ -247,6 +247,42 @@ assets-test: assets-install
     npm run test:coverage
 
 
+# Build a lightweight local development setup using test fixture data.
+build-dbs-for-local-development nuclear="":
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # WARNING: Passing 'nuclear' to this just recipe will create a
+    # backup copy of your current local db, if it exists, then create a
+    # new empty core `db.sqlite3` in the root directory of your
+    # opencodelists project folder.
+    if [ -n "{{ nuclear }}" ]; then
+
+        if [[ -z "${DATABASE_URL:-}" ]]; then
+            CORE_DB_PATH="db.sqlite3"
+            else
+            CORE_DB_PATH="${DATABASE_URL/sqlite:\/\/\//}"
+            fi
+        echo "Nuclear option enabled: moving $CORE_DB_PATH to $CORE_DB_PATH.backup"
+
+        mv "$CORE_DB_PATH" "$CORE_DB_PATH.backup"
+
+        # Set up or update the local dev environment:
+        # - Recreates core DB and applies migrations
+        just dev-setup
+
+        # Run custom command to:
+        # - Load CodingSystemReleases needed for the test data fixtures
+        # - Remove old coding system release dbs (with confirmation)
+        # - Create and migrate new coding system release dbs
+        # - Load test data into coding system release dbs
+        $BIN/python manage.py setup_local_dev_databases
+    else
+        echo "Skipping creation of a new empty core db.sqlite3. Run with 'nuclear' parameter to enable."
+    fi
+
+
+
 # build docker image env=dev|prod
 docker-build env="dev": _env
     {{ just_executable() }} docker/build {{ env }}
--- a/opencodelists/management/commands/setup_local_dev_databases.py
+++ b/opencodelists/management/commands/setup_local_dev_databases.py
@@ -1,0 +1,120 @@
+import sys
+
+from django.core.management import BaseCommand, call_command
+
+from coding_systems.versioning.models import (
+    CodingSystemRelease,
+    build_db_path,
+    update_coding_system_database_connections,
+)
+from opencodelists import settings
+from opencodelists.tests.fixtures import build_fixtures
+
+
+class Command(BaseCommand):
+    help = """Populates local opencodelists dbs with test fixture data:
+           Loads the CodingSystemReleases needed for the snomed and dmd
+           fixtures into the Core db (you will be prompted for consent
+           to do this), deletes existing test coding system release dbs,
+           creates new coding system release dbs, populates these with
+           test fixture data, instantiates a universe of fixture
+           objects, and creates a superuser. Running this command
+           directly is not recommended, instead use `just
+           build-dbs-for-local-development`"""
+
+    def handle(self, *args, **options):
+        if not settings.DEBUG:
+            self.stderr.write(
+                "This management command will not run in a production environment."
+            )
+            sys.exit(1)
+
+        load_versioning_fixture_answer = input(
+            "Running this command will load the CodingSystemReleases needed for the snomed and dmd test fixtures, into the Core db. Any local test coding system release dbs will then be deleted and recreated. Do you want to proceed? (Y/n)"
+        )
+
+        if load_versioning_fixture_answer == "n":
+            return
+
+        call_command(
+            "loaddata",
+            "coding_systems/versioning/fixtures/coding_system_releases.json",
+        )
+
+        self.stdout.write("Deleting local coding system release dbs...")
+        for coding_system_release in CodingSystemRelease.objects.all():
+            db_path = build_db_path(coding_system_release)
+            if db_path.exists():
+                db_path.unlink()
+                self.stdout.write(f"{str(db_path)} has been deleted")
+
+        self.stdout.write(
+            "Migrating coding system release dbs and populating them with test fixture data"
+        )
+
+        update_coding_system_database_connections()
+
+        # migrate snomedct test db and load test fixtures
+        call_command("migrate", "snomedct", database="snomedct_test_20200101")
+
+        call_command(
+            "loaddata",
+            "coding_systems/snomedct/fixtures/core-model-components.snomedct_test_20200101.json",
+            database="snomedct_test_20200101",
+        )
+
+        call_command(
+            "loaddata",
+            "coding_systems/snomedct/fixtures/tennis-elbow.snomedct_test_20200101.json",
+            database="snomedct_test_20200101",
+        )
+
+        call_command(
+            "loaddata",
+            "coding_systems/snomedct/fixtures/tennis-toe.snomedct_test_20200101.json",
+            database="snomedct_test_20200101",
+        )
+
+        # migrate dmd test db and load test fixture
+        call_command("migrate", "dmd", database="dmd_test_20200101")
+
+        call_command(
+            "loaddata",
+            "coding_systems/dmd/fixtures/asthma-medication.dmd_test_20200101.json",
+            database="dmd_test_20200101",
+        )
+
+        # migrate bnf test db and load test fixture
+        call_command("migrate", "bnf", database="bnf_test_20200101")
+
+        call_command(
+            "loaddata",
+            "coding_systems/bnf/fixtures/asthma.bnf_test_20200101.json",
+            database="bnf_test_20200101",
+        )
+
+        # migrate icd10 test db and load test fixture
+        call_command("migrate", "icd10", database="icd10_test_20200101")
+        call_command(
+            "loaddata",
+            "coding_systems/icd10/fixtures/icd10.icd10_test_20200101.json",
+            database="icd10_test_20200101",
+        )
+
+        # instantiate the test data universe
+        build_fixtures()
+
+        call_command(
+            "createsuperuser",
+            no_input=True,
+            username="localdev",
+            email="localdev@example.com",
+        )
+
+        self.stdout.write(
+            "Local setup complete! You can now:\n"
+            " - Log in as `localdev`\n"
+            " - Search for 'arthritis', 'tennis', and 'elbow'\n"
+            " - Build codelists with the concepts returned from these searches (see /opencodelists/opencodelists/tests/fixtures.py for more info)\n"
+            " - View a BNF codelist, a minimal codelist, and a new-style SNOMED CT codelist"
+        )
--- a/opencodelists/tests/management/commands/test_setup_local_dev_databases.py
+++ b/opencodelists/tests/management/commands/test_setup_local_dev_databases.py
@@ -1,0 +1,14 @@
+import pytest
+from django.core.management import call_command
+
+# from opencodelists.management.commands import setup_local_dev_databases
+from opencodelists import settings
+
+
+def test_setup_local_dev_databases_exits_early_when_in_prod_environment(monkeypatch):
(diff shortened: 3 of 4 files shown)


justfile
```diff
@@ -247,6 +247,32 @@ assets-test: assets-install
...
+    set -euo pipefail
+
+    # WARNING: Passing 'nuclear' to this just recipe will delete
+    # `db.sqlite3` if it exists in the project root.
+    if [ -n "{{ nuclear }}" ]; then
+        echo "Nuclear option enabled: deleting db.sqlite3"
+        rm -f db.sqlite3
+        # Set up or update the local dev environment:
+        # - Recreates core DB and applies migrations
+        just dev-setup
```
Jongmassey, line 262: What was your thinking about referencing this here rather than at the start of the recipe? E.g.

`build-dbs-for-local-development nuclear="": dev-setup`

opencodelists/management/commands/setup_local_dev_databases.py
```diff
@@ -0,0 +1,103 @@
...
+from opencodelists import settings
+from opencodelists.tests.fixtures import build_fixtures
+
+
+class Command(BaseCommand):
+    help = """Populates local opencodelists dbs with test fixture data: Loads the CodingSystemReleases needed for the snomed and dmd fixtures into the Core db (you will be prompted for consent to do this), sets up new coding system release dbs, populates them with test fixture data, instantiates a universe of fixture objects, and creates a superuser."""
+
+    def handle(self, *args, **options):
+        load_versioning_fixture_answer = input(
+            "Running this command will load the CodingSystemReleases needed for the snomed and dmd test fixtures, into the Core db. Do you want to proceed? (Y/n)"
```
Jongmassey, line 13: There is no warning about deleting all the databases here!

opencodelists/management/commands/setup_local_dev_databases.py
```diff
@@ -0,0 +1,103 @@
...
+            "\nFound the following databases in your local opencodelists setup:\n"
+        )
+
+        for path in database_paths:
+            self.stdout.write(str(path))
+
+        self.stdout.write("Deleting local coding system release dbs...")
+        for path in database_paths:
+            path.unlink()
+            self.stdout.write(f"{str(path)} has been deleted")
```
Jongmassey, line 36: I should have read the code before running...


```sh
Deleting local coding system release dbs...
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2023-220_20230220.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-920_20240916.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2023-1220_20231218.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2025-320_20250317.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-1230_20241223.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-720_20240715.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-620_20240617.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-1000_20241007.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2023-1000_20231002.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2025-310_20250310.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_1210_20221212.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-1130_20241125.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-1100_20241104.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-700_20240701.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2023-320_20230320.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-120_20240115.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2023-620_20230619.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2023-920_20230918.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2023-1110_20231113.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-810_20240812.sqlite3 has been deleted
/home/jongmassey/opencodelists-data/coding_systems/dmd/dmd_2024-500_20240506.sqlite3 has been deleted
```

I guess it shows it picks up the database path from the env vars correctly!

Jongmassey: Perhaps asking for confirmation would have been nicer!

opencodelists/management/commands/setup_local_dev_databases.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self.stdout.write(
+            "Migrating coding system release dbs and populating them with test fixture data"
+        )
+
+        update_coding_system_database_connections()
```
Jongmassey, line 42: Running this management command fails at this step or thereabouts for me

```sh
Migrating coding system release dbs and populating them with test fixture data
Operations to perform:
  Apply all migrations: snomedct
Running migrations:
  Applying snomedct.0001_initial... OK
  Applying snomedct.0002_historysubstitution_querytablerecord... OK
  Applying snomedct.0003_auto_20200806_1428... OK
  Applying snomedct.0004_remove_querytablerecord_and_historysubstitution... OK
Installed 438 object(s) from 1 fixture(s)
Installed 717 object(s) from 1 fixture(s)
Installed 736 object(s) from 1 fixture(s)
Operations to perform:
  Apply all migrations: dmd
Running migrations:
  Applying dmd.0001_initial... OK
Installed 13 object(s) from 1 fixture(s)
Traceback (most recent call last):
  File "/home/jongmassey/opensafely-core/opencodelists/.venv/lib/python3.12/site-packages/django/db/backends/base/base.py", line 279, in ensure_connection
    self.connect()
  File "/home/jongmassey/opensafely-core/opencodelists/.venv/lib/python3.12/site-packages/django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jongmassey/opensafely-core/opencodelists/.venv/lib/python3.12/site-packages/django/db/backends/base/base.py", line 256, in connect
    self.connection = self.get_new_connection(conn_params)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jongmassey/opensafely-core/opencodelists/.venv/lib/python3.12/site-packages/django/utils/asyncio.py", line 26, in inner
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jongmassey/opensafely-core/opencodelists/.venv/lib/python3.12/site-packages/django/db/backends/sqlite3/base.py", line 200, in get_new_connection
    conn = Database.connect(**conn_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sqlean.dbapi2.OperationalError: unable to open database file
```

I'll do some digging

opencodelists/management/commands/setup_local_dev_databases.py
Jongmassey, line 1: Is the ICD 10 fixture db loadable to make another bit of development data available?

Jongmassey (commented):
I'm having some trouble getting this running on my machine, I'll do a bit more investigating to see why that might be.

opencodelists/management/commands/setup_local_dev_databases.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if load_versioning_fixture_answer == "n":
+            return
+
+        call_command(
+            "loaddata",
+            "coding_systems/versioning/fixtures/coding_system_releases.json",
+        )
+
+        database_paths = find_dbs(settings.CODING_SYSTEMS_DATABASE_DIR)
```
Jongmassey, line 24: Do you need to delete _all_ coding system database file, or just the ones which will conflict with the `migrate` command calls below? Could you work out what the destination paths for those calls are (e.g. from the `database` kwarg) and just delete those instead?

justfile
```diff
@@ -247,6 +247,41 @@ assets-test: assets-install
...
+build-dbs-for-local-development nuclear="":
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # WARNING: Passing 'nuclear' to this just recipe will create a
+    # backup copy of your current local db, if it exists, then create a
+    # new empty core `db.sqlite3` in the root directory of your
+    # opencodelists project folder.
+    if [ -n "{{ nuclear }}" ]; then
+        echo "Nuclear option enabled: deleting db.sqlite3"
```
Jongmassey, line 260: instead of this, 
```
echo "Nuclear option enabled: moving $CORE_DB_PATH to $CORE_DB_PATH.backup"
```

after the following `if` block and before the `mv` would be the most informative

opencodelists/tests/management/commands/test_setup_local_dev_databases.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+from django.core.management import call_command
+
+from opencodelists.management.commands import setup_local_dev_databases
+
+
+def test_setup_local_dev_databases_exits_early_when_in_prod_environment(monkeypatch):
+    monkeypatch.setattr(setup_local_dev_databases, "in_production", lambda: True)
```
Jongmassey, line 8: A purist might say this doesn't quite test whether this would run in production or not. If there were some bug in `in_production()`'s interpretation of django settings then this test could give an erroneously false pass

Jongmassey, line 8:
```suggestion
from opencodelists import settings


def test_setup_local_dev_databases_exits_early_when_in_prod_environment(monkeypatch):
    monkeypatch.setattr(settings, "DEBUG", False)
```

opencodelists/management/commands/setup_local_dev_databases.py
```diff
@@ -0,0 +1,124 @@
...
+           fixtures into the Core db (you will be prompted for consent
+           to do this), deletes existing test coding system release dbs,
+           creates new coding system release dbs, populates these with
+           test fixture data, instantiates a universe of fixture
+           objects, and creates a superuser. Running this command
+           directly is not recommended, instead use `just
+           build-dbs-for-local-development`"""
+
+    def handle(self, *args, **options):
+        if in_production():
```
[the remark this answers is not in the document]
KatieB5: Patching the `DEBUG` env var and the `DEBUG` setting in `settings.py` were both tried and unfortunately didn't work. I don't have definitive proof, but from what I can tell this is due to the way the OpenCodelists app is initialized during testing: App set up seems to happen very early, i.e. right at the start of a test session. This would mean Django has already initialized the app, including reading in the env vars and settings module, before any tests run. If I change a setting in the settings module during a test, the app isn’t reinitialized to pick up that change, so the original setting e.g., `DEBUG=True`, persists. 

To try and work around this, I created a separate function—`in_production()`—which reads the current value of the `DEBUG` setting, and can be monkeypatched during testing to simulate different environments.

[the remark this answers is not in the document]
mikerkelly: `pytest-django` provides a fixture to override settings. [Docs](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings). I think that would be the more idiomatic approach.

opencodelists/tests/management/commands/test_setup_local_dev_databases.py
mikerkelly: `pytest-django` provides a fixture to override settings. [Docs](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings). I think that would be the more idiomatic approach.

PR #2432, opened 2025-04-14
Replace `bleach` with `nh3` 

### LINKED ISSUE #2169: Switch from bleach to nh3 for HTML cleaning
We use bleach to clean HTML for OpenCodelists, and [bleach is deprecated](https://github.com/mozilla/bleach/issues/698) (although is still get sporadic updates).

We switched to the [nh3 package in job-server](https://github.com/opensafely-core/job-server/issues/4254), and nh3 is maintained.

Jongmassey:
Fixes #2169 

`bleach` is no longer maintained, `ammonia` being a popular replacement along with `nh3` which are its python bindings.

We have already switched to using `nh3` for job-server.


tomodwyer (approved):
Confirmed by adding tags to the `userdocs` markdown files, and output is as expected.

PR #2438, opened 2025-04-15
Remove outdated 'We plan to support ...' message (closes #2437)

### LINKED ISSUE #2437: Remove the text stating that "We plan to support ... in the future"
# Why are we doing this?

The text is out-dated and provides too much information to users at the wrong point.

The codelist builder screen only has relevant and up-to-date information.

# What are we doing?

Removing all instances of the text stating that "We plan to support ... in the future" on the codelist builder screen.

---

lyutok:
Removed outdated message "We plan to support ..." from the UI, as per issue #2437.

Closes #2437

### DIFF
--- a/assets/src/scripts/components/SearchForm.tsx
+++ b/assets/src/scripts/components/SearchForm.tsx
@@ -54,10 +54,6 @@ export default function SearchForm({
             Otherwise, searching will return all concepts with a description
             containing the search term.
           </p>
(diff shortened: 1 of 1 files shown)


lyutok:
Hello, I've removed the outdated message about supporting boolean search operators. This was related to issue #2437. Let me know if anything needs to be changed. Thanks.

Issue #2393, opened 2025-04-02
Dokku deployments: Improve data migrations reliability


lucyb:
# Why are we doing this?

[Dokku recommends](https://dokku.com/docs/advanced-usage/deployment-tasks/) that database migrations are done during deployment but before an app is deployed. They recommend that the migrations are applied during the "release" phase.

We recently had an [outage](https://docs.google.com/document/d/1L9qMOAOe65Fwl7pHkbbCJ0otRsOEQvNHc-O68U-6rvg/edit?tab=t.0#heading=h.1ytdczc27rgj) where migrations were not fully applied during a deployment and the app needed to be redeployed in order for the migrations to be applied fully.

# How will we know when it's done?

Migrations can be successfully applied during the release phase rather than during app start up.

# What are we doing?

Moving the application of Django migrations to the `Procfile` : `release` and removing them from the entrypoint script that's run on container startup.

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

lucyb:
I've created a branch for this work [here](https://github.com/opensafely-core/opencodelists/pull/2457). I'm not sure when I'm going to have time to pick it up again, so would be happy for someone else to continue the work.

PR #2457, opened 2025-04-22
Move migrations to Procfile

lucyb:
[Dokku recommends](https://dokku.com/docs/advanced-usage/deployment-tasks/) that database migrations are not done _during_ deployment, but _before_ an app is deployed. They recommend that the migrations are applied during the "release" phase. This should stop deployments failing due to health checks timing out during a long running migration.

Refs: #2393

TODO:
- [ ] test change works with dokku
- [x] figure out how to test migrations in CI (EDIT: this is now being done during unit testing of the dev image and during the smoke test of the prod image)
- [x] ensure migrations are still run in local dev (EDIT: this change only affects the prod image, the dev image is unchanged)

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+web: /opt/venv/bin/gunicorn --config /app/deploy/gunicorn/conf.py opencodelists.wsgi
+release: /app/deploy/prod.sh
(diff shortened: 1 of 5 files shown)


lucyb:
I've tried to test this change using Steve's [PR](https://github.com/opensafely-core/opencodelists/pull/2487) to run it in devcontainers (with some small tweaks). However, this isn't pulling in the Profile, so I don't know if that's configured correctly (which is the part that I really want to test).

PR #2458, opened 2025-04-22
Ignore local dev artifacts in Docker image

lucyb:
This ignores artifacts that are likely to be in a development environment. This shouldn't be an issue for production images, because they're built in a clean CI environment, however it can cause issues with large docker images when building locally.


lucyb:
>Is there not a circumstance when one might want the core and coding system DBs included when running a locally-built docker image?

Yes, you're right. I'll close this and rethink the approach.

PR #2460, opened 2025-04-23
Replace tree view React app with HTML

### LINKED ISSUE #2427: Replace the tree view app with HTML
# Why are we doing this?

The current [Tree view on a codelist page](https://www.opencodelists.org/codelist/primis-covid19-vacc-uptake/preg/v2.5/#tree) uses a modified version of the CodelistBuilder app to display the tree. Based on the fact this tree does not require user interaction, other than opening and closing sections, we could replace this with standard HTML elements including the `<details>` element.

This would mean that the Builder could be made more specific to the requirement of building, and speed up the page load of the tree view.

When the CodelistBuilder app is no longer used for the Tree view.

# What are we doing?

- Use HTML elements to build the tree
- Replace any references to the Tree view in the builder app

---

tomodwyer:
Closes #2427

🚧


tomodwyer:
Replaced by https://github.com/opensafely-core/opencodelists/issues/2466

Issue #2451, opened 2025-04-17
Update search box to toggle searches by term OR clinical code


tomodwyer:
# Why are we doing this?

Currently there is help text under the search box explaining that users can search for a clinical term, or they can enter `code:` to search for a code. This is confusing, and requires a user to read, parse, and understand the required syntax to search for a code.

We should change this to add a radio button which toggles the user between searching for a term or a code, and we can then prepend `code:` to the code searches.

# How will we know when it's done?

When users can search for a term or code by selecting which type of search they wish to perform.

# What are we doing?

- Add radio buttons to allow for searching by term or code
- If the user is searching for a term, pass the term to the search
- If the user is searching for a code, prepend the search with `code:` removing any leading spaces in their search term
- When the user clicks on a previous search, restore the state of the search box and radio button to the state it was when they made that search

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

rw251:
This should wait until #2465 is merged as it affects the same part of the UI.

That PR enforces search length. Currently it enforces the model string lengths (255 for search terms, 18 for code searches) on the server side, but only 255 on the client-side. When we add the term/code toggle here, we should also make the client-side validation different for codes - limiting to 18 characters and changing the warning message.

PR #2465, opened 2025-04-24
Enforce character length for search terms and codes on client and server side

### LINKED ISSUE #2434: Add validation to limit the amount of text inputted into the search box
# Why are we doing this?

We want to give users a good search experience and we don't want extremely long strings in the database.

It's currently possible to add an arbitrary length string field into the search criteria. Although we've set a max length of 255 on the Django model, it doesn't seem as though this is being enforced by the database.

It's not possible for an accidental or malicious user to input a string that is too long into the search results.

# What are we doing?

Adding client and server-side validation to prevent overly long strings being searched for.

We'll need to decide how long the string should sensibly be. At most it should be 255 characters, but it could be less if that makes sense.

We should also provide a sensible error to users, telling them what they need to do differently.

---

rw251:
Fixes #2434 

This adds client-side and server-side validation to ensure searches (both codes and terms) are <= the `max_length` of the underlying `Search` model (currently 255 chars for search terms, and 18 chars for code searches). Displays a warning message like this when you're at the limit.

image

The reason the `max_length` property was not enforced server side is because we currently create a new Search object with `draft.searches.get_or_create()` which bypasses model validation. This change means we now validate the submitted term or code against the model definition.

This enforces term and code lengths server-side, but only 255 characters client-side. This can be changed as part of #2451 and I've left a note there to that end.


builder/views.py
```diff
@@ -213,7 +214,7 @@ def update(request, draft):
 @require_http_methods(["POST"])
 @load_draft
 @require_permission
-def new_search(request, draft):
+def new_search(request: HttpRequest, draft: CodelistVersion):
```
Jongmassey, line 217: Although I'm generally a fan of type hits, AFAICT we're not using them anywhere else in this codebase so this driveby feels slightly out of place.

Jongmassey (approved):
I know very little about React so I can only say that from testing it locally it seems to work. My inclination would be to do this using a CSS `:invalid` pseudo-selector but that would probably be more difficult than this!

builder/views.py
rw251: Yeah fair enough. It was nice while I was working on it but probably should be part of a wider piece of work if we go down that route. I'll get rid.

PR #2473, opened 2025-04-29
Completion of the migration squash

rw251:
The first two PRs in this series squashed all the migrations:

- #2463 
- #2467 

Once this has been applied on production and for each local developer, we can do the last step which is to delete all the old migrations, and a couple of other tidy up operations.


mikerkelly (approved):
> Once this has been applied on production and for each local developer, we can do the last step which is to delete all the old migrations, and a couple of other tidy up operations.

I think this PR is "the last step" described here, is that right? Then I'm approving but expecting you'll ask local developers to update before merging this. Or I think it would be equally fine to merge this, but have a message in Slack to developers telling them which commit they need to roll back to to apply the migrations in.

Please do describe explicitly in PR descriptions the specifics of what the PR has done and why that's the right thing to do. I think you're following the steps described here to complete the process of migration squishing:

> https://docs.djangoproject.com/en/5.2/topics/migrations/#migration-squashing
> You must then transition the squashed migration to a normal migration by:
> - Deleting all the migration files it replaces.
> - Updating all migrations that depend on the deleted migrations to depend on the squashed migration instead.
> - Removing the replaces attribute in the Migration class of the squashed migration (this is how Django tells that it is a squashed migration).

This means these migrations can be squished again in future and leaves us with no more work to do here.

Issue #2446, opened 2025-04-17
Move codelist builder metadata under the title


tomodwyer:
This is static content that does not need to stay in the sidebar, which is an area where users interact with elements whilst creating their codelist.

## Screenshot example

![](https://github.com/user-attachments/assets/4a7a420f-c7a3-4f1d-a2c6-87e4feac1997)

PR #2474, opened 2025-04-29
Move metadata under title in builder

tomodwyer:
Closes [#2446](https://github.com/opensafely-core/opencodelists/issues/2446)

Metadata is currently shown in the sidebar, but we are moving to the sidebar being for searching and interacting with the codelist. For this reason, we have moved the metadata under the title.

It also means we can create a `<Metadata />` component, and add tests with 100% coverage for the component.


KatieB5 (approved):
This looks good to me. Thank you Thomas!

PR #2477, opened 2025-04-29
Adds a search box toggle - search by term OR seach by code

### LINKED ISSUE #2451: Update search box to toggle searches by term OR clinical code

rw251:
Fixes #2451

New search box now looks like:

image

All help text (about searching with `code:` prefixes is now gone), except for the bit that tells you that you can do wildcard searching for ICD-10 codes:

image

### DIFF
--- a/assets/src/scripts/components/SearchForm.tsx
+++ b/assets/src/scripts/components/SearchForm.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { Button, Form } from "react-bootstrap";
+import { Button, Card, Form, InputGroup } from "react-bootstrap";
 import { getCookie } from "../_utils";
 import { PageData } from "../types";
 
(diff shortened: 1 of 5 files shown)


assets/src/scripts/components/SearchForm.tsx
```diff
@@ -12,77 +12,121 @@ export default function SearchForm({
   codingSystemName,
   searchURL,
 }: SearchFormProps) {
-  const [searchTerm, setSearchTerm] = useState("");
-
   // NB, if you change the max search length, remember to change it on
   // the server side as well (codelists/models.py - Search - term)
   const MAX_SEARCH_LENGTH = 255;
+  const MAX_CODE_LENGTH = 18;
```
mikerkelly, line 18: I think we should change the breadcrumb in the comment on the preceding lines to reflect that now both the term and the code search length are enforced in the React app (`assets/src/scripts/components/SearchForm.tsx`) as well as in the Django model code (`codelists/models.py`). And we should change it in `codelists/models.py` too, as that also only refers to the term, not the code. These were added in #2465.

tomodwyer (approved):

PR #2495, opened 2025-05-06
Make remaining coding systems tests independent

### LINKED ISSUE #2301: Coding system tests are not isolated
It's possible to run one subset of tests, such as the BNF tests and have them fail. It's necessary to run all these tests in one session, to get them to pass.

This is because test fixtures are loaded into a database that persists across all of the tests in a run.

See discussion in #2115.

### LINKED ISSUE #2493: Make `test_import_data_utils` run independently
# Why are we doing this?

Two `coding_systems/base/tests/test_import_data_utils.py` test cases can be run successfully along with other test cases, but fail when run individually, which makes them harder to maintain. We should fix this.

# What are we doing?
 
- `coding_systems/base/tests/test_import_data_utils.py::TestSaveCodelistDraftUpdatesCompatibilityMultipleReleases::test_save_codelist_draft_updates_compatibility_multiple_releases`
- `coding_systems/base/tests/test_import_data_utils.py::TestSaveCodelistDraftUpdatesCompatibility::test_save_codelist_draft_updates_compatibility`

These test cases are using the `DynamicDatabaseTestCase` approach added in #2115. Running them individually results in an exception like:

```
django.test.testcases.DatabaseOperationForbidden: Database threaded connections to 'bnf_import-data_20221101' are not allowed in this test. Add 'bnf_import-data_20221101' to coding_systems.base.tests.test_import_data_utils.TestSaveCodelistDraftUpdatesCompatibility.databases to ensure proper test isolation and silence this failure.
```

... this is exactly the issue that the `DynamicDatabaseTestCase` approach attempts to resolve. They do have the named database configured, it's not clear why the mechanism works when run alongside other tests but not when run independently. Maybe other tests have fixtures or other side effects that are somehow beneficial.

---

StevenMaude:
Fixes #2493 and fixes #2301. See #2403.

This facilitates running two tests that were not made independently-runnable in #2492.

A slightly earlier version of this PR was tested with all tests running independently, except for the database timeout which was changed in case there were issues with it. (Which was mentioned in the comments for the helper script to run the tests independently.)

The changes made since then are essentially comment/commit message changes and rebasing.

### DIFF
--- a/coding_systems/base/tests/dynamic_db_classes.py
+++ b/coding_systems/base/tests/dynamic_db_classes.py
@@ -27,15 +27,13 @@ class DynamicDatabaseTestCase(TestCase):
     """
 
     @property
-    def db_alias(self):
-        """The database alias that needs to be accessed in this test class, str."""
-        raise NotImplementedError(
-            "This test class requires a database alias to be set."
-        )
+    def db_aliases(self):
+        """The database aliases that need to be accessed in this test class, list[str]."""
+        raise NotImplementedError("This test class requires db_aliases to be set.")
 
     import_data_path = None
 
-    def add_to_testcase_allowed_db_aliases(self, *db_aliases):
+    def add_to_testcase_allowed_db_aliases(self, db_aliases):
         """Add database names to the test instance's allowed database aliases.
 
         This mutates the *class* state by changing the `databases` attribute.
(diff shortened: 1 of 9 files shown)


coding_systems/base/tests/dynamic_db_classes.py
```diff
@@ -27,15 +27,13 @@ class DynamicDatabaseTestCase(TestCase):
...
 
     @property
-    def db_alias(self):
-        """The database alias that needs to be accessed in this test class, str."""
-        raise NotImplementedError(
-            "This test class requires a database alias to be set."
-        )
+    def db_aliases(self):
+        """The database aliases that need to be accessed in this test class, list[str]."""
+        raise NotImplementedError("This test class requires db_aliases to be set.")
```
StevenMaude, line 32: thought: As mentioned in the commit message, I changed this because I thought it was required to solve the problem. It's not, but it allows making the database dependencies more obvious in the updated tests here, so it may be worth retaining.

PR #2498, opened 2025-05-07
Move Codelist Builder form buttons to the top-right

### LINKED ISSUE #2447: Move codelist management form buttons to the top-right
As this is a pattern we use across OpenSAFELY products, and means the sidebar is clearly for searching and filtering.

tomodwyer:
Closes #2447

### DIFF
--- a/assets/src/scripts/__tests__/components/ManagementForm.spec.tsx
+++ b/assets/src/scripts/__tests__/components/ManagementForm.spec.tsx
@@ -174,9 +174,9 @@ describe("ManagementForm", () => {
     });
     await user.click(discardButton);
 
-    const discardDraftButton = screen.getByText("Discard draft");
-    expect(discardDraftButton.closest("form")).toContainElement(
-      screen.getByDisplayValue("discard"),
-    );
+    const discardDraftButton = screen.getByRole("button", {
+      name: "Discard draft",
+    });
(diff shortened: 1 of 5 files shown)


assets/src/scripts/__tests__/components/ManagementForm.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+    // Modal form action
+    const discardButton = screen.getByRole("button", {
+      name: "Discard",
+    });
+    await user.click(discardButton);
+
+    const discardDraftButton = screen.getByText("Discard draft");
+    expect(discardDraftButton.closest("form")).toContainElement(
+      screen.getByDisplayValue("discard"),
+    );
```
rw251, line 180: Am I right that this is basically testing that: The form containing the "Discard draft" button, also contains an element with a `value` of `discard`? If so, what's the reasoning for doing it this way, rather than by checking the button with the `toHaveAttribute` method as with the other buttons in this test?

rw251 (approved):

assets/src/scripts/__tests__/components/ManagementForm.spec.tsx
tomodwyer: That was a mistake, this was my initial way of testing the form, which I then updated to use the `toHaveAttribue`, but missed this last one. I'll fix it now.

Issue #2450, opened 2025-04-17
Update the "searches" component to a card


tomodwyer:
So that it matches the other card components, makes it more obvious these are clickable items, and makes clear that you can remove a search.

## Screenshot example

![](https://github.com/user-attachments/assets/6b2e8d70-8a16-4749-9136-c36a67407071)

rw251:
Closed by #2500

PR #2500, opened 2025-05-07
Make the "previous searches" component into a Card

rw251:
As per this ticket (#2450) this changes the previous searches component into a Card element. It now looks like this:
image

and like this when a search is selected:
image

### DIFF
--- a/assets/src/scripts/components/Search.tsx
+++ b/assets/src/scripts/components/Search.tsx
@@ -53,7 +53,7 @@ export default function Search({ draftURL, searches }: SearchProps) {
                     type="submit"
                     size="sm"
                     variant={
-                      search.url === activeUrl ? "light" : "outline-primary"
+                      search.url === activeUrl ? "light" : "outline-danger"
                     }
                   >
                     Remove
--- a/assets/src/scripts/components/Search.tsx
+++ b/assets/src/scripts/components/Search.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Button, Form, ListGroup } from "react-bootstrap";
+import { Button, Card, Form, ListGroup } from "react-bootstrap";
 import { getCookie } from "../_utils";
 import { PageData } from "../types";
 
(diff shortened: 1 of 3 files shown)


assets/src/scripts/components/Search.tsx
```diff
@@ -57,14 +71,14 @@ export default function Search({ draftURL, searches }: SearchProps) {
         {searches.some((search) => search.active) ? (
           <ListGroup.Item
             action
-            className="py-1 px-2 font-italic"
+            className="font-italic"
             href={encodeURI(draftURL)}
+            onClick={handleClick(draftURL)}
```
tomodwyer, line 76: I don't think we need the onClick handler here

assets/src/scripts/components/Search.tsx
```diff
@@ -9,18 +9,31 @@ interface SearchProps {
...
             {search.delete_url ? (
               <ListGroup.Item
                 action
-                active={search.active}
-                className="py-1 px-2"
+                active={search.url === activeUrl}
                 href={search.url}
+                onClick={handleClick(search.url)}
               >
                 <Form action={search.delete_url} method="post">
```
tomodwyer, line 38: I'd like to get rid of the float for the button, so I'd recommend the following CSS:

```jsx
<Form
  action={search.delete_url}
  className="d-flex justify-content-between align-items-center"
  method="post"
>
```

```jsx
<Button
  aria-label="remove search"
  className="ml-2"
  name="delete-search"
  type="submit"
  size="sm"
  variant={
    search.url === activeUrl ? "light" : "outline-primary"
  }
>
```

assets/src/scripts/components/Search.tsx
rw251: My rationale was that if you click on "show all" and don't use the onClick handler, then until the page reloads the "show all" _and_ the previous search are both highlighted in blue. Putting it here means the highlighted item updates immediately which I think looks a bit better. If we keep it like this, there's an argument to add this click logic to the included/unresolved filter card as well.

tomodwyer: Ah ok, that makes sense 👍 

Another side effect of us having the mixed nature of the React app and the Django app storing state.

tomodwyer (approved):

PR #2511, opened 2025-05-09
Update the sidebar and versions component styles

### LINKED ISSUE #2496: Update the Versions component to a card
So that it matches the other sidebar components.

It should also use the same "draft" button styling as in the title.

tomodwyer:
Closes #2496

### DIFF
--- a/assets/src/scripts/__tests__/components/Version.spec.tsx
+++ b/assets/src/scripts/__tests__/components/Version.spec.tsx
@@ -1,0 +1,89 @@
+import "@testing-library/jest-dom";
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import { describe, expect, it } from "vitest";
+import Version from "../../components/Version";
+import { VersionT } from "../../types";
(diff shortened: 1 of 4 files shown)


rw251:
I'm not sure I like the solid grey background for the currently selected draft version. Partly because it makes me think it's disabled (which I guess it is because you can't click it - but "selected" seems a bit different to "disabled"), but also because it's different to the other components which are blue when selected. I'll approve anyway, but just flagging in case you want to change it.

We should probably add a ticket to make this same change to the published and under review screens. They're not part of the builder, but it's now a bit jarring when you click on a non-draft version and the style of the Version component changes.

rw251 (approved):

tomodwyer:
@rw251 Agreed on the colouring, ideally I wanted something subtle, but Bootstrap doesn't offer that.

I think it would be good to see from users if they actually use the version card as it is. I don't think linking back to old versions from the builder is actually that useful.

PR #2512, opened 2025-05-09
Create an `is_empty_codelist` variable and pass to the front end

### LINKED ISSUE #2468: Provide a value to the React app if the codelist builder is empty
# Why are we doing this?

Expanding on https://github.com/opensafely-core/opencodelists/pull/2424

The React app has a confusing way of checking for a blank codelist, so that it can provide the relevant instructions to a user. This was flagged in the comment:

> I think I'd prefer this to be server side - but maybe just something to replace the `resultsHeading` string comparison as it's a bit odd to pass a heading string to the front end, only to replace it conditionally with something else. Maybe just a `has_no_codes` server flag that's set at the same time that heading string is set?

_Originally posted by @rw251 in https://github.com/opensafely-core/opencodelists/pull/2424#discussion_r2060313715_
            
When the React app is only checking a boolean to know whether to show the blank codelist instructions or not.

# What are we doing?

- [ ] Create a server-side variable for a blank codelist
- [ ] Pass it to the React app
- [ ] Only show the `<EmptyState />` component based on the state of the variable

---

rw251:
Fixes #2468 

Defines an `is_empty_codelist` variable so the front end can display the empty state and help message.

Also
- Added a better heading in the situation where you have 1 or more searches, but no matching codes. Previously the heading said "Start building your codelist by searching for a term or a code", but it now more accurately says "None of your searches match any concepts"
- Drive-by addition to the heading to include the number of displayed codes


tomodwyer (approved):
Happy with the JS 👍

PR #2533, opened 2025-05-13
Define builder-coding system interface

Jongmassey:


codelists/views/version.py
```diff
@@ -14,7 +16,7 @@ def version(request, clv):
     code_to_term = None
     parent_map = None
     tree_tables = None
-    if clv.coding_system_id in ["bnf", "ctv3", "icd10", "snomedct"]:
+    if issubclass(clv.coding_system.__class__, BuilderCompatibleCodingSystem):
```
mikerkelly, line 19:
```suggestion
    if isinstance(clv.coding_system, BuilderCompatibleCodingSystem):
```

I think checking `isinstance` is sufficient, more idiomatic, and easier to read than accessing `__class__`, here.

Jongmassey: I can't believe I missed the bleeding obvious there :facepalm:  thanks!

coding_systems/base/coding_system_base.py
```diff
@@ -92,21 +92,34 @@ def release(self):
...
+    Represents a CodingSystem that will never have an associated database.
+    """
+
+    has_database = False
+
+    def code_to_term(self, codes):
+        return {code: "unknown" for code in codes}
+
+
+class BuilderCompatibleCodingSystem(BaseCodingSystem):
```
mikerkelly, line 113: Looks good. Could we have a docstring for each functions like the useful `matching_codes` docstring explaining what they do and how they're used? And briefly for the class itself? I think that would make this much more useful for understanding how the builder uses a coding system.

mikerkelly (approved):

coding_systems/base/coding_system_base.py
Jongmassey: Yeah, that'd be an improvement. I was maybe in a bit of a rush to unblock tickets that depends on this. I'll spend the time to document this a bit better now.

PR #2539, opened 2025-05-14
Update codelists version page to match builder changes

### LINKED ISSUE #2526: Update the Codelist version page to match the Builder changes
Many changes have been made to the Builder page, but not the "codelist viewer" page. We should update the styling on that page to match the updates to the builder.

KatieB5:
Closes #2526 

### Style changes made to codelists version page

[ ] Restyled and moved the create new/update, publish and delete version buttons to the top right of the page
[ ] Updated sidebar styling
[ ] Moved the codelist metadata to sit under the title
[ ] Added coloured badges for codelist status to "Versions" component

#### Builder draft page
image

#### Codelists version page
_Under review codelist_
image

_Published codelist_
image

_Published codelist with an organisation_
image

### DIFF
--- a/assets/src/styles/base.css
+++ b/assets/src/styles/base.css
@@ -102,7 +102,6 @@ a:focus {
 
 #codelist-builder-container {
   overflow-x: hidden;
-  padding: 0.25rem;
 }
 
 .builder__sidebar {
--- a/assets/src/styles/base.css
+++ b/assets/src/styles/base.css
@@ -110,3 +110,9 @@ a:focus {
   flex-direction: column;
   row-gap: 1rem;
 }
+
+.version__sidebar {
+  display: flex;
+  flex-direction: column;
+  row-gap: 2rem;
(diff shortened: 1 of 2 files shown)


rw251:
Looks great. Two minor things:

1. When you toggle between a draft and a published version, the page shifts slightly. Getting rid of the padding here (which I don't think we need) should sort it: https://github.com/opensafely-core/opencodelists/blob/d2b6313973e993265632a386968d7b42e90df065/assets/src/styles/base.css#L103-L106
2. Not for now, but we now have two different versions of the Versions card. I much prefer the styling of the one you've just updated (I prefer the lighter grey highlight, and the extra info). But I think I've talked to Thomas about whether we even want that card on the draft page at all. So in a future ticket we should consider that, and either remove it, or bring its style in line with the one on the in-review and published pages.

rw251 (approved):

PR #2540, opened 2025-05-14
Better help and layout on the "no results" page in the builder

### LINKED ISSUE #2522: Improve the layout of the "no results" page
## Why are we doing this?

The current layout of this page is confusing. The main section contains no information, but there is a banner at the top, and the title giving the user two slightly different messages.

Screenshot of the current layout

## How will we know when it's done?

When the page provides the user with clear instructions on what has happened, and what they can do next.

## What are we doing?

- [ ] Remove the Django built-in banner
- [ ] Update the title and text in the React app
- [ ] Provide the user with next steps
- [ ] What do we do about a search with no results being added to the sidebar?

---

rw251:
If you do a search that returns no results we now:

- display some helpful text so it's both clearer that nothing was found, but also what you might want to do next
- no longer display the banner at the top as it duplicates info elsewhere

Now looks like this:

image

Fixes #2522

### DIFF
--- a/assets/src/scripts/components/EmptySearch.tsx
+++ b/assets/src/scripts/components/EmptySearch.tsx
@@ -1,0 +1,24 @@
+import React from "react";
+
+export default function EmptyState() {
+  return (
+    <section style={{ maxWidth: "80ch" }}>
+      <p className="lead">Your search returned no concepts. You may want to:</p>
+      <ol>
+        <li>Double check the spelling</li>
+        <li>
+          Make sure that if you're searching for a code (rather than a search
+          term) that you have selected the correct option ("Search by code"
+          above the search box on the left)
+        </li>
+        <li>Try searching for a similar term</li>
+      </ol>
+      <p>
+        If this search was an error, you can click "Remove" from the "Previous
+        searches" box on the left to get rid of it. However, if you want other
+        people to know that you tried this search, but it returned no results,
+        then you can leave it as a blank search.
+      </p>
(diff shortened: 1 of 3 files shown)


assets/src/scripts/components/EmptySearch.tsx
```diff
@@ -0,0 +1,24 @@
...
+          above the search box on the left)
+        </li>
+        <li>Try searching for a similar term</li>
+      </ol>
+      <p>
+        If this search was an error, you can click "Remove" from the "Previous
+        searches" box on the left to get rid of it. However, if you want other
+        people to know that you tried this search, but it returned no results,
+        then you can leave it as a blank search.
+      </p>
```
mikerkelly, line 21:
```suggestion
        <li>Check your spelling</li>
        <li>Try searching for a similar term</li>
        <li>
            Ensure you have selected the correct option above the search box
            ("Search by code" if searching by code rather than term)
        </li>
      </ol>
      <p>
        If this search was an error, click "Remove" in the "Previous searches"
        box on the left to delete it. If you'd like to record that you
        performed this search but it returned no results, you can leave it as
        is.
      </p>
```
- avoiding "double check" and "get rid of" as slightly tricky for non-native speakers
- putting the two bullets about term-based search first and together
- slight rephrasings for clarity / brevity

rw251: Very nice - thanks!

Issue #2466, opened 2025-04-25
Simplify the Tree view application


tomodwyer:
Replaces #2427 
Makes #2460 now redundant

[🧵 Slack thread 🧵 ](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1745574262654459)

## Why are we doing this?

As part of the discussion in to #2427 it was identified that recursive loading of templates in Django can be slow. On this point, @tomodwyer investigated whether a simplified React app would be faster. Full details, as sent in Slack are below.

## How will we know when it's done?

The tree view on a Codelist page contains the updated styles for a tree, loaded from the simplified tree data structure JSON.

## What are we doing?

- [x] https://github.com/opensafely-core/opencodelists/pull/2551
- [ ] https://github.com/opensafely-core/opencodelists/pull/2559

## Investigation notes

As discussed in stand up, after the presentation yesterday when Peter mentioned about the slow nature of recursive templates in Django, I wondered if actually having a very small React app utilising the `<details>` element and recursively loading the tree in JS might be faster.

To test this, I've made a small React app version, and switched my local development mode to run as close to production as possible. Results are in the screenshots below, but the key parts are:

1. The HTML version creates a much larger document, 3.3 mB vs the 599 kB size of the page which loads React
    1. This means that even with all the assets, the React app page is **75% smaller** for the user to download
1. The initial HTML page load for the **React version is 173ms faster**
1. The overall page load time for the **React version is 560ms faster**
1. The overall size of the React app is 142 lines of code (without any tests added)
1. Without any specific caching, and without allowing the browser cache, both versions get significantly faster on second load (attached screenshots)

### Screenshots

This was a clean start of the app with caching disabled in Chrome DevTools.

#### HTML version - initial load

![Image](https://github.com/user-attachments/assets/18f35f8a-184c-43cc-ad73-63d548b94a31)

#### HTML version - second load

![Image](https://github.com/user-attachments/assets/52bd46fd-cac7-4a51-88a0-a86d05c1cae5)

#### React app - initial load

![Image](https://github.com/user-attachments/assets/40ab0555-0870-404f-aa39-31d707a8ffed)

#### React app - second load

![Image](https://github.com/user-attachments/assets/83a845ba-686e-4b4a-bb13-744bd0dd380c)

tomodwyer:
Closed by https://github.com/opensafely-core/opencodelists/pull/2559

PR #2551, opened 2025-05-15
Construct tree view object server side and pass to front end as json script

rw251:
This is the server side part of #2466.

We construct a dictionary server side in the format that will be expected by the React component, and pass it as a json object to the view.

### DIFF
--- a/codelists/tests/test_tree_data.py
+++ b/codelists/tests/test_tree_data.py
@@ -1,0 +1,228 @@
+"""Tests for tree data building functionality."""
+
+import pytest
+
+from codelists.tree_data import build_tree_data
+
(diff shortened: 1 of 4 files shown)


Jongmassey (approved):
After much thinking and experimenting about how this might be better placed somewhere else in the codebase I've come round to thinking how this is is perfectly good. Does this make `tree_tables` redundant in the front end code now?

rw251:
I think it removes the need for _all_ the things that are parameters to `build_tree_data` to be sent to the front end as soon as the front end changes are complete. @tomodwyer knows this and will tidy up during or after his front end PR for this.

PR #2560, opened 2025-05-16
Update the styles and text of the "Create a new codelist" form

### LINKED ISSUE #2412: Improve the "create new codelist" screen
# Why are we doing this?

To improve the user experience of creating a new codelist.

The suggested changes below are implemented.

# What are we doing?

The screen currently looks like this:

Image

It would be good to:
- provide examples of good (and bad?) codelist names
- provide guidance on each available coding system to assist with user selection (e.g. SNOMED for GP data, ICD10 for hospital, BNF for meds)
- a better flow might be to have the first thing be a question of whether you want to “create codelist from scratch” or “create codelist by uploading a CSV file”, then make the form adjust to just ask what is needed for each one. This avoids the "optional" upload CSV file box potentially confusing people. It also makes clear from the very start what the two options are.

---

tomodwyer:
Closes #2412


templates/opencodelists/user_create_codelist.html
```diff
@@ -3,12 +3,89 @@
...
+      {% endfragment name_help_text %}
+      {% form_input field=form.name help_text=name_help_text %}
+    </div>
+
+    <div class="bg-white px-6 py-12 shadow-sm sm:rounded-lg sm:px-12">
+      {% fragment as coding_system_help_text %}
+        <ul>
+          <li><strong>CTV3 (Read v3)</strong>: Clinical events, drugs, and appliances (historic; Primary and Secondary Care)</li>
+          <li><strong>ICD10</strong>: Internationally standardised diagnosis codes diagnosis codes (mainly Secondary Care and Death Registry)</li>
+          <li><strong>Psuedo-BNF</strong>: Primary Care prescribing (historic; can be converted to NHS Dictionary of Medicines and Devices)</li>
```
Jongmassey, line 57: I'm not sure Pseudo-BNF can be described as historic. It's still being actively updated and used in various NHS processes.

opencodelists/forms.py
```diff
@@ -49,6 +49,7 @@ class CodelistCreateForm(forms.Form):
         for id, system in CODING_SYSTEMS.items()
         if issubclass(system, BuilderCompatibleCodingSystem)
     )
+    CODING_SYSTEM_CHOICES.sort(key=lambda x: x[1].lower())
```
Jongmassey, line 52: This would probably be better implemented as a change to the previous statement rather than re-sorting an already-sorted list. There's no need for the intermediate sorting

templates/opencodelists/user_create_codelist.html
```diff
@@ -3,12 +3,89 @@
...
+        <p>For example: <strong>"Type 2 diabetes without retinopathy"</strong> is a better name than <strong>"diabetes"</strong>.</p>
+      {% endfragment name_help_text %}
+      {% form_input field=form.name help_text=name_help_text %}
+    </div>
+
+    <div class="bg-white px-6 py-12 shadow-sm sm:rounded-lg sm:px-12">
+      {% fragment as coding_system_help_text %}
+        <ul>
+          <li><strong>CTV3 (Read v3)</strong>: Clinical events, drugs, and appliances (historic; Primary and Secondary Care)</li>
+          <li><strong>ICD10</strong>: Internationally standardised diagnosis codes diagnosis codes (mainly Secondary Care and Death Registry)</li>
```
Jongmassey, line 56:
```suggestion
          <li><strong>ICD10</strong>: Internationally standardised diagnosis codes (mainly Secondary Care and Death Registry)</li>
```

templates/opencodelists/user_create_codelist.html
```diff
@@ -3,12 +3,89 @@
...
+        <p>You should use sentence case and spaces between words, rather than underscores or hyphens. Being descriptive is more important than being concise.</p>
+        <p>For example: <strong>"Type 2 diabetes without retinopathy"</strong> is a better name than <strong>"diabetes"</strong>.</p>
+      {% endfragment name_help_text %}
+      {% form_input field=form.name help_text=name_help_text %}
+    </div>
+
+    <div class="bg-white px-6 py-12 shadow-sm sm:rounded-lg sm:px-12">
+      {% fragment as coding_system_help_text %}
+        <ul>
+          <li><strong>CTV3 (Read v3)</strong>: Clinical events, drugs, and appliances (historic; Primary and Secondary Care)</li>
```
Jongmassey, line 55: Can't help feeling that these descriptions would be better implemented as properties of their respective coding system classes (as part of the `BuilderCompatibleCodingSystem` interface) rather than hard coding them in the template.

Jongmassey (commented):
There's a few things that I think need changing. Some important (wording), others are implementation details I feel moderately strongly about.

templates/opencodelists/user_create_codelist.html
KatieB5: Suggest to use ICD-10 (hyphenated) here, as this is [what the WHO use](https://iris.who.int/handle/10665/246208), and we use this format elsewhere on OpenCodelists.

Jongmassey: sounds good. The [`name`](https://github.com/opensafely-core/opencodelists/blob/4fdc23d8292b1815a463186019c3d18d46183793/coding_systems/icd10/coding_system.py#L12) of this coding system is "ICD-10", I suggest we use this property instead, per [this commit](https://github.com/opensafely-core/opencodelists/commit/c139706893c7eb0a99196abf53cc482f978c85f9)

opencodelists/forms.py
tomodwyer:
```py
    CODING_SYSTEM_CHOICES = [("", "---")] + sorted(
        [
            (id, system.name)
            for id, system in CODING_SYSTEMS.items()
            if issubclass(system, BuilderCompatibleCodingSystem)
        ],
        key=lambda x: x[1].lower(),
    )
```

Is this what you mean?

Jongmassey: yep :)

Jongmassey:
Screenshots after adding f8b3e03

image

image

Jongmassey (approved):

PR #2573, opened 2025-05-20
Set up plumbing to allow images in the docs to be generated by playwright

### LINKED ISSUE #2514: Use PlayWright to generate screenshots for the docs
# Why are we doing this?

Keeping the docs up to date with the latest screenshots is tricky because it is not always obvious when a UI change needs reflecting in the docs. It is easy to get out of sync. Airlock uses PlayWright to generate all screenshots used in the docs.

The screenshots in the docs will be up to date, and easily maintained.

# What are we doing?

Creating a single PlayWright test that navigates the site taking screenshots of relevant pages and components which are then referenced in the docs.
---

rw251:
Fixes #2514 

Implements a playwright test that creates images via screenshot for the docs.

The screenshot test runs as part of the functional tests, but only actually produces screenshot files if you set the `TAKE_SCREENSHOTS` env variable to `True`. This is so the test can always run, and therefore can be fixed if code changes break it, but does not lead to screenshot churn where minor changes (like date stamps) cause lots of images to change and appear as changes according to git. This is in line with how this is done in the airlock codebase.

This implements 5 screenshots. I've raised #2574 to cover the creation of the rest.


Jongmassey (approved):
Seems to work fine for me and we get a bonus screenshot updated to the new visual style

PR #2579, opened 2025-05-21
Expand dm+d test fixture data

Jongmassey:
The current dm+d test coding system fixture is extremely minimal and inadequate for the forthcoming tasks to make the `dmd` coding system a `BuilderCompatibleCodingSystem`.

In order to keep in the theme of "asthma medication", this PR adds subset of the salbutamol tree to the dmd fixture:

- Add "salbutamol" VTM and Ing,
- plus 10 descendant VMPS,
- plus 10 descentant AMPS,
- plus all supporting referenced entities.

This should give adequate data to test the search and hierarchy methods required to satisfy the
`BuilderCompatibleCodingSystem` interface and give reasonable numbers of entities to test template rendering for e2e playwright tests without becoming unreasonably large on disk or intractable.

JSON chunks were added by navigating the model tree from the "salbutamol" VTM and running commands of the form
```py
json.dumps([{"model": "dmd.amp", "pk":m.id,"fields":{k:f for k,f in vars(m).items() if k not in ["_state","id"] }} for m in amps], cls=DjangoJSONEncoder)
```

and loading of the fixture was tested by adding a "dmd_test" database to `django.db.connections` and running
```py
call_command("migrate","dmd",database="dmd_test")
call_command("loaddata","coding_systems/dmd/fixtures/asthma-medication.dmd_test_20200101.json",database="dmd_test")
```

which revealed some missing referenced entities ( licensing authorities, availability restrictions, virtual product prescribing statuses) which were then added.

### DIFF
--- a/coding_systems/dmd/fixtures/asthma-medication.dmd_test_20200101.json
+++ b/coding_systems/dmd/fixtures/asthma-medication.dmd_test_20200101.json
@@ -113,6 +113,16 @@
     }
 },
 {
+    "model": "dmd.unitofmeasure",
+    "pk": "408102007",
+    "fields": {
+        "cd": "408102007",
+        "cddt": "2022-01-17",
+        "cdprev": "3319711000001103",
+        "descr": "unit dose"
+    }
+},
+{
     "model": "dmd.basisofname",
     "pk": 1,
     "fields": {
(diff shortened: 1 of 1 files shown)


rw251:
Looks good. I wondered if choosing salbutamol might not be sufficient for testing things to do with route - but it seems that there are salbutamol tablets so that would be fine.

rw251 (approved):

Jongmassey:
I'd not given Route much thought as it's not in scope right now. Unless I'm mistaken the tablets aren't actually included in this subset. I'll see if I can re-run this in such a way as to get a diversity of route and strength rather than just the first 10.

PR #2583, opened 2025-05-22
Implement VTM-VMP-AMP hierarchy

### LINKED ISSUE #2506: Allow dm+d codes to be viewed as a tree in codelist builder
# Why are we doing this?

We wish to provide a NHS Dictionary of Medicines and Devices (dm+d) codelist builder.
Part of the [interface](https://github.com/opensafely-core/opencodelists/blob/main/coding_systems/base/coding_system_base.py#L113) a `BuilderCompatibleCodingSystem` must present to the builder are means to ascertain the ancestor and descendant relationships of a set of codes such that a hierarchy can be built from them.
This hierarchy then enables viewing codes in a tree structure.

* `ancestor_relationships` and `descendant_relationships` methods are implemented in the dm+d coding system, based on a hierarchy of `VTM` -> `VMP` -> `AMP`
* Adequate unit tests exist for these methods

# What are we doing?

* writing an [`ancestor_relationships`](https://github.com/opensafely-core/opencodelists/blob/73aca2898cdc0d543ff8445c46769b0f8c8c8ab5/coding_systems/base/coding_system_base.py#L122-L129) method in the dm+d coding system which will accept a set of `codes` and from them produce a set of (`parent`,`child`) tuples of all the ancestors of the provided codes based on the above stated hierarchy. [Example](https://github.com/opensafely-core/opencodelists/blob/Jongmassey/spike_dmd_codelist_builder/coding_systems/dmd/coding_system.py#L16-L25) from spike branch
* writing an [`descendant_relationships`](https://github.com/opensafely-core/opencodelists/blob/73aca2898cdc0d543ff8445c46769b0f8c8c8ab5/coding_systems/base/coding_system_base.py#L139-L146) method in the dm+d coding system which will accept a set of `codes` and from them produce a set of (`parent`,`child`) tuples of all the descendents of the provided codes based on the above stated hierarchy. [Example](https://github.com/opensafely-core/opencodelists/blob/Jongmassey/spike_dmd_codelist_builder/coding_systems/dmd/coding_system.py#L26-L35) from spike branch
* testing that these methods return correct hierarchies given sets of codes likely to be returned by the coding system's [search methods](https://github.com/opensafely-core/opencodelists/issues/2505)
  * NB not all VMPs in the real dm+d data have VTMs

* Adding end-to-end Playwright tests for the features implemented in this ticket (where possible)

_NB: Example code from the dm+d builder spike branch is illustrative of how these features may be implemented but should not be considered to be 100% correct or bug-free_

---

Jongmassey:
Define a hierarchy of entites from a subset of the dm+d coding_system which will be presented to the builder via the `ancestor_relationships` and `descendant_relationships` methods.

All AMPs have an ancestor VMP, most but not all VMPs have an ancestor VTM so this must be handled in these methods.

AMPPs and VMPPs are excluded from this hierarchy as no user need for them has been determined, and codes for these types of entity are not present in the data sets made available through OpenSAFELY. Additionally, AMPPs have dual ancestors of AMPs and VMPPs which would be challenging to represent in these methods.

fixes #2506

### DIFF
--- a/coding_systems/base/coding_system_base.py
+++ b/coding_systems/base/coding_system_base.py
@@ -42,6 +42,10 @@ class BaseCodingSystem(ABC):
     # are not guaranteed to be exhaustive in future.
     csv_headers = {"code": ["code"], "term": ["term"]}
 
+    @classmethod
+    def is_builder_compatible(self):
+        return issubclass(self, BuilderCompatibleCodingSystem)
+
     def __init__(self, database_alias):
         self.database_alias = database_alias
 
--- a/coding_systems/dmd/coding_system.py
+++ b/coding_systems/dmd/coding_system.py
@@ -11,6 +11,31 @@ class CodingSystem(BaseCodingSystem):
         "term": ["term", "dmd_name", "name", "nm", "description"],
     }
 
+    def ancestor_relationships(self, codes):
+        amps = AMP.objects.using(self.database_alias).filter(id__in=codes)
+
+        # get VMPs that are either in `codes` or are ancestors
+        # of the AMPs in `codes`
+        codes = set(codes) | {amp.vmp_id for amp in amps}
+        vmps = VMP.objects.using(self.database_alias).filter(id__in=codes)
+
+        # exclude null VTM-VMP relationships (i.e. VMPs with no VTM)
+        return {(amp.vmp_id, amp.id) for amp in amps} | {
+            (vmp.vtm_id, vmp.id) for vmp in vmps if vmp.vtm_id
+        }
+
+    def descendant_relationships(self, codes):
+        vmps_from_vtms = VMP.objects.using(self.database_alias).filter(vtm__in=codes)
+
+        # get AMPs that have ancestor VMPs that are either in `codes`
+        # or are descendants of VTMs that are in `codes`
+        codes = set(codes) | {vmp.id for vmp in vmps_from_vtms}
+        amps_from_vmps = AMP.objects.using(self.database_alias).filter(vmp_id__in=codes)
+
+        return {(vmp.vtm_id, vmp.id) for vmp in vmps_from_vtms} | {
+            (amp.vmp_id, amp.id) for amp in amps_from_vmps
+        }
+
     def lookup_names(self, codes):
         # A code is a unique identifier in dm+d which corresponds to a SNOMED-CT code
         # It could be the identifier for any of AMP, VMP, VTM, VMPP, AMPP
--- a/coding_systems/dmd/tests/test_coding_system.py
+++ b/coding_systems/dmd/tests/test_coding_system.py
@@ -73,6 +73,13 @@ def test_ancestor_relationships(dmd_data, coding_system):
         == expected_relationships
     )
 
+    # 3436211000001104: "Oxygen composite cylinders 1360litres B10C with integral headset" has no VTM
+    # and three descendant AMPS: 4086311000001109,4086111000001107,3106411000001109
+    oxygen_vmp = "3436211000001104"
+    oxygen_amps = ["4086311000001109", "4086111000001107", "3106411000001109"]
+    expected_relationships = {(oxygen_vmp, amp) for amp in oxygen_amps}
+    assert coding_system.ancestor_relationships(oxygen_amps) == expected_relationships
+
 
 def test_descendant_relationships(dmd_data, coding_system):
     """
@@ -109,3 +116,12 @@ def test_descendant_relationships(dmd_data, coding_system):
     # 10 VTM-VMP relationships + 10 VMP-AMP == 20
     # minus the four we have manually declared above
     assert len(descendant_relationships - expected_relationships) == 16
+
+    # 3436211000001104: "Oxygen composite cylinders 1360litres B10C with integral headset" has no VTM
+    # and three descendant AMPS: 4086311000001109,4086111000001107,3106411000001109
+    oxygen_vmp = "3436211000001104"
+    oxygen_amps = ["4086311000001109", "4086111000001107", "3106411000001109"]
+    expected_relationships = {(oxygen_vmp, amp) for amp in oxygen_amps}
+    assert (
+        coding_system.descendant_relationships([oxygen_vmp]) == expected_relationships
+    )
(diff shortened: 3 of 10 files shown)


coding_systems/dmd/coding_system.py
```diff
@@ -11,6 +11,34 @@ class CodingSystem(BaseCodingSystem):
...
+        # or are descendants of VTMs that are in `codes`
+        codes = set(codes) | {vmp.id for vmp in vmps_from_vtms}
+        amps_from_vmps = AMP.objects.using(self.database_alias).filter(vmp_id__in=codes)
+
+        # exclude null VTM-VMP relationships (i.e. VMPs with no VTM)
+        return (
+            {(vmp.vtm_id, vmp.id) for vmp in vmps_from_vtms}
+            | {(vmp.vtm_id, vmp.id) for vmp in vmps_from_vtms if vmp.vtm_id}
+            | {(amp.vmp_id, amp.id) for amp in amps_from_vmps}
+        )
```
rw251, line 40: I think the first two parts of this set union return the same thing i.e. I don't think you need to exclude the null VTM-VMPs when doing descendants. By definition `vmps_from_vtms` is just vmps with a vtm_id, so the second statement is redundant. Another way of looking at it, the parent codes are either going to be VTMs or VMPs - if all VTMs have VMP children, and all VMPs have AMP children, then it's all fine.

coding_systems/base/coding_system_base.py
```diff
@@ -42,6 +42,8 @@ class BaseCodingSystem(ABC):
     # are not guaranteed to be exhaustive in future.
     csv_headers = {"code": ["code"], "term": ["term"]}
 
+    is_builder_compatible = False
+
     def __init__(self, database_alias):
         self.database_alias = database_alias
```
rw251, line 48: I wonder if something like this declared on `BaseCodingSystem` would be better:

```python
    @property
    def is_builder_compatible(self):
        return isinstance(self, BuilderCompatibleCodingSystem)
```
I appreciate it could still be overridden just as easily on any coding system class, but it keeps the logic in one place, and I think makes it clearer what makes something builder compatible.

coding_systems/dmd/tests/test_coding_system.py
```diff
@@ -25,3 +25,87 @@ def test_code_to_term(dmd_data, coding_system):
...
+
+    # 3293111000001105:"Aerolin 100micrograms/dose Autohaler" has an ancestor of VMP 35936411000001109
+    # and, in turn, VTM 777483005
+    salbutamol_amps = ["3293111000001105"]
+    expected_relationships |= {("35936411000001109", "3293111000001105")}
+    assert (
+        coding_system.ancestor_relationships(salbutamol_vmps + salbutamol_amps)
+        == expected_relationships
+    )
+
```
rw251, line 75: Should we add a test for a VMP/AMP without a VTM?

coding_systems/dmd/tests/test_coding_system.py
```diff
@@ -25,3 +25,87 @@ def test_code_to_term(dmd_data, coding_system):
...
+    # Given VTM has 10 descendant VMPs, which have in total 10 descendant AMPs
+    expected_relationships |= {(salbutamol_vtm, salbutamol_vmp)}
+    descendant_relationships = coding_system.descendant_relationships([salbutamol_vtm])
+
+    assert expected_relationships.issubset(descendant_relationships)
+
+    # 10 VTM-VMP relationships + 10 VMP-AMP == 20
+    # minus the four we have manually declared above
+    assert len(descendant_relationships - expected_relationships) == 16
```
rw251, line 111: These tests are all fine. Though I did wonder whether it would be clearer if you defined the expected hierarchy in a {parent: [children]} dict and created methods that for a given code traversed the tree to get its ancestors/descendants. You could then iterate over every code in the tree and confirm the anc/desc were as expected. It's a bit circular because you need to define the tree traversal methods in the test - but that's probably ok because they're simpler than the ones we're actually testing. I'm happy to approve without changes here, so only do anything if you think it would be better. I think the main improvements would be to make it easier to maintain because if we changed the underlying fixtures, we'd just need to update the expected hierarchy in one place - but also that it automatically would test every code in the hierarchy rather than a subset.

coding_systems/base/coding_system_base.py
Jongmassey: Oh I like this idea

coding_systems/dmd/tests/test_coding_system.py
Jongmassey: Yes good point. I'm not sure I put one of those into the test data :facepalm:

coding_systems/base/coding_system_base.py
Jongmassey: turns out there's no such thing as a class property, so had to use a `@classmethod` instead

coding_systems/dmd/tests/test_coding_system.py
Jongmassey: Bit of a slog, but I've added another commit for this

coding_systems/dmd/coding_system.py
```diff
@@ -11,6 +11,32 @@ class CodingSystem(BaseCodingSystem):
...
+
+    def descendant_relationships(self, codes):
+        vmps_from_vtms = VMP.objects.using(self.database_alias).filter(vtm__in=codes)
+
+        # get AMPs that have ancestor VMPs that are either in `codes`
+        # or are descendants of VTMs that are in `codes`
+        codes = set(codes) | {vmp.id for vmp in vmps_from_vtms}
+        amps_from_vmps = AMP.objects.using(self.database_alias).filter(vmp_id__in=codes)
+
+        # exclude null VTM-VMP relationships (i.e. VMPs with no VTM)
```
rw251, line 35: micro nit - you don't need this comment

rw251 (approved):

coding_systems/dmd/coding_system.py
Jongmassey: :+1:

opencodelists/forms.py
```diff
@@ -44,15 +43,9 @@ def clean_new_password2(self):
...
-        [
-            (id, system.name)
-            for id, system in CODING_SYSTEMS.items()
-            if issubclass(system, BuilderCompatibleCodingSystem)
-        ],
-        key=lambda x: x[1].lower(),
-    )
-
+    CODING_SYSTEM_CHOICES = [("", "---")] + [
+        (system.id, system.name) for system in builder_compatible_coding_systems()
```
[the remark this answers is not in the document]
Jongmassey: I checked and they are the same in all cases. The folder names are used to answer the question of "what coding systems are there" - the population of `CODING_SYSTEMS` could be changed to use the `id` property instead which might be clearer but is functionally moot

PR #2584, opened 2025-05-22
Categorise all dmd codes in builder as "Product"

Jongmassey:


coding_systems/dmd/coding_system.py
```diff
@@ -37,3 +37,36 @@ def code_to_term(self, codes):
...
+                .filter(id__in=codes)
+                .values_list("id", flat=True)
+            )
+            + list(
+                AMP.objects.using(self.database_alias)
+                .filter(id__in=codes)
+                .values_list("id", flat=True)
+            )
+        )
+
```
rw251, line 71: erm - can you just return `{"Product": codes}` without all the db calls? I suppose this is slightly different in that if any of the `codes` are incorrect then this filters them out - but we don't do that for other coding systems. Am I missing something?

Jongmassey: AFAICT the other coding systems put "unknown" codes in their own "unknown" category. We should probably do that here.

rw251 (approved):

Issue #2523, opened 2025-05-13
Spike: don't persist fixtures in test database between tests


mikerkelly:
# Why are we doing this?

The persistence of the "universe" of potential fixture objects in the database throughout the session risks making tests non-independent. For example, a test requiring no instances of a certain model cannot run alongside a test using fixtures, and the Playwright tests (https://github.com/opensafely-core/opencodelists/issues/2408) need a workaround for the non-empty database.

Always instantiating the full universe of instances also means the dependencies between fixtures are somewhat implicit.

# How will we know when it's done?

We have considered and prototyped approaches to the above issues, and raised one or more issues for future work.

# What are we doing?

An alternative approach would be to store fixture instances as in-memory objects rather than in the database and save them to the database only when a test needs them. The database could then roll back to an empty state between tests.

This approach would allow us to invert the hierarchy of the universe so that only required instances are present in the database, not all of them.

The testing coding system databases would likely remain unchanged. 

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

PR #2597, opened 2025-05-28
Use fast password hasher in tests

mikerkelly:
Use a faster password hashing algorithm when running tests, as the test environment doesn't need to be as secure as production.

On my local machine with coverage disabled this speeds running `just test-py` from about 20s to about 15s, a 25% improvement. This will be more significant if the changes in https://github.com/opensafely-core/opencodelists/issues/2523 mean tests instantiate `User` instances more often.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -96,6 +96,16 @@ def reset_connections():
         del connections.databases[db]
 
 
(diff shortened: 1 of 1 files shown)


sebbacon:
No way!

Issue #2555, opened 2025-05-15
Change text in the "more" link for each code in the builder


tomodwyer:
# Why are we doing this?

Currently each code in the builder has an ellipsis at the end of the line. This ellipsis is a button which opens a modal. Users are not aware this is a link, nor are they aware what content is inside the modal until they click it. We should make it more obvious this can be clicked, and inform them ahead of time what data will be in the modal.

# How will we know when it's done?

When each code row has new text in the button.

# What are we doing?

- Figure out the best word for the button text
- Add a tooltip explaining what the button does (maybe?)
- Make the button look more like a link or button

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

PR #2606, opened 2025-06-04
Update the styling and content of the tree row modal

tomodwyer:
Closes https://github.com/opensafely-core/opencodelists/issues/2555

Changes made as part of this PR:

- Overflow scrolling is moved to a per section basis, so the whole page no longer scrolls horizontally for one particularly long code name
- The more info modal button now displays the text "More info"
  - This button only shows on hover of a row
  - The row that is being hovered has a background to highlight it
- To simplify the visual design, the brackets have been removed from around the code, and the colouring has been updated to more clearly identify the codes which have been excluded, or are in conflict

### After

https://github.com/user-attachments/assets/d65c2471-c740-4631-9357-3ffb457dea42

### DIFF
--- a/assets/src/scripts/components/CodelistBuilder.tsx
+++ b/assets/src/scripts/components/CodelistBuilder.tsx
@@ -184,7 +184,7 @@ export default class CodelistBuilder extends React.Component<
               <EmptyState />
             </Col>
           ) : (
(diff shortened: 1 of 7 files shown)


KatieB5 (approved):
Thank you Thomas!

PR #2607, opened 2025-06-04
Update version card using existing data

### LINKED ISSUE #2562: Update Version cards on builder draft and codelist version pages
# Why are we doing this?

We recently did some work to align the styling of the builder draft codelist page and the codelist version page (#2539). During the PR review, differences in the styling of the Versions cards across the two pages was [highlighted](https://github.com/opensafely-core/opencodelists/pull/2539#issuecomment-2886427225), as was the question of whether to include the Version card on the builder draft page at all. We should discuss our preferred approach and act accordingly.

(see screenshots below)

The Version card will be removed from the builder draft page, and the Version card will remain as is on the codelist version page. Or, the Version cards will be styled consistently across the builder draft page and the codelist version page.

# What are we doing?

1. As a team, discuss whether to keep or remove the Version card from the builder draft codelist page
2. If the decision is made to keep the Version card on the builder draft page, align the styling of this to the codelist version page

### Builder draft page

Image

### Codelist version page

Image

---

KatieB5:
Closes #2562

### DIFF
--- a/assets/src/scripts/components/Versions.tsx
+++ b/assets/src/scripts/components/Versions.tsx
@@ -12,7 +12,7 @@ interface VersionProps {
   versions: VersionT[];
 }
 
-function VersionSingle({ version }: { version: VersionT }) {
+function Version({ version }: { version: VersionT }) {
   const createdAt = new Date(version.created_at);
   const userTimeZone =
     Intl?.DateTimeFormat()?.resolvedOptions()?.timeZone || "UTC";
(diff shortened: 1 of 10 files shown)


assets/src/scripts/components/Versions.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <span className="created d-block p-0">
+        <OverlayTrigger
+          placement="right"
+          overlay={
+            <Tooltip id={version.tag_or_hash}>
+              When this codelist version was added to OpenCodelists
+            </Tooltip>
+          }
+        >
+          <abbr className="font-weight-bold">Created</abbr>
```
rw251, line 67: Bit of a nit, but the `abbr` tag in the version.html file also has the `title` attribute, which causes this behaviour for published codelists:
image

and this for draft codelists:
image

I don't mind which we do, but probably good to have the same behaviour for both.

PR #2608, opened 2025-06-05
Add validation for minimum search length

### LINKED ISSUE #2601: Require a minimal set of characters before performing a search by term
# Why are we doing this?

Currently it's possible to do a term search for a single character. For example, searching for "a" returns back 11,735 concepts in ICD10 (presumably many more in SNOMED CT). It's highly unlikely that a single character search would be useful. It causes unnecessary load on the server and inconveniences users. 

Users are unable to perform a search by term unless they've entered at least the minimum number of characters.

# What are we doing?

* Select a minimal number of characters to search for, e.g limit to a minimum of three characters.
* Add frontend and backend validation to prevent users from running a term search with less than the minimum number of characters.

---

StevenMaude:
This PR adds client-side and server-side validation for a minimum search length.

Fixes #2601.

### DIFF
--- a/assets/src/scripts/components/SearchForm.tsx
+++ b/assets/src/scripts/components/SearchForm.tsx
@@ -28,7 +28,6 @@ export default function SearchForm({
       minLength: number;
       maxLength: number;
       validationMaxLengthMsg: string;
-      validationMinLengthMsg: string;
     };
   } = {
     term: {
@@ -37,7 +36,6 @@ export default function SearchForm({
       placeholder: "Enter a search term…",
       minLength: MIN_TERM_LENGTH,
       maxLength: MAX_TERM_LENGTH,
-      validationMinLengthMsg: `Your search term is below the minimum length of ${MIN_TERM_LENGTH} characters`,
       validationMaxLengthMsg: `Your search term has reached the maximum length of ${MAX_TERM_LENGTH} characters`,
     },
     code: {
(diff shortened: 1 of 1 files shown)


builder/tests/test_views.py
```diff
@@ -379,3 +380,92 @@ def test_search_length_validation(client, draft_with_no_searches):
...
+    # A minimum search code length of 1 means the value
+    # below the minimum is zero.
+    # What happens if we enter a search below the minimum
+    # code length is that we encounter the case where no
+    # no search is entered. That case is handled differently
+    # by new_search() to guard against searches
+    # that cannot be slugified.
+    # We do not reach the Django MinValidationError,
+    # unlike for terms.
+    assert len(code_with_min_chars) == 1
```
StevenMaude: thought: This part being slightly unusual does feel a little bit like a hint the code is more complex than it could be. But I think this is OK, because the function of the slugify check is a special case to ensure there are valid characters.

builder/tests/test_views.py
```diff
@@ -379,3 +380,92 @@ def test_search_length_validation(client, draft_with_no_searches):
...
+    # A minimum search code length of 1 means the value
+    # below the minimum is zero.
+    # What happens if we enter a search below the minimum
+    # code length is that we encounter the case where
+    # no search is entered. That case is handled differently
+    # by new_search() to guard against searches
+    # that cannot be slugified.
+    # We do not reach the Django MinValidationError,
+    # unlike for terms.
+    assert len(code_with_min_chars) == 1
```
StevenMaude, line 471: thought: This part being slightly unusual does feel a little bit like a hint the code is more complex than it could be. But I think this is OK, because the function of the slugify check is a special case to ensure there are valid characters.

StevenMaude:
I've changed the behaviour in line with the suggestion by @rw251. The search box no longer shows as the error state for searches with too few characters, but the user still gets a prompt from the browser.

rw251 (approved):

PR #2625, opened 2025-06-11
Prevent users from creating codelists with non-alphanumeric names

### LINKED ISSUE #2576: NoReverseMatch: Reverse for 'user_version' with keyword arguments '{'username': 'rw251', 'codelist_slug': '', 'tag_or_hash': '53b0d563'}' not found. 1 pattern(s) tried: ['codelist/user/(?P<username>[^/]+)/(?P<codelist_slug>[^/]+)/(?P<tag_or_hash>[^/]+)/\\Z']
Sentry Issue: [OPENCODELISTS-SF](https://ebm-datalab.sentry.io/issues/6565692433/?referrer=github_integration)

```
NoReverseMatch: Reverse for 'user_version' with keyword arguments '{'username': 'rw251', 'codelist_slug': '', 'tag_or_hash': '53b0d563'}' not found. 1 pattern(s) tried: ['codelist/user/(?P<username>[^/]+)/(?P<codelist_slug>[^/]+)/(?P<tag_or_hash>[^/]+)/\\Z']
(29 additional frame(s) were not displayed)
...
  File "opencodelists/views/user.py", line 49, in user
    return render(request, "opencodelists/this_user.html", ctx)
  File "codelists/models.py", line 306, in get_absolute_url
    return reverse(
```

Jongmassey:
Fixes #2576

### DIFF
--- a/opencodelists/forms.py
+++ b/opencodelists/forms.py
@@ -4,6 +4,7 @@ from io import StringIO
 
 from django import forms
 from django.contrib.auth import password_validation
+from django.core.validators import RegexValidator
 
 from codelists.coding_systems import CODING_SYSTEMS, builder_compatible_coding_systems
 from codelists.models import Handle
@@ -54,7 +55,17 @@ def form_field_from_model(model_class, field_name, **kwargs):
       appropriate for their specific form."""
     model_field = model_class._meta.get_field(field_name)
     field_class = model_field.formfield
-    return field_class(validators=model_field.validators, **kwargs)
+    validators = model_field.validators
+    for validator in validators:
+        if isinstance(validator, RegexValidator):
+            kwargs["widget"] = forms.TextInput(
+                attrs={
+                    "pattern": f".*{validator.regex.pattern}.*",
+                    "title": validator.message,
+                }
+            )
+            break
+    return field_class(validators=validators, **kwargs)
 
 
 class CodelistCreateForm(forms.Form):


codelists/models.py
```diff
@@ -203,17 +203,27 @@ class Handle(models.Model):
     codelist = models.ForeignKey(
         "Codelist", on_delete=models.CASCADE, related_name="handles"
     )
-    name = models.CharField(max_length=255)
+    name = models.CharField(
+        max_length=255,
+        validators=[
+            RegexValidator(
+                regex=r"^[\w _-]+$",
+                message="Codelist names must be non-blank, and contain only letters, numbers, underscores, or hyphens.",
```
mikerkelly, line 211: Have you considered using a `SlugField` here? I think this may do exactly what you want here, and avoids us needing to create and test our own implementation.

Jongmassey: OK I've altered the error string now

codelists/models.py
```diff
@@ -203,17 +203,27 @@ class Handle(models.Model):
     codelist = models.ForeignKey(
         "Codelist", on_delete=models.CASCADE, related_name="handles"
     )
-    name = models.CharField(max_length=255)
+    name = models.CharField(
+        max_length=255,
+        validators=[
+            RegexValidator(
+                regex=r"^[\w _-]+$",
```
mikerkelly, line 210: Nit: `\w` included underscores so this is slightly redundant.

mikerkelly:
Could you explain why `codelists/actions.py` is the right place to invoke `clean_fields` model validation, and why we should validate all model fields rather than just the ones based on user free input (only `name`?)? Validating user input doesn't seem like it belongs in the business logic layer. The form or view code seems like a better fit. Currently I think you do form validation, then invoke the action, which does the model validation, then catch the ValidationError and tweak the form again. Is it somewhat redundant?

Instead, could we attach the desired validator to the `name` property of the two `CodelistCreateForm` either directly or by reference to the model property validator? Then the form validation lifecycle within the view will perform the validation as standard. Rather than having the view code catch and handle a `ValidationError` from `actions.py` then "manually" adjust the form errors, we can separate all the validation concern into the form code. Hmm, do you understand why this does not already work in `codelist/forms.py` via  `name = model_field(Handle, "name")`?

This also avoids the need for changing `blank=True` on `user`/`organisation`, which seems a little strange when those properties don't appear directly in a form and when they do appear they're part of the `owner` `ChoiceField`.

I wondered about converting the form to a `ModelForm` based on `Handle` or `Codelist` but maybe this isn't a good fit because of the multiple models updated on form save.

The new `codelists/tests/test_models.py` tests seem to be tests of the validator. I'm not sure that our tests should be testing that Django's code "works", and I'm not sure these tests are achieving anything else?

I think the "help_text" on the form ought to explain the restrictions, before form validation fails. It's currently baked into `templates/opencodelists/user_create_codelist.html` which isn't lovely. Can we change "Describe the purpose of this codelist in 255 characters or fewer." to "Describe the purpose of this codelist. Must be non-blank and contain only letters, numbers, underscores, spaces, or hyphens. Max 255 characters." maybe?

Jongmassey:
Righto, that's really helpful. From my read of the django docs and field validation I got the steer that the validation should live in the model (and that's probably me with my database head as well). 

If there's a way of doing the validation in the Form field (which is generated from/is the same as the model field?) then that feels like a much more logical place and as you say avoids the bespoke and duplicated error handling code.

Care to point me as to how to do it in the Form?

opencodelists/forms.py
```diff
@@ -42,9 +43,23 @@ def clean_new_password2(self):
...
+      appropriate for their specific form."""
+    model_field = model_class._meta.get_field(field_name)
+    field_class = model_field.formfield
+    return field_class(validators=model_field.validators, **kwargs)
+
+
 class CodelistCreateForm(forms.Form):
     owner = forms.ChoiceField()
-    name = forms.CharField(max_length=255, label="Codelist name")
+    name = form_field_from_model(Handle, "name", label="Codelist name")
```
StevenMaude, line 62: The original issue (#2576) mentioned frontend validation; maybe something like this would suffice?

```suggestion
    name = form_field_from_model(
        Handle,
        "name",
        label="Codelist name",
        widget=forms.TextInput(
            attrs={
                "pattern": ".*[A-Za-z0-9].*",
                "title": "The name of the codelist (must include at least one or more alphanumeric characters)",
            }
        ),
    )
```

The `.*` in `pattern` is required because `pattern` matches the entire string, and we only require that there are alphanumeric characters in the string.

StevenMaude (commented):
The code here looks fine to me as is.

It's currently missing the suggestion for frontend validation as in #2576.

codelists/tests/test_models.py
```diff
@@ -105,6 +106,47 @@ def test_handle_name_cannot_be_reused_by_user(user_codelist):
...
+def test_handle_name_validation_valid(user_codelist):
+    handle = Handle.objects.create(
+        codelist=user_codelist,
+        name="new-codelist_validname123!,\"'.",
+        slug="new-slug",
+        is_current=True,
+        user=user_codelist.user,
+    )
+    handle._meta.get_field("name").clean(handle.name, handle)
+
```
StevenMaude, line 148: question: Since this is valid, could we also `assert` the expected `name` here?

Jongmassey: that seems very reasonable

codelists/models.py
```diff
@@ -203,7 +203,17 @@ class Handle(models.Model):
...
     )
-    name = models.CharField(max_length=255)
+    name = models.CharField(
+        max_length=255,
+        validators=[
+            RegexValidator(
+                # Codelist names get slugified, which must result in a valid slug.
+                # Let's require alphanumeric characters as it's easier to explain.
+                regex=r"[a-zA-Z0-9]+",
+                message="Codelist names must contain at least one letter or number.",
```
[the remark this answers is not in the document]
Jongmassey: Could you point to where this is duplicated?

opencodelists/forms.py
```diff
@@ -54,7 +55,17 @@ def form_field_from_model(model_class, field_name, **kwargs):
...
-    return field_class(validators=model_field.validators, **kwargs)
+    validators = model_field.validators
+    for validator in validators:
+        if isinstance(validator, RegexValidator):
+            kwargs["widget"] = forms.TextInput(
+                attrs={
+                    "pattern": f".*{validator.regex.pattern}.*",
+                    "title": validator.message,
+                }
+            )
```
StevenMaude, line 66: thought: We could maybe validate that we only have one RegexValidator here and raise an exception otherwise, but it's probably fine as is.

PR #2626, opened 2025-06-11
Show "no search term" when last search deleted

### LINKED ISSUE #2534: Appearance of `[no search term]` search may not be consistent
Perhaps this is expected behaviour, but it seemed quirky to me as I observed it while writing up how to reproduce #2532.

It's possible for `[no search term]` to suddenly appear, after deleting a search, and then performing another search.

# To reproduce

1. Create a new codelist with PseudoBNF.
1. Search for "syrup".
1. Select "GSF-Syrup (`0601040I0BD`)".
1. Delete the "syrup" search.
1. Observe that no searches are listed at all, not even `[no search term]`.
1. Search for "paracetamol".
1. See that `[no search term]` appears, along with the search for "paracetamol".

# Video

https://github.com/user-attachments/assets/9fab2c0b-0a05-4194-8533-d1a5869a105f

StevenMaude:
Fixes #2534.

The previous behaviour was such that:

1. Enter a search.
1. That search appears in the list of searches.
1. Select a code.
1. Delete the search.
1. See no searches listed at all.
1. Search again.
1. See your search and "no search term" then appear as a search.

This commit changes the behaviour so that "no search term" appears even
with no other valid searches. This makes the UI more consistent by not
making "no search term" appear when you actually enter another search.

### DIFF
--- a/builder/views.py
+++ b/builder/views.py
@@ -90,7 +90,7 @@ def _draft(request, draft, search_id):
         for s in draft.searches.order_by("term")
     ]
 
-    if searches and draft.code_objs.filter(results=None).exists():
+    if draft.code_objs.filter(results=None).exists():
         searches.append(
             {
                 "term_or_code": "[no search term]",


StevenMaude:
I think this is correct, but I'm not sure if there's any other quirky cases why the original code checked for the presence of searches, before adding "no search term".

PR #2630, opened 2025-06-12
Allow users to edit metadata in the draft phase of a codelist

rw251:
A draft codelist now has a metadata tab where users can edit the description/methodology/references sections of the metadata. It looks like this:

[Screencast from 12-06-25 15:22:56.webm](https://github.com/user-attachments/assets/96bd8edf-cad7-4582-ab22-cd488ff11af4)

### DIFF
--- a/assets/src/scripts/components/CodelistBuilder.tsx
+++ b/assets/src/scripts/components/CodelistBuilder.tsx
@@ -39,6 +39,30 @@ interface CodelistBuilderProps extends PageData {
   };
 }
 
+/**
+ * Creates a fetch options object with standard headers including CSRF token
+ * @param body - Object to be sent as JSON in the request body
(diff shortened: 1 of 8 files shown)


assets/src/scripts/components/CodelistBuilder.tsx
```diff
@@ -14,25 +14,270 @@ import Title from "./Title";
...
+  const handleCancel = () => {
+    setEditingIndex(null);
+  };
+
+  return (
+    <div className="card">
+      <div className="card-body">
+        <h3 className="h5 card-title">References</h3>
+        <hr />
+        <p style={{ fontStyle: "italic" }}>
```
tomodwyer, line 140:
```suggestion
        <p className="font-italic">
```

PR #2632, opened 2025-06-13
Improve error message when a database is missing (in dev)

rw251:
When developing locally, if you try and view a codelist from a coding_system release that is not available locally, you get an unhelpful message like:  "no such table: dmd_amp".  This change gives a more helpful message by saying which database is missing, and looks like this:

image

### DIFF
--- a/codelists/views/decorators.py
+++ b/codelists/views/decorators.py
@@ -1,5 +1,6 @@
 from functools import wraps
 
+from django.db import OperationalError
 from django.db.models import Q
 from django.shortcuts import get_object_or_404, redirect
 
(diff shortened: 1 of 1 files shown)


Jongmassey:
>When developing locally,

Is it possible to add a check for settings.DEBUG in this code and only show this error in dev?

Jongmassey:
Nonetheless seems like a great QoL devex improvement

rw251:
> Is it possible to add a check for settings.DEBUG in this code and only show this error in dev?

rw251:
> Is it possible to add a check for settings.DEBUG in this code and only show this error in dev?

I thought about that, but presumably this could also happen in production if somehow we lost one of the databases - perhaps someone trying to copy one locally but doing the wrong command? Maybe it's unlikely, but I suspect we'd still value the better logs - and it's not as if end users see this error message in production.

Jongmassey:
That's true, I'll approved

Jongmassey (approved):

PR #2638, opened 2025-06-16
Use path converters to ensure `int` used in `search` and `search-delete` URL patterns

### LINKED ISSUE #2609: Use path converters for URL type checking and conversion
# Why are we doing this?

We have a number of views that raise an unhandled `ValueError` and 500 if a misformatted URL is passed. This is because they perform operations that require values of certain types but neither they nor the URL dispatcher do anything to ensure passed values are of those types or handle what happens if the values they are using are not. Normally users follow valid URLS we construct, but sometimes they enter them manually incorrectly as in the example below. They should get a 404 or other error in such case, not a 500. The user should not be able to easily trigger unhandled exceptions. This also makes Sentry noise.

For example:

- `/builder/60b18755/search/lives/delete/` fails because the view is expecting the penultimate group to be an integer to match pattern: `<hash>/search/<search_id>/<search_slug>/`. This is a real error that a user hit, see [this Sentry report](https://ebm-datalab.sentry.io/issues/6650965346/?alert_rule_id=1877862&alert_type=issue&notification_uuid=aa54f357-485c-47c0-9fcb-bb36e2268893&project=5248013&referrer=slack). It seems they were trying to manually enter the URL to delete a search.
- `/builder/h/` fails as the view is expecting the draft ID to be hex in pattern `<hash>/`. Local development example:
```
  File "/opencodelists/builder/decorators.py", line 15, in wrapped_view
    id = unhash(hash, "CodelistVersion")
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opencodelists/opencodelists/hash_utils.py", line 55, in unhash
    return int(h, 16) * inv_c % N
           ^^^^^^^^^^
ValueError: invalid literal for int() with base 16: 'h'
```

There is defensive code in place to handle such bad URLs being entered, either in the URL dispatcher, view code, or `actions.py` code, or more than one of these.

# What are we doing?

The problem is URL patterns are matching URLs views can't handle and passing bad parameters to those views. We should use Django's in-built [path converters](https://docs.djangoproject.com/en/5.2/topics/http/urls/#path-converters) feature in the URL dispatchers to avoid this. Then, URLs with the wrong types won't match the relevant patterns and the user will get a 404, instead of unhandled exceptions being raised and leading to a 500, and a Sentry event. We can write custom converters for things like the draft version identifiers which are hex. E.g., builder/urls.py could look something like:

```python 
class CodelistVersionHashConverter:
    regex = '[0-9a-fA-F]+'

    def to_python(self, value):
        return value  

    def to_url(self, value):
        return value

from django.urls import path, register_converter
register_converter(CodelistVersionHashConverter, 'hash')

urlpatterns = [
    path("<hash:hash>/", views.draft, name="draft"),
    path("<hash:hash>/search/<int:search_id>/<slug:search_slug>/", views.search, name="search"),
    path(
        "<hash:hash>/search/<int:search_id>/<slug:search_slug>/delete/",
        views.delete_search,
        name="delete-search",
    ),
    path("<hash:hash>/no-search-term/", views.no_search_term, name="no-search-term"),
    path("<hash:hash>/update/", views.update, name="update"),
    path("<hash:hash>/search/", views.new_search, name="new-search"),
]
```

We could handle some of the conversion currently done in `builder/decorators.py` in the converter, as well as validating that the value is hex. We should apply this approach systematically across all the URL dispatcher `urls.py` modules, to avoid as many such issues as possible, and so we have a consistent approach to build on. We may be able to remove per-view code that handles, for example, `str` to `int` conversion as the converter can handle it.

For belt-and-braces we could also make each view perform type checking before calling unsafe functions on values passed from parameters, or handle the resulting `ValueError` (or other errors) sensibly. The cost of that may be slightly less readable view code. EDIT: I'm leaning towards treating the view-and-URL-dispatcher as one unit to test with a simulated client, and not having type-checking/error handling within the view code. Probably view code is only ever called after URL dispatch so it's reasonable to treat that as the entrypoint for tests and slim down the view code.

There are a number of views that have relevant parameters that might benefit from path converters, probably we should split this issue into smaller tasks/PRs:

- [x] The numerical `search_id` in builder views in the first example under "Why are we doing this" -- https://github.com/opensafely-core/opencodelists/pull/2638
- [X] Path parameters called slugs in their variable names -- https://github.com/opensafely-core/opencodelists/pull/2641
- [X] orgs/usernames in `opencodelists` -- also slugs. Same PR -- https://github.com/opensafely-core/opencodelists/pull/2641
- [ ] Builder hex hashes (second example under "Why are we doing this" and similar, code example above).
- [x] `tag_or_hash` in `codelists/urls.py`. https://github.com/opensafely-core/opencodelists/pull/2651 Corresponds to `CodelistVersion` property `tag_or_hash`. Presumably the hash is hex characters, what can tag be? `codelist/models.py` doesn't explain what tags are so a little poking around required to see how it's used... turns out free text that's currently mostly an historical thing, [Slack thread](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1750261936708709) thinking about how that could change. Upshot for this, probably just a `str:` converter...

---

mikerkelly:
This ensures that the view cannot receive non-int `search_ids`, which would cause a `ValueError` when passed to `get_object_or_404` as the `id` parameter.

Partially resolves #2609.

### DIFF
--- a/builder/tests/test_views.py
+++ b/builder/tests/test_views.py
@@ -1,6 +1,7 @@
 import pytest
 from django.core.validators import MinLengthValidator
 from django.urls import reverse
+from django.urls.exceptions import NoReverseMatch
 
 from codelists.actions import create_codelist_from_scratch
 from codelists.coding_systems import most_recent_database_alias
@@ -54,6 +55,16 @@ def test_search(client, draft_with_some_searches):
         rsp.context["results_heading"]
         == f'Showing {num_concepts} concepts matching "{slug}"'
     )
+
+
+def test_search_non_int_search_id(client, draft_with_some_searches):
+    """Test that the URL dispatcher can't call the search view with a
+    non-integer string as `search_id`."""
+    slug = "arthritis"
+    search_id = "bad_search_id"
+
+    with pytest.raises(NoReverseMatch):
+        client.get(draft_with_some_searches.get_builder_search_url(search_id, slug))
 
 
 def test_no_search_term(client, draft_with_some_searches):
--- a/builder/tests/test_views.py
+++ b/builder/tests/test_views.py
@@ -518,3 +518,15 @@ def test_search_delete_get(client, minimal_draft):
 
     # 405 Method not allowed.
     assert rsp.status_code == 405
+
+
+def test_search_delete_non_int_search_id(client, minimal_draft):
+    """Test that the URL dispatcher can't call the `delete_search` view with a
+    non-integer string as `search_id`."""
+    client.force_login(minimal_draft.author)
+
+    slug = "tennis-toe"
+    search_id = "bad_search_id"
+
+    with pytest.raises(NoReverseMatch):
+        client.post(minimal_draft.get_builder_delete_search_url(search_id, slug))
--- a/builder/tests/test_views.py
+++ b/builder/tests/test_views.py
@@ -480,3 +480,41 @@ def test_min_search_length_validation(client, draft_with_no_searches):
     # We do not reach the Django MinValidationError,
     # unlike for terms.
     assert len(code_with_min_chars) == 1
+
+
+def test_search_delete(client, minimal_draft):
+    """Test that a POST to the `delete_search` view deletes the selected
+    search, and redirects to the draft view."""
+    client.force_login(minimal_draft.author)
+
+    term = "tennis toe"
+    slug = "tennis-toe"
+    searches = minimal_draft.searches.all()
+    # Two search terms pre-populated in fixture.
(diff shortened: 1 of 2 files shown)


mikerkelly:
> Looks good to me.
> 
> (Your phrasing in the opening message — "Partially resolves" — is still going to close the issue on merge, since GitHub only matches the "resolves". Either you'll need to rephrase the message or reopen the issue when it's automatically closed.)

Thanks. Yes I thought this would be less effort than sub-issues in this case.

PR #2643, opened 2025-06-18
Cache hierarchies as needed during import

Jongmassey:
The coding system import routines make an assumption that if a `CodelistVersion` is in a coding system that _supports_ hierarchies then it must already have a cached hierarchy. This is not always the case, especially when a coding system has just been updated to support hierarchies/the builder. This failure of this assumption causes the import process to fail.

This is in part due to the confusion and overloading of the `has_hierarchy` property which we should seek to refactor in future.

This commit mirrors the behaviour of the codelist version view which checks for the presence of a cached_hierarchy and creates one if required.

This means that the coding system import process will fill in all missing cached hierarchies every time it's run, which is a nice side benefit and means if we restore from an old backup the next time an import runs, it'll update the hierarchies as needed.

The downside is that the first run of the dm+d import will take a lot longer than usual while all these hierarchies are calculated (effectively we'll be calculated 2 hierarchies per `CodelistVersion` rather than the usual 1).

### DIFF
--- a/coding_systems/base/import_data_utils.py
+++ b/coding_systems/base/import_data_utils.py
@@ -10,6 +10,7 @@ from django.db.models import Q
 from django.utils import timezone
 from tqdm import tqdm
 
+from codelists.actions import cache_hierarchy
 from codelists.coding_systems import CODING_SYSTEMS
 from codelists.hierarchy import Hierarchy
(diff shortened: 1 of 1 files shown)


Jongmassey:
Addressed by #2657  instead

PR #2657, opened 2025-06-19
Script to update missing cached_hierarchies

Jongmassey:


codelists/scripts/update_missing_cached_hierarchies.py
```diff
@@ -0,0 +1,36 @@
+"""
+Update a missing cached_hierarchy for all CodelistVersions in a given Coding System.
+
```
rw251, line 3: Perhaps a bit of text saying why we created it (fixing dmd codelists when we implemented the builder), and when it might need running again (maybe never, maybe just if you implement a builder for a coding system that already has codelists?)

PR #2660, opened 2025-06-20
Document BNF to dmd conversion

Jongmassey:
Improvements in #2635  made me think we should document this feature, particularly if we are going to announce those improvements in Platform News.


StevenMaude (approved):
This is fine.

I've a couple of editing suggestions that I think make the instructions easier to read, but feel free to use them or not, as you prefer.

PR #2667, opened 2025-06-25
Improve the styling of the My Codelists page

### LINKED ISSUE #2538: Style and content improvements to the my codelists page
# Why are we doing this?

Currently the "My Codelists" page splits codelists into 4: published by user; published by user on behalf of an org; drafts under review; drafts. Some metadata is also displayed (coding system, number of codes, last updated date). A similar view is when you look at the codelists of another user, but this just displays the published codelists of that user.

We could improve the user's ability to find the right codelist by improving the style of the page, and also by displaying more metadata about each codelist.

For reference, My Codelists currently looks like this:

Image

The view of another users codelists looks like this:

Image

Users can more easily find the codelist they want to edit/view.

# What are we doing?

These are suggestions based on feedback during the work to add the metadata to the codelist list. The actual work may differ if better ideas are thought of.
- [ ] Convert the lists of codelists into a single table with searching and column sorting
- [ ] Think of a good way to display a codelist that has a published AND a draft version. Currently the codelist appears twice on the page, once under each relevant heading. In a table we probably want one row per codelist, but need a good way to show that the codelist has a published and a draft version, and a link to get to either one
- [ ] Split out the owner from the codelist link ("owned by xxx") into a separate column - perhaps just an optional column that displays the org for a codelist i.e. don't display the user name as its always just the logged in person
- [ ] The badges probably don't need to be badges any more if they have a designated column - but maybe for style consistency we could keep them for the coding system badge
- [ ] If space allows we can add the following metadata if users would find it useful
  - [ ] Number of versions
  - [ ] The coding system version
  - [ ] Whether an old-style CSV upload (and therefore non-editable)

---

rw251:
Fixes #2538 

The My Codelists page is now a searchable table as shown here:

[Screencast from 25-06-25 11:34:04.webm](https://github.com/user-attachments/assets/b11dcb1b-3527-42c3-b191-e9fe989bbd08)

When viewing another user's (published) codelists it looks like:

image

### DIFF
--- a/assets/src/scripts/codelists-list.js
+++ b/assets/src/scripts/codelists-list.js
@@ -1,0 +1,129 @@
+/*
+ * Interactivity for the codelist table.
+ *
+ * Each row in the table represents a single codelist. The behaviour changes
+ * depending on whether the codelist has one visible version, or multiple
+ * visible versions. A codelist can have just a single version when:
+ *  - viewing the latest published versions for another user
+ *  - if a codelist only has one version
+ *  - if filtering the table leads to just one version (e.g. filtering to drafts)
+ *
+ * When just one version is visible then we want the whole row to highlight on
+ * hover, and clicking anywhere in the row open that version. If multiple
+ * versions are visible, then the hover behaviour should only occur in the nested
+ * table containing the versions.
+ *
+ * This file has the logic for the above, and for the search box filtering.
+ */
+
+// Get all the codelist rows into a data array for future use
+const codelistRows = Array.from(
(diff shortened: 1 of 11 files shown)


Jongmassey (approved):
This all seems good.

One thing I've noticed when I log in locally as my user (which has a lot of user and organisational codelists) is that I miss the separation of these two codelist types.

An idea to get a bit of that functionality back would be to make the table sortable (such that one could sort by Owner) - is that feasible?

Issue #2674, opened 2025-06-26
Extra ICD description text for "more info" modal


rw251:
# Why are we doing this?

ICD10 has extra information about codes that we are not currently showing to users.

Here is an example from the ICD10 browser:
![Image](https://github.com/user-attachments/assets/a6f0fa0a-c52e-4de1-9409-2fd17b2ab895)

And here is what we display:

![Image](https://github.com/user-attachments/assets/ccd768a5-90d6-4745-a648-91d0f37644f2)

While users could use the ICD10 browser side by side, this extra context should probably appear in the "more info" modal.

NB - no user has requested this it just seems like a useful thing to display.

# How will we know when it's done?

The ICD10 more info modal will display this information.

# What are we doing?

- [ ] Update the ICD `import_data.py` file. For each code we currently pull the `Rubric` tag with `kind="preferredLong"` if it exists, falling back to `kind="preferred"`.  We should also parse the extra `Rubric` fields - I think this is limited to
    - `kind="description"` - e.g. "Kwashiorkor - E40" has a description field of "Severe malnutrition with nutritional oedema with dyspigmentation of skin and hair. "
    - `kind="text"` - seems similar to "description" e.g. "Tetralogy of Fallot - Q21.3" has a text field of "Ventricular septal defect with pulmonary stenosis or atresia, dextroposition of aorta and hypertrophy of right ventricle. "
    - `kind="inclusion"` - e.g. the screenshot above shows which types of DM are included...
    - `kind="exclusion"` - ... and excluded
    - `kind="coding-hint"` - might not want this as it's targeted at people doing the coding e.g. "Use additional code, if desired, to identify the obstetric cause (direct or indirect)" - but then again sometimes useful things like "Meconium ileus in cases where cystic fibrosis is known not to be present"
    - `kind="note"` - similar to "coding-hint" in that it's targeted at the coders, but sometimes has useful info
    - `kind="modifierlink"` - we should almost certainly continue to ignore these. It's for when a series of codes (E10, E11, E12 etc) have the same modifiers (i.e. 0,1,2,3..). So if 0 means "with coma", then E100 is E10 with coma, and E110 is E11 with coma. These modifierlinks are text telling you where to find the modifiers - but we already parse them into separate codes.
    - There is also "footnote", but that only occurs once to say something about smallpox, and "introduction" (9 occurrences) which provide an intro to a large block of codes. I don't think either would be useful to add
- [ ] Provide the extra info in the "more info" modal
[...]

PR #2677, opened 2025-06-26
Add synonyms to the "more info" modal pop up

### LINKED ISSUE #2556: Add SNOMED-CT synonyms for each code
# Why are we doing this?

SNOMED-CT contains synonyms for each code, we should display these to users so they can search for synonyms.

When users can see synonyms for each code in their codelist.

# What are we doing?

- Add an option to the ellipsis menu (#2555) to show synonyms
- Display synonyms in a modal
- Add an option to search for a synonym(?)

---

rw251:
Fixes #2556 

Synonyms (for SNOMED) now appear in the "more info" modal. Like this:

image

CTV3 also supports synonyms (via a "term" code) but I'm not sure we support that properly and can't see them in our database. Given it's been retired now for more than 5 years and not used anymore in opensafely (I think!) I don't think there's value in adding them.

ICD10 doesn't have synonyms, but it does have additional data that we could display. But this requires changing the import script. I have raised #2674 for this.

### DIFF
--- a/assets/src/scripts/__tests__/components/MoreInfoModal.spec.tsx
+++ b/assets/src/scripts/__tests__/components/MoreInfoModal.spec.tsx
@@ -1,5 +1,5 @@
 import "@testing-library/jest-dom";
-import { render, screen, waitFor, within } from "@testing-library/react";
+import { act, render, screen, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import React from "react";
 import { afterEach, beforeEach, expect, it, vi } from "vitest";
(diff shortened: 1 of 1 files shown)


assets/src/scripts/__tests__/components/MoreInfoModal.spec.tsx
```diff
@@ -90,12 +81,11 @@ it("shows 'No synonyms' if fetch fails", async () => {
   );
   const user = userEvent.setup();
   render(<MoreInfoModal {...props} />);
-  user.click(screen.getByRole("button", { name: /more info/i }));
 
-  // expect(await screen.findByText(/loading synonyms/i)).toBeInTheDocument();
-  await waitFor(() => {
-    expect(screen.getByText(/no synonyms/i)).toBeInTheDocument();
+  await act(async () => {
+    await user.click(screen.getByRole("button", { name: /more info/i }));
```
rw251, line 86: Based on the other tests can we get rid of the `await act()` wrapper function here?

PR #2678, opened 2025-06-26
Change the sort order for ICD10 and CTV3 to sort by code rather than term

### LINKED ISSUE #2645: Return ICD-10 codes in code order rather than alphabetical
# Why are we doing this?

The codes in ICD-10 are prefix-ordered e.g. E10 comes logically before E11, E10-E14 represent the group of "diabetes mellitus" codes etc. A user has said that it would be easier to navigate if the codes were returned by order of code rather than alphabetically according to the description (as it is now).

ICD10 codes will display in code order.

# What are we doing?

Updating the default sort for ICD10 codes in the hierarchy to use the code rather than the term.

---

rw251:
Fixes #2645 

Adds a `sort_by_term` boolean to the `BuilderCompatibleCodingSystem` class which defaults to `True`. This is then overridden in the ICD10 and CTV3 coding systems.


tomodwyer:
One thing I'm trying to do in the refactor is to reduce the amount of prop drilling (_passing the same property down through many components to the component it is required in_) that we're doing. 

As this `sortByTerm` is only used in one location, the `<Tree />` component, I think that is the point we should retrieve the script from the page. Example below:

```tsx
import React from "react";
import Hierarchy from "../_hierarchy";
import { readValueFromPage } from "../_utils";
import { AncestorCode, PageData, ToggleVisibility } from "../types";
import TreeRow from "./TreeRow";

interface TreeProps {
  allCodes: PageData["allCodes"];
  ancestorCode: AncestorCode;
  codeToStatus: PageData["codeToStatus"];
  codeToTerm: PageData["codeToTerm"];
  hierarchy: Hierarchy;
  isEditable: PageData["isEditable"];
  toggleVisibility: ToggleVisibility;
  updateStatus: Function;
  visiblePaths: PageData["visiblePaths"];
}

export default function Tree({
  allCodes,
  ancestorCode,
  codeToStatus,
  codeToTerm,
  hierarchy,
  isEditable,
  toggleVisibility,
  updateStatus,
  visiblePaths,
}: TreeProps) {
  const sortByTerm = readValueFromPage("sort-by-term");

  return hierarchy
    .treeRows(ancestorCode, codeToStatus, codeToTerm, visiblePaths, sortByTerm)
    .map((row) => (
      <TreeRow
        allCodes={allCodes}
        code={row.code}
        codeToStatus={codeToStatus}
        codeToTerm={codeToTerm}
        hasDescendants={row.hasDescendants}
        hierarchy={hierarchy}
        isEditable={isEditable}
        isExpanded={row.isExpanded}
        key={row.path}
        path={row.path}
        pipes={row.pipes}
        status={row.status}
        term={row.term}
        toggleVisibility={toggleVisibility}
        updateStatus={updateStatus}
      />
    ));
}
```

This would remove the need for a lot of the lines of code simply shuffling the boolean value around.

PR #2692, opened 2025-07-01
Add djhtml to indent and format HTML templates

tomodwyer:
[DjHTML is a pure-Python Django/Jinja template indenter](https://pypi.org/project/djhtml/), so rather than relying on users to hand-indent our template files, we can have a tool do it for us.

This PR:
- Adds the dependency
- Updates existing just recipes to use djhtml
- Removes some unused just recipes that weren't being used
- Adds a pre-commit hook to fix files
- Formats the existing files in one commit

### DIFF
--- a/templates/versioning/latest_releases.html
+++ b/templates/versioning/latest_releases.html
@@ -5,17 +5,14 @@
 {% block title_extra %}: Coding System Releases{% endblock %}
 
 {% block content %}
-
   <h1> Coding System Releases</h1>
 
   {% for coding_system_release in latest_releases %}
     <h4>{{ coding_system_release.name }}</h4>
     <ul>
       <li>Latest release: {{ coding_system_release.release_name }}</li>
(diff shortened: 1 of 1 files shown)


templates/versioning/latest_releases.html
```diff
@@ -6,16 +6,16 @@
...
-        <li>Latest release: {{ coding_system_release.release_name }}</li>
-        <li>Valid from: {{ coding_system_release.release.valid_from }}
+      <li>Latest release: {{ coding_system_release.release_name }}</li>
+      <li>Valid from: {{ coding_system_release.release.valid_from }}
         <li>Imported on: {{ coding_system_release.release.import_timestamp }}
-    </ul>
+        </ul>
 
-{% endfor %}
+  {% endfor %}
```
rw251, line 19: I realise this is a different issue, but the odd indenting here made me realise that this doc is missing some closing `</li>` tags

rw251 (approved):

PR #2695, opened 2025-07-01
Open details element to show CSV error message

tomodwyer:
So that users know there is an error message linked to the CSV upload

## Before

https://github.com/user-attachments/assets/4358d061-50fb-417d-abee-77c91d93490f

## After

https://github.com/user-attachments/assets/98220317-0c8e-47b6-8817-24072cf3260d


StevenMaude (approved):
This looks good.

It works fine locally with a bad code being correctly identified.

I did initially get a little confused by the video. The error it shows includes a Django template tag in the CSV, and wondered if there was a strange bug somewhere. But I guess that actually a template was uploaded as an "incorrect CSV"?

Error message for the OpenCodelists upload CSV function, where the message includes a Django template tag.

tomodwyer:
@StevenMaude Ha! Yes I just saved the open HTML file as a `.csv` so I knew it was invalid. Didn't even think about the error code showing that 😆

PR #2700, opened 2025-07-02
Add a link from a codelist to the other codelists owned by that user or org

### LINKED ISSUE #2669: Button to navigate to a person's codelists
# Why are we doing this?

[A user has said](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1750767973374669?thread_ts=1750263160.857559&cid=C069SADHP1Q) that it would be really useful to have a button to go from a single codelist to the list of all codelists from that user. Currently they do this by manually creating the URL.

The button exists.

# What are we doing?

- [ ] Add a button to the single codelist view called "Show all codelists by <username>". Maybe just says "Show all my codelists" if you are codelist owner - or maybe it just isn't there at all given you can go to "My codelists" quite easily

---

rw251:
Fixes #2669 
- Add a conditional element that shows the author if it is not an organisational codelist
- Make both the new author field, and the existing org field, into links that take you to the list of codes owned by that person
- Drive-by update of the justfile. A recent change meant the pre-commit hook was failing because without `*args` the hook was interpreting `just check <filename>` to mean that both `check` and `<filename>` were recipes. Now it knows that <filename> is a parameter to the recipe `check`.

image
image

### DIFF
--- a/codelists/tests/views/test_version.py
+++ b/codelists/tests/views/test_version.py
@@ -1,3 +1,5 @@
+from django.urls import reverse
+
 from builder.actions import save as save_for_review
 from codelists.actions import publish_version
 from codelists.models import Status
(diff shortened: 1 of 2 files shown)


justfile
Jongmassey, line 1: Might be cleaner to have this in a separate commit as it's a driveby fix

codelists/tests/views/test_version.py
```diff
@@ -102,3 +104,40 @@ def test_medication_banner_for_owner_of_codelist(client, bnf_version_asthma):
         b"there is already a draft version for this codelist which could be published"
         not in rsp.content
     )
+
+
+def test_organisation_hyperlink_in_version_detail(client, organisation_codelist):
+    """Test that organisation-owned codelists have a hyperlink to organisation index page."""
+    # Get a published version of the organisation codelist
+    version = organisation_codelist.versions.first()
```
Jongmassey, line 112:
```suggestion
    version = organisation_codelist.latest_published_version()
```

codelists/tests/views/test_version.py
```diff
@@ -102,3 +104,40 @@ def test_medication_banner_for_owner_of_codelist(client, bnf_version_asthma):
...
+
+    # Check url exists
+    assert f'href="{org_url}"'.encode() in rsp.content
+    assert organisation_codelist.organisation.name.encode() in rsp.content
+
+
+def test_author_hyperlink_in_version_detail(client, user_codelist):
+    """Test that user-owned codelists have a hyperlink to author's page."""
+    # Get a published version of the user codelist
+    version = user_codelist.versions.first()
```
Jongmassey, line 132:
```suggestion
    version = user_codelist.latest_published_version()
```

Jongmassey (approved):
Couple of suggestions to make the tests more resilient to fixture changes/less dependent on ordering

justfile
rw251: Good point. I'll do that.

PR #2710, opened 2025-07-04
Jongmassey/op links in modal

### LINKED ISSUE #2672: Add link to OpenPrescribing dm+d browser to "more info" modal
# Why are we doing this?

User feedback from the pilot launch of the dm+d builder included a feature request for the "more info" modal dialogue for a dm+d code to include a link to that code's page on the [OpenPrescribing dm+d browser](https://openprescribing.net/dmd/).

We are including a link to the homepage of this tool in #2659 but it is more convenient to have a direct link for each code's page.

When a user clicks on the "more info" button on a dm+d code in the builder, the resultant modal dialogue should contain a link to the OpenPrescribing dm+d browser page for that code.

# What are we doing?

Altering the "More info" modal component to include the relevant OpenPrescribing dm+d browser link for the selected code where the codelist loaded in the builder is of the dm+d coding system.

---

Jongmassey:
Fixes #2672

### DIFF
--- a/assets/src/scripts/components/MoreInfoModal.tsx
+++ b/assets/src/scripts/components/MoreInfoModal.tsx
@@ -4,7 +4,7 @@ import Hierarchy from "../_hierarchy";
 import { getCookie, readValueFromPage } from "../_utils";
 import { Code, PageData, Status, Term } from "../types";
 
-interface CreateModalTextProps {
+interface CreateStatusTextProps {
   allCodes: PageData["allCodes"];
   code: Code;
   codeToStatus: PageData["codeToStatus"];
@@ -13,14 +13,14 @@ interface CreateModalTextProps {
   status: Status;
 }
 
-function createModalText({
+function createStatusText({
   allCodes,
   code,
   codeToStatus,
   codeToTerm,
   hierarchy,
   status,
-}: CreateModalTextProps) {
+}: CreateStatusTextProps) {
   const included = allCodes.filter((c) => codeToStatus[c] === "+");
   const excluded = allCodes.filter((c) => codeToStatus[c] === "-");
   const significantAncestors = hierarchy.significantAncestors(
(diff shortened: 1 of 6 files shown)


assets/src/scripts/components/MoreInfoModal.tsx
```diff
@@ -118,9 +119,11 @@ function MoreInfoModal({
             data.synonyms[code].filter((synonym: string) => synonym !== term) ||
               [],
           );
+          setReferences(data.references[code] || []);
```
tomodwyer, line 122: I can't do the suggestion for some reason in the UI, but you'll need to make sure that we're checking for the right data when only synonyms or references are being returned, otherwise the filter may fail:

```jsx
setSynonyms(
  data.synonyms?.[code]?.filter(
    (synonym: string) => synonym !== term,
  ) || [],
);
setReferences(data.references?.[code] || []);
```

Optional chaining the values here will do that.

assets/src/scripts/components/MoreInfoModal.tsx
```diff
@@ -176,6 +178,22 @@ function MoreInfoModal({
...
+            <ul>
+              {!references || references?.length === 0 ? (
+                <li>No references</li>
+              ) : (
+                references.map((reference, idx) => (
+                  <li key={idx}>
+                    <a href={reference[1]}>{reference[0]}</a>
+                  </li>
+                ))
+              )}
```
tomodwyer, line 194: We can update this and the synonyms above to use optional chaining to check the array length, and simplify this comparison:

```suggestion
              {!references?.length ? (
                <li>No references</li>
              ) : (
                references.map((reference, idx) => (
                  <li key={idx}>
                    <a href={reference[1]}>{reference[0]}</a>
                  </li>
                ))
              )}
```

assets/src/scripts/components/MoreInfoModal.tsx
Jongmassey: It's annoying isn't it?

I've integrated those changes manually.

Issue #2671, opened 2025-06-25
Automate and schedule NHS PCD Refset updates


Jongmassey:
# Why are we doing this?

We have user requests to update these periodically, which is a manual process.

The most recent of these is in #2668.

#2250  made improvements to this process, but there still remain manual steps.

We should make this process more automated and schedule it to be run on a regular basis such that these Refsets remain up to date.

# How will we know when it's done?

The NHS Primary Care Domain Reference Sets (PCD Refsets) on Opencodelists in the `NHSD` organisation are kept up to date on an automatic basis.

We monitor the Refset load process such that know they are successfully being kept up to date.

# What are we doing?

- [ ] Integrating with the [Refset portal](https://digital.nhs.uk/data-and-information/data-collections-and-data-sets/data-collections/quality-and-outcomes-framework-qof/quality-and-outcome-framework-qof-business-rules/primary-care-domain-reference-set-portal) or [TRUD](https://isd.digital.nhs.uk/trud/users/guest/filters/2/categories/1/items/659/releases) (with which we have other integrations) to download the refset files.
- [ ] Handle exceptions during the load (e.g. codes in refset not matching the latest release of SNOMED CT - this is often because the Refsets are defined using an earlier release. E.g. The May 2025 Refsets are defined in the 24th Sept SNOMED CT Release)
- [ ] Create a cron job for the load process on a suitable schedule, with appropriate Sentry monitoring.

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

Issue #2163, opened 2024-11-26
Update NHSD refsets


Jongmassey:
A [SNOMED CT Reference Set](https://confluence.ihtsdotools.org/display/DOCGLOSS/SNOMED+CT+reference+set)  (aka "refset") is 

>A standard format for maintaining and distributing a set of references to [SNOMED CT components](https://confluence.ihtsdotools.org/display/DOCGLOSS/SNOMED+CT+component).
>
>Notes
>A reference set can be used to represent a subset of components ([concepts](https://confluence.ihtsdotools.org/display/DOCGLOSS/concept), [descriptions](https://confluence.ihtsdotools.org/display/DOCGLOSS/description) or [relationships](https://confluence.ihtsdotools.org/display/DOCGLOSS/relationship)).
>
>A reference set may also associate referenced components with additional information such as:
Ordered lists of components
Sets of associations between components
Mapping between [SNOMED CT concepts](https://confluence.ihtsdotools.org/display/DOCGLOSS/SNOMED+CT+concept) and other systems codes, classifications, or knowledge resources.

They can be thought of as officially sanctioned "codelists" of SNOMED CT Codes.

NHS Digital publishes its own [Primary Care Domain reference sets](https://isd.digital.nhs.uk/trud/user/guest/group/2/pack/8/subpack/659/releases) which

>are clusters of codes used within the business rules authored and maintained by NHS England's primary care domain.

These are commonly used in QOF rules and have been used by researchers in studies using OpenSAFELY.

We have previously published these refsets as SNOMED CT codelists on OpenCodelists under the [NHSD Primary Care Domain Refsets](https://www.opencodelists.org/codelist/nhsd-primary-care-domain-refsets/) organisation.

The NHS Digital primary care reference sets were last updated in 2023, but the version tag of the underlying lists is 20200812 ([example](https://www.opencodelists.org/codelist/nhsd-primary-care-domain-refsets/cpeptide_cod/20200812/).

The latest version [available from NHSD](https://digital.nhs.uk/data-and-information/data-collections-and-data-sets/data-collections/quality-and-outcomes-framework-qof/quality-and-outcome-framework-qof-business-rules/primary-care-domain-reference-set-portal) is 20240215.

We should probably update these (I had a request from a researcher about this) and develop a process to ensure these are kept up to date.

iaindillingham:
@Jongmassey, could you tell us more about what a _reference set_ is? If you could update the description in-place, rather than add a comment, then even better.

[...]

PR #2716, opened 2025-07-09
Add a script to update the NHS PCD refsets automatically

### LINKED ISSUE #2668: Update NHS PCD Refsets
# Why are we doing this?

Users have asked questions about the up-to-dateness of NHS Primary Care Domain (PCD) Refsets on opencodelists, and in one case we have found a particular list to missing 10 codes vs the latest version ([1](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1750762405410899?thread_ts=1750762405.410899&cid=C07AA91BZD4), [2](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1750421638627129))

This was last run in Dec 2024 - #2163 

The NHS PCD Refsets are up to date on OpenCodelists with those published on the [NHS Portal](https://digital.nhs.uk/data-and-information/data-collections-and-data-sets/data-collections/quality-and-outcomes-framework-qof/quality-and-outcome-framework-qof-business-rules/primary-care-domain-reference-set-portal)

# What are we doing?

Downloading the latest refsets and [running the bulk import script with the NHS PCD Refset config file](https://github.com/opensafely-core/opencodelists/blob/main/codelists/scripts/README.md#nhs-digital-primary-care-domain-refsets)

---

rw251:
Fixes #2668 

Running the script provides an interactive CLI to update the NHS PCD refsets. Previously this was [a manual process](https://github.com/opensafely-core/opencodelists/issues/2163).

This addresses the automation part of #2671 but not the scheduling part.

The commit messages provide more detail as to the exact behaviour of the script. But the key things are:
- we now overwrite the name and description of the codelists as we want them to always match the latest (and sometimes the names change)
- the description now includes the OGLv3 message to comply (we think) with the licence
- any unfound codes (e.g. medication codes in a clinical refset) are automatically omitted and explicitly listed in the description

### Steps when running
The first time this runs we will need to take some extra steps. These will not be necessary for future runs. They also act as a way for a reviewer to test the script. To test properly you would need a version of the database that contains the PCD refsets. You also need a `TRUD_API_KEY` associated with an account that has access to the PCD refsets (https://isd.digital.nhs.uk/trud/users/authenticated/filters/0/categories/8/items/659/releases). It is free to sign up, but takes a day or two to be authorised to access a dataset.

1. Set all PCD refsest to have " (obsolete)" appended to their names:
```bash
just manage shell
```
```python
from codelists.models import Handle
from opencodelists.models import Organisation

org = Organisation.objects.get(name="NHSD Primary Care Domain Refsets")
handles = Handle.objects.filter(organisation=org, is_current=True).exclude(slug__contains="_")
print(f"Found {handles.count()} current handles")

# Update each handle by appending " (obsolete)" to the name
for handle in handles:
    original_name = handle.name
    if not original_name.endswith(" (obsolete)"):
        handle.name = f"{original_name} (obsolete)"
        handle.save()
        print(f"Updated: {original_name} → {handle.name}")
    else:
        print(f"Skipped: {original_name} (already marked as obsolete)")
```
2. Back up your database so we can restore later if it goes wrong
```bash
cp db.sqlite3 db.sqlite3.bak
```
3. Run the update script. This requires both an `API_TOKEN` (for using the opencodelists API), and a `TRUD_API_KEY` to be set.
```bash
API_TOKEN=xxx codelists/scripts/update_pcd_refsets.py --live-run
```
(this will run against localhost by default - live-run is just so it actually makes changes rather than tell you what changes it would make)

4. If any fail, you can rerun but append the cluster ids e.g. to run for just `AST_COD` and `FLU_COD`:
```bash
codelists/scripts/update_pcd_refsets.py --live-run ast_cod flu_cod
```

Here are the four different scenarios and what to expect:

#### Existing refset with the correct slug
http://localhost:7000/codelist/nhsd-primary-care-domain-refsets/ast_cod/
- happy path
- After update all versions, including the latest one, shown correctly.
- The (obsolete) tag will have been removed by the update because names are overwritten

#### New refset that didn't previously exist
http://localhost:7000/codelist/nhsd-primary-care-domain-refsets/mmrvvac1_cod/
- It will have been successfully imported

#### Existing refset that only exists with the wrong slug
http://localhost:7000/codelist/nhsd-primary-care-domain-refsets/acquired-immune-deficiency-syndrome-aids-codes/20241205
- After update the above url will still work, but so will urls with the correct slug of `aids_cod`. If you enter the slug in the url but not the date tag, then the old slug will redirect to the new slug.
- The (obsolete) tag will have been removed by the update because names are overwritten

#### Existing refset that exists with the correct AND the wrong slug
http://localhost:7000/codelist/nhsd-primary-care-domain-refsets/adult-and-child-psoriasis-codes/
http://localhost:7000/codelist/nhsd-primary-care-domain-refsets/psoriasis_cod/

- The latest version will appear under "psoriasis_cod"
- The obsolete label will have been removed by the update because names are overwritten
- The wrong slug will still exist, but now display an (obsolete) message

*NB for future runs we will just need to run the script. The `obsolete` dance was just to fix errors introduced by the previous manual process.*

### DIFF
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -67,6 +67,7 @@ def create_or_update_codelist(
     codes,
     coding_system_database_alias,
     slug=None,
+    new_slug=None,
     tag=None,
     description=None,
     methodology=None,
@@ -77,6 +78,8 @@ def create_or_update_codelist(
     should_publish=False,
 ):
     slug = slug or slugify(name)
+    new_slug = new_slug or slug
+
     references = references or []
     signoffs = signoffs or []
 
@@ -86,7 +89,7 @@ def create_or_update_codelist(
         update_codelist(
             owner=owner,
             name=name,
-            slug=slug,
+            slug=new_slug,
             codelist=codelist,
             description=description,
             methodology=methodology,
@@ -112,7 +115,7 @@ def create_or_update_codelist(
             coding_system_id=coding_system_id,
             coding_system_database_alias=coding_system_database_alias,
             codes=codes,
-            slug=slug,
+            slug=new_slug,
             tag=tag,
             description=description,
             methodology=methodology,
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -401,6 +401,9 @@ def _create_version_with_codes(
             # and want to (a) ignore codes that aren't in the dictionary, (b) add them
             # to the description so it's transparent
             unfound_codes = codes - found_codes
+
+            if not codelist.description:
+                codelist.description = ""
             codelist.description += (
                 "\n\nThis codelist was imported automatically. The following codes "
                 f"were not found in the {codelist.coding_system_cls.name} dictionary "
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -77,19 +77,19 @@ def create_or_update_codelist(
     ignore_unfound_codes=False,
     should_publish=False,
 ):
-    slug = slug or slugify(name)
-    new_slug = new_slug or slug
+    current_slug = slug or slugify(name)
+    slug = new_slug or current_slug
 
     references = references or []
     signoffs = signoffs or []
 
     try:
-        codelist = owner.handles.get(slug=slug).codelist
+        codelist = owner.handles.get(slug=current_slug).codelist
 
         update_codelist(
             owner=owner,
             name=name,
-            slug=new_slug,
+            slug=slug,
             codelist=codelist,
             description=description,
             methodology=methodology,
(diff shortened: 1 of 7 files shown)


codelists/actions.py
Jongmassey, line 1: This seems like a very sensible approach. Could do with adding tests to test_actions.py for the "unfound codes" functionality.

Jongmassey (commented):
This is an improvement to the process that works as described as far as I can tell.

I reckon it could do with some unit tests for the functionality added to actions.py.

I also think the PCD refset update script reinvents some of the existing functionality in the trud utils. I'll add a commit to show how this could be used.

rw251:
@Jongmassey the changes in your commit seemed very sensible so have included them. I've also added unit tests for the 3 extra parameters passed to the create/update methods in `actions.py`.

Jongmassey (approved):
LGTM :shipit: :rocket:

PR #2722, opened 2025-07-11
Display number of included codes in metadata

### LINKED ISSUE #2697: Show number of codes in the metadata bar at the top of the page
# Why are we doing this?

A user has said it would be useful to see how many codes are in each codelist. For published codelists that you don't own, that information is only at the bottom of the table on the "Full list" tab (_showing 1 to n of n entries_).

Users can see at a glance how many codes are in a codelist, so that they can easily compare and evaluate the suitability of a codelist.

# What are we doing?

Extend this metadata bar:

Image

to include the number of codes for the currently selected codelist version.

tomodwyer:
So that users can quickly identify the number of codes in a codelist

Closes #2697

### DIFF
--- a/codelists/views/version.py
+++ b/codelists/views/version.py
@@ -77,5 +77,6 @@ def version(request, clv):
         "can_create_new_version": can_create_new_version,
         "tree_data": tree_data,
         "latest_published_version_url": latest_published_version_url,
+        "count_codes_included": len(rows),
     }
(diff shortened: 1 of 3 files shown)


tomodwyer:
Example screenshot showing the comma in large numbers



[Source URL](https://www.opencodelists.org/codelist/opensafely/ace-inhibitor-medications/2020-05-19/)

PR #2727, opened 2025-07-15
Refactor and rename codelist builder React components

tomodwyer:
The first part of a refactor of the CodelistBuilder application, this PR contains a few parts:

## Minor tweaks

- Use props from Search component in test 
- Simplify the sidebar ternary logic statement
- Check for variable length instead of existence
- Hide builder buttons from anonymous users  
Or users who are logged in but not the creator of the codelist

## Renaming

- `TreeRow` -> `Row`
- `TreeTables` -> `Container`
- `TreeTable` -> `Section`
- `Metadata` and `Title` combined in to `Header`

## File and folder moves

- `Cards/` – used for card components (currently only in the sidebar)
- `Codelist/` – components used on the Codelist tab
- `Header/` – components used in the header section
- `Layout/` – container components for page layout and structure
- `Metadata/` – components used on the Metadata tab

## Coding changes

- Convert TreeTables class to functional component

### DIFF
--- a/assets/src/scripts/components/Layout/Sidebar.tsx
+++ b/assets/src/scripts/components/Layout/Sidebar.tsx
@@ -21,23 +21,16 @@ export default function Sidebar({
 }) {
   return (
     <Col className="builder__sidebar" md="3">
-      {isEmptyCodelist && isEditable ? (
-        <SearchForm />
-      ) : (
-        <>
-          <Summary counts={counts} draftURL={draftURL} />
-          {searches.length > 0 && (
-            <Searches
-              draftURL={draftURL}
-              isEditable={isEditable}
-              searches={searches}
-            />
-          )}
-          {isEditable && <SearchForm />}
-          <Versions />
-        </>
+      {!isEmptyCodelist && <Summary counts={counts} draftURL={draftURL} />}
+      {searches.length > 0 && (
+        <Searches
+          draftURL={draftURL}
+          isEditable={isEditable}
+          searches={searches}
+        />
       )}
-
+      {isEditable && <SearchForm />}
+      {!isEmptyCodelist && <Versions />}
       <Tools />
     </Col>
   );


assets/src/scripts/components/Layout/Sidebar.tsx
```diff
@@ -21,8 +21,8 @@ export default function Sidebar({
 }) {
   return (
     <Col className="builder__sidebar" md="3">
-      {isEmptyCodelist ? (
-        <>{isEditable && <SearchForm />}</>
+      {isEmptyCodelist && isEditable ? (
+        <SearchForm />
       ) : (
```
rw251, line 26: Does this do the same thing? I'm struggling a bit, but if `isEmptyCodelist` is `true` and `isEditable` is `false`, then previously the block would return an empty block `<></>`, but in the change it now returns the bit after the `:` of the ternary operator. Have I read that correctly?

assets/src/scripts/components/Cards/Versions.tsx
```diff
@@ -79,7 +79,7 @@ function Version({ version }: { version: VersionProps }) {
 export default function Versions() {
   const versions: VersionProps[] = readValueFromPage("versions");
 
-  if (!versions) return null;
+  if (!versions?.length) return null;
```
rw251, line 83: If `versions` is an empty array, then `!versions` is false, but `!versions?.length` is true - was that an intentional change?

tomodwyer: Gah, that breaks the old tests 🤦‍♂️ I'll leave it in there for the moment

assets/src/scripts/components/Layout/Sidebar.tsx
tomodwyer: Fixed this in the latest commit

assets/src/scripts/components/Cards/Versions.tsx
rw251: Fine. I was just checking the change in behaviour was intentional - and it was, so great.

rw251 (approved):

PR #2487, opened 2025-05-01
Run OpenCodelists in Dokku in Codespaces

StevenMaude:
See #2483.

This PR is currently a sketch of adding a configuration for running OpenCodelists in Dokku in a dev container/Codespaces. It is not complete enough to be merged, but functional enough to try out. If it's useful, it may warrant some tidying before merging.

It adds a minimal `devcontainer.json` configuration and a script that:

* installs `just`, necessary to build the test databases and Docker image
* installs Dokku via docker-compose (it was tricky to get Dokku working outside of Docker, in the dev container directly)
* builds the OpenCodelists test data
* copies the OpenCodelists tests data into the right place for Dokku
* configures Dokku to run OpenCodelists
* builds the OpenCodelists production Docker image
* deploys that OpenCodelists image with Dokku

You can see the Django logs by running:

```
docker exec -t dokku sh -c "dokku logs opencodelists --tail"
```

PR #2738, opened 2025-07-17
Dokku in a container to simulate production deployments locally

rw251:
Based on the work by @StevenMaude [here](https://github.com/opensafely-core/opencodelists/pull/2487).

If you check out this branch you can simulate deploying production builds locally.

Full instructions in `LOCAL_TEST_DEPLOYMENT.md`.

It still needs tidying up before merging, but works as-is.

Possible improvements:

- [ ] It currently doesn't create any coding_system databases, so although the web app starts, you can only view a couple of pages without throwing errors
- [ ] It works locally in VSCode, but sometimes there are db permission errors when run in codespaces that I haven't investigated.
- [ ] It's perhaps abusing the intent of devcontainers. An alternative would be to do this just with docker, and without using the devcontainer feature.
- [ ] The `local-ci-pipeline.sh` script doesn't work when deploying a remote branch. This is because it relies on things in the justfile in _this_ branch, but when deploying a remote branch it uses the justfile from _that_ branch.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,3 +15,7 @@ node_modules/
 staticfiles/
 # dockerfile image build its own dir
 assets/dist/
+
+# Database files
+**/*.sqlite*
+db.sqlite*


rw251:
This has been superseded by #2774. When that is merged this can be closed.

PR #2729, opened 2025-07-15
Use simpler approach in tests to get response text

StevenMaude:
Fixes #2411.

Reference: https://adamj.eu/tech/2025/04/07/django-whats-new-5.2/#httpresponse-text

PR #2744, opened 2025-07-18
Revert "Use simpler approach in tests to get response text"

Jongmassey:
Reverts opensafely-core/opencodelists#2729


Jongmassey:
Formed part of #2743

PR #2745, opened 2025-07-18
Improve the NHS pcd refset so that it doesn't require config updates

rw251:
In a recent change #2716 we added a script to update the PCD refsets. However this still required a bit of config wrangling, and once the script was run the json config file would need committing. This changes it so that that is no longer required by:

- exposing a PCD refset version on the `latest_releases` page
- using that value to only present newer releases in the CLI
- creating a temp config file for the `bulk_import_script` based on user choices
- prompting the user to see if, following import of the latest refsets, they want to update the latest refset tag

### DIFF
--- a/coding_systems/versioning/tests/test_views.py
+++ b/coding_systems/versioning/tests/test_views.py
@@ -18,3 +18,41 @@ def test_latest_releases(client, setup_coding_systems):
         "readv2",
         "experiment",
     }
+
+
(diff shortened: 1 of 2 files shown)


coding_systems/versioning/models.py
Jongmassey, line 1: There's something unappealing about having something that's not a Coding System live in the `coding_systems` app, but also moving `versioning` out of the `coding_systems` app a pretty unappealing task too!

PR #2751, opened 2025-07-23
Upgrade to JS linter Biome version 2

### LINKED ISSUE #2715: Upgrade to Biome v2
## Why are we doing this?

[Biome v2 has been released](https://biomejs.dev/blog/biome-v2/) and we should upgrade to using it, so that we receive the latest security updates and features.

## How will we know when it's done?

When v2 is used in the schema and for fixing our JS/TS files. When we moved from ESLint to Biome, we brough over a lot of legacy rules. Where possible we should try to simplify the ruleset and/or use the default rules.

## What are we doing?

- Upgrade to v2
- Update the schema to v2
- Remove as many unnecessary rules as possible

---

tomodwyer:
Update to Biome v2, closes #2715

Included in these changes are updates that the linter has identified, see commit history for explanation of each change.

### DIFF
--- a/assets/src/scripts/components/Cards/SearchForm.tsx
+++ b/assets/src/scripts/components/Cards/SearchForm.tsx
@@ -6,8 +6,6 @@ export default function SearchForm() {
   const codingSystemName = readValueFromPage("metadata")?.coding_system_name;
   const searchURL = readValueFromPage("search-url");
 
-  if (!codingSystemName || !searchURL) return null;
-
   // NB, if you change the max search length, remember to change it on
   // the server side as well in the models.py file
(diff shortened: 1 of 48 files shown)


assets/src/scripts/components/Cards/SearchForm.tsx
```diff
@@ -48,6 +46,8 @@ export default function SearchForm() {
     SEARCH_OPTIONS.term.value,
   );
 
+  if (!codingSystemName || !searchURL) return null;
+
```
rw251, line 50: I have no issue with this, but it's not immediately obvious to me what difference the change to the position of this line makes.

PR #2756, opened 2025-07-23
Move metadata editing into modals

### LINKED ISSUE #2748: Disappearing metadata on edit
# Why are we doing this?

It's been [reported](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1753088347849729) that when adding text to both the Description and Methodology sections when building a codelist, then saving one, the text of the other one disappears. 

Note: I haven't attempted to recreate this yet.

The text in the text area that isn't saved is unchanged.

# What are we doing?

...

---

rw251:
Fixes #2748 
This arose from a bug reported by a user where if you tried to edit two metadata fields at the same time, i.e. without clicking save on the first one before moving to the second, then you would lose some changes. This solution is to make all metadata edits go into modals. This means you can only edit one field at a time and so the bug cannot happen. It has the added benefit that you can't tab away from unsaved text - or try to click `Save draft` thinking it would save your metadata changes.

### DIFF
--- a/assets/src/scripts/components/CodelistBuilder.tsx
+++ b/assets/src/scripts/components/CodelistBuilder.tsx
@@ -233,6 +233,12 @@ export default class CodelistBuilder extends React.Component<
     });
   };
 
+  handleFocus = (e: React.FocusEvent<HTMLTextAreaElement>) => {
+    var temp_value = e.target.value;
+    e.target.value = "";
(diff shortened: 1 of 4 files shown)


rw251:
This was actually fixed by #2771 and #2772. A new version of this PR (putting editing into modals) is #2775 so this one can be closed.

PR #2765, opened 2025-07-24
WIP: rework CSV upload checks

StevenMaude:
This is a WIP attempt at having:

* single shared code for both "Create a codelist" and `add/` URL CSV upload forms
* enabling header autodetection for the "Create a codelist" upload route

It is slightly scuppered by me not knowing that `create_old_style_version` also has its own CSV check that mirrors the one that's currently on the `clean_csv_data()` for the `add/` URL.

Most of the existing tests pass, but the version upload view test currently fails.

This makes it more difficult to reuse the mixin, since that function isn't a Django form.

There's a bit of other tidying that would be needed if this is a route to doing all this.

### DIFF
--- a/codelists/tests/test_forms.py
+++ b/codelists/tests/test_forms.py
@@ -118,7 +118,7 @@ def test_codelistform_no_code_header(setup_coding_systems):
     assert len(e.value.messages) == 1
     assert (
         e.value.messages[0]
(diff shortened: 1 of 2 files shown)


StevenMaude:
Closing this one for now as it's not clear when we'll return to this. In the meantime, we added the option for users to select whether a header is present.

It may benefit to see whether we can make the CSV uploading more unified.

PR #2766, opened 2025-07-25
Improve codelist diffing by preventing 500 errors and making assumptions if the rhs is a non unique tag

### LINKED ISSUE #2577: Malformed diff URLs throw 500 errors
Sentry Issue: [OPENCODELISTS-SD](https://ebm-datalab.sentry.io/issues/6565185464/?referrer=github_integration)

```
CodelistVersion.MultipleObjectsReturned: get() returned more than one CodelistVersion -- it returned more than 20!
(1 additional frame(s) were not displayed)
...
  File "codelists/views/decorators.py", line 67, in wrapped_view
    rsp = view_fn(request, version, **view_kwargs)
  File "codelists/views/version_diff.py", line 23, in version_diff
    other_clv = get_object_or_404(CodelistVersion.objects.filter(q))
```

rw251:
Fixes #2577 

No longer throw a 500 error if more than one codelist is matched. Also try to be a bit clever by:
- if the RHS hash_or_tag resolves to a codelistversion id, then diff with that
- if not then we assume the RHS is a tag and if unique diff with that
- if not unique, but the tag is for a version of the LHS codelist, then diff with that
- if none of the above match we 404. We could show a more helpful error e.g. whether it's the lhs or rhs can't be found - but this is better than it was and fixes the bug where previously a user would get a 500 and we'd have an error appear in sentry

### DIFF
--- a/codelists/views/version_diff.py
+++ b/codelists/views/version_diff.py
@@ -1,7 +1,6 @@
 from django.core.exceptions import BadRequest
-from django.db.models import Q
 from django.http import Http404, HttpResponseBadRequest
-from django.shortcuts import get_object_or_404, render
+from django.shortcuts import render
 
 from opencodelists.hash_utils import unhash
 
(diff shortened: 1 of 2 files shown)


codelists/views/version_diff.py
```diff
@@ -12,17 +11,34 @@
 
 @load_version
 def version_diff(request, clv, other_tag_or_hash):
-    q = Q(tag=other_tag_or_hash)
+    other_clv = None
+
+    # 1. If other_tag_or_hash is a hash, try to find the CodelistVersion by ID
     try:
         id = unhash(other_tag_or_hash, "CodelistVersion")
+        qs = CodelistVersion.objects.filter(id=id)
```
Jongmassey, line 19: micronit: I'm guessing `qs` == "QuerySet", I'd be inclined to go for a more meaningful variable name

PR #2770, opened 2025-07-29
Limit number of codes returned by search

### LINKED ISSUE #2733: It is possible to create a search that generates a query with too many SQL variables
Sentry Issue: [OPENCODELISTS-QN](https://ebm-datalab.sentry.io/issues/6057191573/?referrer=github_integration)

```
OperationalError: too many SQL variables
  File "django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  File "django/db/backends/sqlite3/base.py", line 354, in execute
    return super().execute(query, params)

OperationalError: too many SQL variables
(15 additional frame(s) were not displayed)
...
  File "builder/decorators.py", line 18, in wrapped_view
    rsp = view_fn(request, version, **kwargs)
  File "builder/decorators.py", line 36, in wrapped_view
    return view_fn(request, draft, **kwargs)
  File "builder/views.py", line 263, in new_search
    search = actions.create_search(draft=draft, term=term, code=code, codes=codes)
  File "builder/actions.py", line 39, in create_search
    codes_with_existing_code_objs = set(
```

rw251:
Fixes #2733 

If there are too many codes returned we hit a limit in sqlite and an unhandled exception is thrown. This limit is currently 250,000 codes in an `IN` clause. However, large numbers of codes returned by a search is probably a sign the user has made a mistake or has not been specific enough. They would then need to include/exclude all codes which would take too long. So instead we set a much smaller limit (20,000 codes) and tell the user if their query exceeded that.

I have taken a look at existing searches that exceed this limit and have summarised them here. These searches will still exist in the system, but I'm happy that we don't need to support any of them in future.

### Less than 3 characters - i.e. no longer allowed

|Term|Number of codes returned|
|---|---|
|NA|141569|
|PE|123320|
|x|82434|
|a|67427|
|1|21605|

### Words too common in SNOMED

|Term|Number of codes returned|
|---|---|
|finding|149626|
|disorder|146499|
|region|127027|
|clinical findings|92845|
|Procedure by method|60139|
|substance|56723|
|product|55111|
|neck|47479|
|Procedure by site|39292|
|family|36062|
|vision|35465|
|Procedure on organ|34438|
|other|30243|
|assessment|28990|
|function|28132|
|operation|27853|
|vascular|26217|
|poisoning|25407|
|Surgical procedure|24149|
|investigations|24022|
|mental|23923|
|cardiovascular|23717|
|Clinical history and observations|23560|
|accident|22759|
|bacteria|21862|
|appliances|20273|

### Short wildcard strings that are too frequent

|Term|Number of codes returned|
|---|---|
|dia|41686|
|vasc|26342|
|appl|20418|

### Short strings that would be fine if we had exact word matching

|Term|Number of codes returned|
|---|---|
|ace|87099|
|TOM|82966|
|tic|71860|
|age|62339|
|kin|59337|
|late|50919|
|uti|43721|
|ICU|34260|

### Codes

|Code|Term|Number of codes returned|
|---|---|---|
|404684003|Clinical finding|132883|
|64572001|Disease|88388|
|362958002|Procedure by site|39172|

### DIFF
--- a/builder/actions.py
+++ b/builder/actions.py
@@ -56,6 +56,44 @@ def get_codes_to_keep(codeset_version, potential_codes):
 
 @transaction.atomic
 def create_search(*, draft, term=None, code=None, codes):
+    if len(codes) > 20000:
+        # If we have too many codes then we risk breaching the MAX_VARIABLE_SIZE
+        # limit in sqlite in the below queries. However, having too many codes
+        # is also a sign that the search is too broad, and we should probably
+        # not allow it. The current MAX_VARIABLE_SIZE is 250,000, and the default
+        # for sqlite after v3.32.0 (2020-05-22) is 32766, so 20,000 will not hit
+        # that limit.
+        logger.info(
+            "Search request rejected due to large code count",
+            code_count=len(codes),
+            term=term,
+            code=code,
+        )
+
+        message = (
+            (
+                f'Your search for "{term}" returned {len(codes):,} codes, '
+                "which exceeds the maximum limit of 20,000. It's likely that "
+                "you have searched for a very common term, or one that "
+                "appears near the top of the coding system hierarchy. Please "
+                "try to refine your search to return fewer results."
+            )
+            if term
+            else (
+                f'Your search for code "{code}" returned {len(codes):,} codes, '
+                "which exceeds the maximum limit of 20,000. When you search "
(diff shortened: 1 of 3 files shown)


Jongmassey:
Some of these seem like legitimate searches - we might hit some user queries about this in future but I think 20k is a reasonable limit.

PR #2771, opened 2025-07-30
Replace metdata description component in codelist builder

tomodwyer:
The current implementation of the metadata tab requires that all data updates are handled in the CodelistBuilder (class) component. We should simplify this so that the individual components can handle their own state, and POST changes to the server. This PR is the first of a number of PRs to simplify the component architecture.

In this PR I have:
- Created a new MetadataForm component
    - Used that component for the description field
    - Styled it using Bootstrap card classes (as opposed to react-bootstrap, which contains many context providers)
- Used react-query to hold the state of the metadata, and update it upon the POST request to the server
- Added tests which use react-query for testing, and nock for mocking API requests
- Removed any tests for the existing description functionality
- Prepared the MetadataForm component to be used for the Methodology field
- Added types as capitalised constants for data passed directly from the Django app to the React app (where the data is not changed in any way)

## Screenshot

Screenshot showing the metadata tab

## Video

https://github.com/user-attachments/assets/65e2a179-1800-4acc-86d6-2ec1fd891377

### DIFF
--- a/assets/src/scripts/components/Layout/MetadataTab.tsx
+++ b/assets/src/scripts/components/Layout/MetadataTab.tsx
@@ -1,7 +1,10 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 import React from "react";
 import { Form } from "react-bootstrap";
 import type { PageData, Reference } from "../../types";
 import type { MetadataFieldName } from "../CodelistBuilder";
+import MetadataForm from "../Metadata/MetadataForm";
 import ReferenceList from "../Metadata/ReferenceList";
 
 interface MetadataTabProps {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -34,6 +34,8 @@
     "@tailwindcss/forms": "^0.5.10",
     "@tailwindcss/typography": "^0.5.16",
     "@tailwindcss/vite": "^4.1.11",
+    "@tanstack/react-query": "^5.83.0",
+    "@tanstack/react-query-devtools": "^5.83.0",
```
rw251, line 39: Can the devtools go in the `devDependencies`?

assets/src/scripts/components/Layout/MetadataTab.tsx
```diff
@@ -1,7 +1,10 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
```
rw251, line 2: Ah - so yes in it's current form we need the devtools in `dependencies`. Is there a way to only include it for local development? Or is this the best practice way?

tomodwyer: > By default, React Query Devtools are only included in bundles when `process.env.NODE_ENV === 'development'`, so you don't need to worry about excluding them during a production build.

It freaks me out every time, [but the docs say this is the way to do it](https://tanstack.com/query/v5/docs/framework/react/devtools).

tomodwyer:
@rw251 yeah it looks like when the methodology state changes, it resets a lot of the class component state which is causing issues. I think that's the same issue, which will be resolved in the next PR switching over the methodology, as you said.

rw251 (approved):

PR #2774, opened 2025-08-01
Local deployment simulation

rw251:
**UPDATE - I'm putting this back in triage as I tried to quickly remove the `--priveleged` flag but now realise it might become a bit of a rabbit hole so I've stepped away.**

Add a just command (`just local-deploy`) to simulate deployment locally.

Full explanation of usage, architecture etc. in `./deploy/local/LOCAL_DEPLOYMENT_TESTING.md`.

This _shouldn't_ make changes to your local database or environment so is safe to test repeatedly, and I have done locally. But I can't rule out that under different circumstances it might, so good for a few people to test locally and to backup anything in your dev environment (like the db) that you don't want to lose.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -435,3 +435,7 @@ PRs: first, a PR with models and migrations to create the new table or column
 and replicate existing data; then a PR updating the application code to use the
 new table/column; and finally, a PR to remove the old field or model and the
 corresponding migration.
+
+## Testing deployments locally
+
+Please refer to the [local deployment document](./deploy/local/LOCAL_DEPLOYMENT_TESTING.md) for instructions on how to test deployments locally.


deploy/local/container-creation.sh
```diff
@@ -0,0 +1,48 @@
...
+check_docker_installed
+
+run_container() {
+    log "Creating container '$CONTAINER_NAME'..."
+    docker run -d \
+        --name "$CONTAINER_NAME" \
+        -v "$(pwd):$WORKSPACE_DIR" \
+        -w "$WORKSPACE_DIR" \
+        -p "$HOST_PORT:7000" \
+        --privileged \
```
StevenMaude, line 16: thought: `--privileged` isn't [great from a security perspective](https://docs.docker.com/reference/cli/docker/container/run/#privileged). It is necessary in this case as a way of allowing the container to run Docker, which is what we want:

> This flag exists to allow special use-cases, like running Docker within Docker.…
>
> …Use the `--privileged` flag with caution. A container with `--privileged` is not a securely sandboxed process. Containers in this mode can get a root shell on the host and take control over the system.

We don't control all of the code running inside the containers here either.

* It's possible to bind mount the Docker socket instead to allow the container to run Docker, but that also has risks.
* I think it's possible to run Podman rootless inside a container, but that's also more work, and not sure if anyone at Bennett has tried that.
* Maybe there's some other combination of command options that still allows this but mitigates the risk?

It's probably worth asking Simon to review this, and offer any suggestions.

deploy/local/container-deploy.sh
```diff
@@ -0,0 +1,182 @@
...
+    deploy_application
+    echo -e "
+╔═════════════════════════════════════════════════════════════════════════════╗
+║ 🚀 Your current branch is now deployed locally.
+║"
+else
+    log "No deployment needed."
+        echo -e "
+╔═════════════════════════════════════════════════════════════════════════════╗
+║ 🚀 Which ever branch you were running last is now deployed locally.
```
StevenMaude, line 180: nit:

```suggestion
║ 🚀 Whichever branch you were running last is now deployed locally.
```

deploy/local/LOCAL_DEPLOYMENT_TESTING.md
```diff
@@ -0,0 +1,197 @@
...
+# Check inner container logs for the opencodelists app
+docker exec -t opencodelists-local-deployment bash -c "docker logs opencodelists.web.1"
+
+# Check inner container logs for the dokku container
+docker exec -t opencodelists-local-deployment docker logs dokku
+```
+
+## TODO
+
+- These scripts don't do any of the CI stuff from the main.yml and deploy.yml files. A lot of that is linting and testing which can be achieved with `just` commands. But there are some things like the smoke-test which may fail. Currently the smoke-test will fail as soon as you push your branch, so it will be caught before deployment. Therefore it's not essential to replicate that here, but it might be useful to add in the future.
```
StevenMaude, line 196: thought: I think this is fine for now. The benefit is that we can least try out the deployment.

StevenMaude (commented):
summary: This is close to done now.

I tried this out, and it worked fine for me. I've a couple of small comments, and a bigger question about the risks of the Docker setup.

deploy/local/container-creation.sh
rw251: I had a bit of a look at this and was able to remove `--priveleged` and mount the docker socket instead. It kind of worked but there were then other issues - largely because other bits of the scripts were working only because of that flag, and so needed changing. Probably no blockers, but it was beginning to spiral so I'm pulling out and putting back in your triage pile. I have some code in a local branch but pretty messy and the main bit is:
```python
local DOCKER_GID
DOCKER_GID=$(stat -c '%g' /var/run/docker.sock)

log "Creating container '$CONTAINER_NAME' with Docker GID $DOCKER_GID..."
docker run -d \
    --name "$CONTAINER_NAME" \
    -v "$(pwd):$WORKSPACE_DIR" \
    -v "/var/run/docker.sock:/var/run/docker.sock" \
    --group-add "$DOCKER_GID" \
    -w "$WORKSPACE_DIR" \
    -p "$HOST_PORT:7000" \
    "$CONTAINER_IMAGE" \
    sleep infinity
```
I also needed to change how and where the database file was created because that became inaccessible without the flag.

PR #2778, opened 2025-08-04
Put add/edit references into a modal for metadata in the draft view

rw251:
Brings the references in the metadata (in draft view) in line with the description and methodology, by having edits take place in a modal. #2775 was the PR for doing this for description and methodology. The reasons are the same, repeating them here for completeness:

- users can't forget to click save which was possible before this change as you could tab away with no warning
- also reduces confusion as you don't have the choice of "Save" for the field and "Save draft". Users might think "Save draft" would also save their metadata edits.

Within this change I've also added:
- autofocus the first field in the modal
- add a key handler so escape closes the modal - similar to the other metadata fields

### DIFF
--- a/assets/src/scripts/components/Metadata/Reference.tsx
+++ b/assets/src/scripts/components/Metadata/Reference.tsx
@@ -2,7 +2,7 @@ import { useMutation, useQueryClient } from "@tanstack/react-query";
 import React, { useState } from "react";
 import { postFetchWithOptions, readValueFromPage } from "../../_utils";
 import type { IS_EDITABLE, METADATA } from "../../types";
-import ReferenceForm from "./ReferenceForm";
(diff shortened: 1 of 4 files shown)


tomodwyer:
Could you add a test for the escape key code? Coverage has dropped in the JS tests:

```
-----------------------------|---------|----------|---------|---------|-------------------------------
File                         | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-----------------------------|---------|----------|---------|---------|-------------------------------
 scripts/components/Metadata |   99.01 |    98.33 |   96.15 |   99.01 |
  ReferenceFormModal.tsx     |   94.91 |    85.71 |     100 |   94.91 | 23-25
```

tomodwyer (approved):

PR #2788, opened 2025-08-06
Replace React Bootstrap with HTML classes

tomodwyer:
Where the only thing being applied is styling. This reduces the amount of components and context providers in the React component tree.

Images below from the React Dev Tools

### DIFF
--- a/assets/src/scripts/components/Layout/Header.tsx
+++ b/assets/src/scripts/components/Layout/Header.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { Badge, ListGroup } from "react-bootstrap";
 import type { PageData } from "../../types";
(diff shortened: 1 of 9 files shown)


assets/src/scripts/components/Layout/Header.tsx
```diff
@@ -18,46 +17,46 @@ export default function Header({
...
             Draft
-          </Badge>
+          </span>
         </h1>
         {isEditable && <ManagementForm counts={counts} />}
       </div>
 
-      <ListGroup horizontal className="small">
-        <ListGroup.Item className="py-1 px-2">
+      <dl className="small list-group list-group-horizontal small">
```
rw251, line 27: `small` appears twice

rw251 (approved):

PR #2803, opened 2025-08-13
Update the documentation following the CSV header option addition

StevenMaude:
This PR:

* Adds a `take_screenshots` context processor to toggle the feedback form button when taking screenshots; the button was appearing over text in a couple of screenshots.
* Updates the documentation for the new CSV header upload option.
* Rebuilds the screenshots; some have been updated incidental to this PR.


opencodelists/settings.py
```diff
@@ -162,6 +164,11 @@ def get_env_var(name):
     },
 ]
 
+if TAKE_SCREENSHOTS:
```
lucyb, line 167: What's the reasoning for conditionally including the context processor rather than always including it? It means the `if TAKE_SCREENSHOTS` is duplicated in two places (the settings.py and in the context processor itself). It might be clearer just to always include it, unless I've missed something?

opencodelists/settings.py
```diff
@@ -81,6 +81,8 @@ def get_env_var(name):
     SESSION_COOKIE_SECURE = True
     CSRF_COOKIE_SECURE = True
 
+TAKE_SCREENSHOTS = os.environ.get("TAKE_SCREENSHOTS", default=False) == "True"
```
lucyb, line 84: I like the use of a specific env var for taking screenshots rather than trying to overload IN_PRODUCTION. I also appreciate that you added your reasoning in the commit message.

lucyb (approved):
A very neat change, with well explained commit messages. Just one comment to respond to.

opencodelists/settings.py
StevenMaude: Thanks for spotting that.

I don't think there is a reason, other than me probably misreading how the `IN_PRODUCTION` context processor worked, and being overzealous in trying to add a toggle.

opencodelists/settings.py
StevenMaude: Thanks. I thought it was quite reasonable that you may want to run locally, and still have the feedback button operational.

Sticking everything into `IN_PRODUCTION` wouldn't allow that.

PR #2818, opened 2025-08-22
Jongmassey/fix clone org codelist same name

### LINKED ISSUE #2817: Handles can collide when re-cloning codelists from another owner
Sentry Issue: [OPENCODELISTS-TB](https://ebm-datalab.sentry.io/issues/6828276886/?referrer=github_integration)

```
IntegrityError: UNIQUE constraint failed: codelists_handle.user_id, codelists_handle.name
  File "django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  File "django/db/backends/sqlite3/base.py", line 354, in execute
    return super().execute(query, params)

IntegrityError: UNIQUE constraint failed: codelists_handle.user_id, codelists_handle.name
(20 additional frame(s) were not displayed)
...
  File "codelists/views/decorators.py", line 40, in wrapped_view
    return view_fn(request, codelist)
  File "codelists/views/codelist_clone.py", line 15, in codelist_clone
    cloned_codelist = actions.clone_codelist(codelist, request.user)
  File "codelists/actions.py", line 812, in clone_codelist
    cloned_codelist = _create_codelist_with_handle(
  File "codelists/actions.py", line 234, in _create_codelist_with_handle
    owner.handles.create(codelist=codelist, slug=slug, name=name, is_current=True)
```

Jongmassey:
Fixes #2817

### DIFF
--- a/codelists/tests/test_actions.py
+++ b/codelists/tests/test_actions.py
@@ -792,3 +792,14 @@ def test_re_clone_organisational_codelist(new_style_codelist, user):
 
     assert new_style_codelist.name in clone2.name
     assert "clone" in clone2.name
+
+
+def test_ten_clone_limit(user_codelist, user):
+    with pytest.raises(
+        ValueError, match="No more than 10 clones of a codelist can be created."
+    ):
+        n = 0
+        while n <= 11:
+            actions.clone_codelist(user_codelist, user)
+            n += 1
+    assert n == 11
--- a/codelists/tests/test_actions.py
+++ b/codelists/tests/test_actions.py
@@ -781,3 +781,14 @@ def test_clone_codelist_user_own_codelist(user_codelist, organisation_user):
 
     assert user_codelist.name in clone2.name
     assert "clone1" in clone2.name
+
+
+def test_re_clone_organisational_codelist(new_style_codelist, user):
+    cloned_codelist = actions.clone_codelist(new_style_codelist, user)
+
+    assert new_style_codelist.name in cloned_codelist.name
+
+    clone2 = actions.clone_codelist(new_style_codelist, user)
+
+    assert new_style_codelist.name in clone2.name
+    assert "clone" in clone2.name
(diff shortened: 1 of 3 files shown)


codelists/tests/test_actions.py
```diff
@@ -781,3 +781,14 @@ def test_clone_codelist_user_own_codelist(user_codelist, organisation_user):
 
     assert user_codelist.name in clone2.name
     assert "clone1" in clone2.name
+
+
+def test_re_clone_organisational_codelist(new_style_codelist, user):
```
lucyb, line 786: I tried running this test against the old code, just to check that it would fail, and it did but with a slightly different error message compared to the original sentry error. 

With the failing unit test, I saw:
`django.db.utils.IntegrityError: UNIQUE constraint failed: codelists_handle.user_id, codelists_handle.slug`

However, in Sentry the error is:
`IntegrityError: UNIQUE constraint failed: codelists_handle.user_id, codelists_handle.name`

i.e one is complaining about the slug and one is complaining about the name. Is this definitely solving the same problem?

Jongmassey: e.g. 

```py
if (
        Handle.objects.filter(user=new_owner, name=name).exists()
        | Handle.objects.filter(user=new_owner, slug=codelist.slug).exists()
    ):
```

Jongmassey: @lucyb  any thoughts?

lucyb: Thinking about it, it's possible my local database is in a weird state, I've not updated it recently and if you're getting the expected error then it's probably fine.

The extra check is probably fine, although I guess there's a slightly increased risk of an infinite loop here maybe?
```
        n = 0
        while True:
            try:
                new_name = f"{codelist.name} (clone{str(n) if n > 0 else ''})"
                Handle.objects.get(user=new_owner, name=new_name)
            except Handle.DoesNotExist:
                name = new_name
                break
            n += 1
```
I appreciate this hasn't changed in this PR, but would it be worth putting an upper limit on n?

lucyb: Oh sorry, I misread the code! Yes, you're absolutely right. Although, it's probably good that we find out if people have been creating lots of cloned codelists.

lucyb (approved):

PR #2835, opened 2025-09-10
Fix API to return codelists with more than 1 handle

### LINKED ISSUE #2829: Fix api to return codelists with multiple handles
**NB the PCD refsets should not be updated until this is fixed**

# Why are we doing this?

The api only returns codelists with exactly one handle, but it should be valid for a codelist to have multiple handles. E.g. before the latest upload of PCD refsets there were some codelists with the wrong handles (the slugs should all be short codes like `ast_cod` but some had their names slugified like `asthma-codes`. By adding a new handle for these codelists with the corrected slug it means we can more easily update them in future while maintaining the the old URLs - which still work by redirecting to the latest slug.

The api will return all codelists with 1 or more handles. A call to https://www.opencodelists.org/api/v1/codelist/nhsd-primary-care-domain-refsets/ will return 959 codelists rather than 763 at present.

# What are we doing?

This code was changed to ensure a codelist had a handle: 

https://github.com/opensafely-core/opencodelists/blob/47deda127862e9d88056ad0edabdccbdad0a36e9/codelists/api.py#L145-L151

but means that anything with more than 1 handle is filtered out. Instead we could make use of the `is_current` field because even through a codelist may have multiple handles, only one will have `is_current==1`.

rw251:
Fixes #2829 

A previous fix to avoid errors from codelists with no handles added a filter to the api to ensure only codelists with a single handle were returned. In fact a codelist can have multiple handles. This corrects the filter to limit to codelists with a `current` handle which better reflects the underlying models.

### DIFF
--- a/codelists/api.py
+++ b/codelists/api.py
@@ -17,7 +17,6 @@ requiring a new version.
 import json
 import re
 
-from django.db.models import Count
 from django.http import JsonResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_http_methods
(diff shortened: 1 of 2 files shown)


StevenMaude:
Thanks!

PR #2840, opened 2025-09-15
Use slug and name to drive unique handle generation for clones

Jongmassey:
Further fix for #2817

### DIFF
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -806,10 +806,12 @@ def clone_codelist(codelist, new_owner):
         while True:
             if n > 10:
                 raise ValueError("No more than 10 clones of a codelist can be created.")
-            try:
-                new_name = f"{codelist.name} (clone{str(n) if n > 0 else ''})"
-                Handle.objects.get(user=new_owner, name=new_name)
-            except Handle.DoesNotExist:
+            new_name = f"{codelist.name} (clone{str(n) if n > 0 else ''})"
+            new_slug = slugify(new_name)
+            if not (
+                Handle.objects.filter(user=new_owner, name=new_name).exists()
+                | Handle.objects.filter(user=new_owner, slug=new_slug).exists()
+            ):
                 name = new_name
                 break
             n += 1


codelists/tests/test_actions.py
```diff
@@ -803,3 +803,18 @@ def test_ten_clone_limit(user_codelist, user):
...
+
+    # rename the clone but don't change the slug
+    cloned_handle = cloned_codelist.current_handle
+    cloned_handle.name = "renamed clone"
+    cloned_handle.save()
+
+    # re-clone the first codelist
+    re_cloned_codelist = actions.clone_codelist(user_codelist, user)
+    assert "clone" in re_cloned_codelist.slug
+    assert "clone" in re_cloned_codelist.name
```
StevenMaude, line 820: nit: We could *possibly* check for the existence of all named clones, a bit like the ten clone limit above, and that each one is correctly named. I'm not sure how important that is though.

codelists/actions.py
```diff
@@ -806,10 +806,12 @@ def clone_codelist(codelist, new_owner):
         while True:
             if n > 10:
                 raise ValueError("No more than 10 clones of a codelist can be created.")
```
StevenMaude, line 808: question: Unrelated to this PR, but I was looking at this code. I haven't run it, and I may be being very silly, but is there an off-by-one error here?

Let's assume all the possible clones already exist and we loop until reaching this exception.

1. n starts at 0.
2. The first clone is `clone` with no numeric suffix.
3. n increments by 1 if the new name or slug exists.
4. We get `clone1` … `clone10` because when n is 10 we don't raise an exception (we stop when n is 11).
5. `clone` plus `clone1`…`clone10` is 11, by my count.

StevenMaude (approved):

codelists/tests/test_actions.py
Jongmassey: I think there's adequate coverage of the clone naming in the other tests. This test is for a very specific circumstance and I've tried to constrain the test to that!

PR #2843, opened 2025-09-18
Remove litecli dev dependency

lucyb:
We're not using litecli anywhere. I believe it was installed as a convenience for developers. However, it pulls in a large number of dependencies that we then need to maintain. I'm removing it and recommending that people install a tool of their choice for interacting with sqlite dbs.


tomodwyer:
[Associated thread in Slack](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1758125979037709)

PR #2903, opened 2025-11-17
Upgrade to Django 5.2 (after fixing prefetch issue)

### LINKED ISSUE #2747: Django 5.2 -  upgrade issues
# Why are we doing this?

#2713 resulted in an [incident](https://docs.google.com/document/d/1nu9FRBBvTC83bw7m0C_VsPyzCIWvik-GCQuTVG2Z3gY) where several codelists returned HTTP 50x errors.

The error that resulted in a HTTP 500 being thrown was an SQLite error `Expression tree is too large (maximum depth 1000)`  when [fetching](https://github.com/opensafely-core/opencodelists/blob/main/codelists/presenters.py#L3) `Search`es and their related `SearchResult`s and `CodeObj`s.

We can assume that the SQL generated by Django 5.1.x is different for this than for 5.2.x, we should investigate why this is the case so we can proceed with the 5.2 upgrade more safely in future.

We understand the root cause of why the Django 5.2 upgrade failed.
# What are we doing?

Investigating the difference in SQL queries generated for the affected method between Django 5.1 and 5.2

---

mikerkelly:
Fixes #2747 

I found that upgrading to Django 5.2 caused the query used with the nested `select_related` exploded into a very long series of `OR` clauses that exceeded database limits. Using `prefetch_related` is more natural based on the analysis below and resolves that issue with upgrading to 5.2.

More info below and in the issue description (more or less the same).

### DIFF
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -452,9 +452,9 @@ distlib==0.4.0 \
     --hash=sha256:9659f7d87e46584a30b5780e43ac7a2143098441670ff0a49d5f9034c54a6c16 \
     --hash=sha256:feec40075be03a04501a973d81f633735b4b69f98b05450592310c0f401a4e0d
     # via virtualenv
-django==5.1.14 \
-    --hash=sha256:2a4b9c20404fd1bf50aaaa5542a19d860594cba1354f688f642feb271b91df27 \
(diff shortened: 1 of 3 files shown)


StevenMaude (approved):
Looks good to me; thank you for the thorough investigation and detailed explanation.

PR #2911, opened 2025-11-27
Fix cached version hierarchy comparison

### LINKED ISSUE #2912: Deserialise cached hierarchies with `list` not `set` to be order-invariant
# Why are we doing this?

A Hierarchy contains a dict which contains collections of codes stored as sets.

We cache the hierarchy as a json string for performance reasons.

This cached version is deserialised and used for comparisons to ascertain whether the CodelistVersion it belongs to is compatible with various releases of the underlying coding system.
The standard json deserialisation uses lists not sets.
This makes the comparison of these collections of codes order-sensitive, which is not required or appropriate.

Hierarchies are compared using order-insensitive sets rather than order-sensitive lists regardless of their provenance.

# What are we doing?

Derserialising cached hierarchies in such as way as to match the typing of the original source Hierarchy

---

Jongmassey:
A `Hierarchy` contains a dict which contains collections of codes stored as `set`s.

We cache the hierarchy as a json string for performance reasons.

This cached version is deserialised and used for comparisons to ascertain whether the `CodelistVersion` it belongs to is compatible with various releases of the underlying coding system.
The standard `json` deserialisation uses `list`s not `set`s.
This makes the comparison of these collections of codes order-sensitive, which is not required or appropriate.

This change provides a custom deserialiser for cached hierarchies which uses `set`s in place of `list`s such that these comparisons are order-insensitive.

Fixes #2912

### DIFF
--- a/codelists/hierarchy.py
+++ b/codelists/hierarchy.py
@@ -31,6 +31,21 @@ class Hierarchy:
         self._descendants_cache = {}
         self._ancestors_cache = {}
 
+    def __eq__(self, other):
+        """Compare hierarchies excluding cache items, resolving possible deserialisation differences"""
+
+        def resolve_deser_nulls(_dict):
+            return {k or "null": {i or "null" for i in v} for k, v in _dict.items()}
+
+        return (
+            self.root == other.root
+            and self.nodes == other.nodes
+            and resolve_deser_nulls(self.child_map)
+            == resolve_deser_nulls(other.child_map)
+            and resolve_deser_nulls(self.parent_map)
+            == resolve_deser_nulls(other.parent_map)
+        )
+
     @classmethod
     def from_codes(cls, coding_system, codes):
         """Build a hierarchy containing the given codes, and their ancestors/descendants
--- a/coding_systems/base/import_data_utils.py
+++ b/coding_systems/base/import_data_utils.py
@@ -1,4 +1,3 @@
-import json
 import shutil
 from itertools import islice
 from pathlib import Path
@@ -339,11 +338,6 @@ def _check_version_by_hierarchy(coding_system, version):
     and identical hierarchies.
     """
 
-    def _hierarchy_excl_cache_items(hierarchy):
-        # `_descendants_cache` and `_ancestors_cache` will not have been generated for the
-        # new hierarchy
-        return {k: v for k, v in hierarchy.items() if not k.endswith("_cache")}
-
     # All codelist versions that existed prior to implementation of coding system releases
     # have an "unknown" coding system release, which represents the coding system data at that
     # point. For older codelist versions, that is not the actual coding system release that was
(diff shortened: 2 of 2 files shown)


evansd:
Arghh, such a subtle nasty bug! Nice work in tracking this down.

This is just a drive-by comment to say that an alternative fix would be to use `pickle` instead of JSON for this which will naturally preserve the native Python types. There are downsides to that of course in that it's not human readable and you cant' easily work with it outside of Python. But depending on exactly how these cached values are used it might be an option.

You could do an incremental migration here by checking the first character of the serialised data for `{` and using json or pickle to deserialize accordingly. And then update all the cached values to use pickle, and then remove the backwards compatibility code.

That's obviously more work than the approach you've taken here, but possibly solves the issue in a more general way.

Anyway, this was just to flag that as an option for consideration, not a recommendation.

coding_systems/base/tests/test_import_data_utils.py
```diff
@@ -323,3 +324,46 @@ def test_save_codelist_draft_updates_compatibility_multiple_releases(self):
...
+            self.bnf_review_version_with_search.cached_hierarchy.data
+        )
+        modified_existing_cached_hierarchy["nodes"] = list(
+            reversed(modified_existing_cached_hierarchy["nodes"])
+        )
+        modified_existing_cached_hierarchy["child_map"] = {
+            k: list(reversed(v))
+            for k, v in modified_existing_cached_hierarchy["child_map"].items()
+        }
+        modified_existing_cached_hierarchy["parent_map"] = {
```
lucyb: Do you need to modify the nodes, child_map and parent_map in order for this test to fail? i.e is there a simpler way of achieving the same thing?

Jongmassey: I'd like at least nodes and one of the maps to ensure that the recursive function is properly casting everything it needs to to a set

Jongmassey:
> I'm wondering how Hierarchy.from_cache and Hierarchy.data_for_cache fits into this? 

Very reasonable questions.

`data_for_cache` does indeed use `lists` not `sets` because:

```py
>>> import json
>>> foo = {"x":{1,2,3}}
>>> json.dumps(foo)
TypeError: Object of type set is not JSON serializable
```


`Hierarchy.from_cache` does indeed do the proper type conversion and is used elsewhere. I'd somewhat forgotten about that in this PR. I wonder if it could be leveraged instead of the additions I've made. It returns a `Hierarchy` rather than a `dict` as  needed in the import utils, but maybe `__dict__` could do the job. I'll have a poke at it.

Jongmassey:
I've refactored to use the existing methods.

I've elected to override `Hierarchy`'s `__eq__` - AFAICT this isn't used anywhere else, and all tests pass but aware that this carries some risk and could equally be implemented as a standalone instance method.

As for performance, the most expensive test for the version compatibility updates is `test_save_codelist_draft_updates_compatibility_multiple_releases`  and it doesn't seem to have negatively affected performance (checked with multiple runs):
dd52c069   4.90s, 5.01s, 4.98s (before change)
33d6d0e1   4.84s, 5.00s, 4.87s

PR #2973, opened 2026-02-10
ICD 9 support prototype

Jongmassey:


Jongmassey:
Closed as decided not to progress with icd9 support

PR #2977, opened 2026-02-16
Fix deployment SHA retrieval for opencodelists to fix the deploy workflow

nishtha-kalra:
On Thursday Feb 12, the deploy workflow for JobServer started failing. On Friday Feb 13, RAP team noticed that the workflow for bennettbot started failing too. They investigated the issue and found that the order of RepoDigests changed and was unreliable. Deploy was always getting the first one which wasn't correct.
Replacing that logic with getting all repo digests and then selecting the fully qualified name worked for bennettbot in commit bennettoxford/bennettbot@970e970.

This commit does the same for opencodelists in order to fix the failing deploy workflow.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -83,7 +83,6 @@ jobs:
 
       - name: Deploy
         run: |
-            set -x
             SHA=$(docker inspect --format='{{join .RepoDigests "\n"}}' "$PUBLIC_IMAGE_NAME:latest" | grep "$PUBLIC_IMAGE_NAME")
             echo "Deploying $SHA to dokku3"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -83,7 +83,9 @@ jobs:
 
       - name: Deploy
         run: |
-            SHA=$(docker inspect --format='{{index .RepoDigests 0}}' $PUBLIC_IMAGE_NAME:latest)
+            set -x
```
lucyb: Very minor point, but I don't think we need both the `set -x` and the `echo` command below. I noticed when watching the Job Server deployment that they're essentially duplicating the same thing. The `echo` is probably clearer though, so maybe we can just stick with that?

PR #3026, opened 2026-04-13
Fix rendering of cloned codelist name underscores

### LINKED ISSUE #2980: Bug report: formatting of cloned codelist source [tiny bug!]
# Summary
When a codelist is cloned the source text is inserted into metadata but appears to be formatted as markdown rather than keeping the symbols such as underscores. [[Example](https://www.opencodelists.org/codelist/user/helen-curtis/delivery_codes/23333edf/)]

Image

# What did you do and what did you expect to happen?

Expected text to be kept as it appears in the source codelist name but underscores are lost.

# Impact of the bug
Minor impact! But looks silly / hard to read if a lot of underscores are lost and words merged together.

KatieB5:
Closes #2980

A user reported a small bug that caused cloned codelist names containing underscores (e.g. "Codelist_name_example") to render with unintended italics in the Methodology section of the About tab of the codelist version page (original issue contains user's screenshot of this). This occurred because the name was interpreted as
[markdown](https://www.markdownguide.org/basic-syntax/#italic), when we pass the `codelist.methodology` (that we set [here](https://github.com/opensafely-core/opencodelists/blob/bc3b482376e2580dff46d90b86b2657fc6bb18d5/codelists/actions.py#L811)) through `markdown_filter` in the `_about_tab.html` template (as far as I understand it).

This PR introduces a [custom template filter](https://docs.djangoproject.com/en/6.0/howto/custom-template-tags/#writing-custom-template-filters) to escape underscores in cloned codelist names within `codelist.methodology` prior to markdown rendering, ensuring names are displayed correctly with underscores preserved.

The fix is applied at render time to keep the fix scoped to presentation logic rather than, for example, changing what we store in `codelist.methodology` when we clone a codelist. 

**Screenshot with template filter applied:**
cloned_codelist_name_after

**Notes for the reviewer:**

- I struggled with sensible/reasonable names for my functions and variables in this PR so if you have any alternative suggestions I'd be grateful for them

### DIFF
--- a/opencodelists/tests/templatetags/test_markdown_filter.py
+++ b/opencodelists/tests/templatetags/test_markdown_filter.py
@@ -1,0 +1,46 @@
+import pytest
+
+from opencodelists.templatetags.methodology_cloned_codelist_name_filter import (
+    methodology_cloned_codelist_name_filter,
+)
+
+
+@pytest.mark.parametrize(
+    "input,expected",
+    [
+        (
+            "Cloned from codelist : [Codelist_to_be_cloned](/codelist/user/katieb5/codelist_to_be_cloned/2ed738d9/)\r\nSome additional methodology text",
+            "Cloned from codelist : [Codelist\\_to\\_be\\_cloned](/codelist/user/katieb5/codelist_to_be_cloned/2ed738d9/)\r\nSome additional methodology text",
+        ),
+        (
+            "Cloned from codelist : [Codelist_to_be_cloned](/codelist/user/katieb5/codelist_to_be_cloned/2ed738d9/)\r\nSome_additional_methodology_text_with_underscores",
+            "Cloned from codelist : [Codelist\\_to\\_be\\_cloned](/codelist/user/katieb5/codelist_to_be_cloned/2ed738d9/)\r\nSome_additional_methodology_text_with_underscores",
+        ),
+        (
+            "Cloned from codelist : [ClonedCodelistWithoutUnderscores](/codelist/user/katieb5/clonedcodelistwithoutunderscores/2ed738d9/)\r\nSome additional methodology text",
+            "Cloned from codelist : [ClonedCodelistWithoutUnderscores](/codelist/user/katieb5/clonedcodelistwithoutunderscores/2ed738d9/)\r\nSome additional methodology text",
+        ),
+        (
+            "Methodology text without cloned codelist",
+            "Methodology text without cloned codelist",
+        ),
+        (
+            "",
+            "",
+        ),
+        (
+            "Some other text. Cloned from codelist : [Codelist_to_be_cloned](/codelist/user/katieb5/codelist_to_be_cloned/2ed738d9/)\r\nSome additional methodology text",
+            "Some other text. Cloned from codelist : [Codelist\\_to\\_be\\_cloned](/codelist/user/katieb5/codelist_to_be_cloned/2ed738d9/)\r\nSome additional methodology text",
+        ),
+    ],
+    ids=[
+        "cloned_codelist_with_underscores",
+        "cloned_codelist_with_underscores_and_underscored_following_text",
+        "cloned_codelist_without_underscores",
+        "no_cloned_codelist_prefix",
+        "empty_string",
+        "cloned_codelist_not_at_start_of_text",
+    ],
+)
+def test_methodology_cloned_codelist_name_filter(input, expected):
+    assert methodology_cloned_codelist_name_filter(input) == expected
--- a/opencodelists/templatetags/methodology_cloned_codelist_name_filter.py
+++ b/opencodelists/templatetags/methodology_cloned_codelist_name_filter.py
@@ -1,0 +1,26 @@
+import re
+
+from django import template
+
+
+register = template.Library()
+
+# Reference: see `validate_slug` definition in Django source.
+CLONED_CODELIST_NAME_PATTERN = re.compile(
+    r"Cloned from codelist : \[(?P<name>[-a-zA-Z0-9_]+)\]"
+)
+
+
+@register.filter
+def methodology_cloned_codelist_name_filter(methodology_text):
+    """Template filter to escape methodology markdown formatting for cloned codelist names with underscore characters."""
+    match = re.search(CLONED_CODELIST_NAME_PATTERN, methodology_text)
+    if not match:
+        return methodology_text
+
+    escaped_name = match.group("name").replace("_", r"\_")
+    return (
+        methodology_text[: match.start("name")]
+        + escaped_name
+        + methodology_text[match.end("name") :]
+    )
--- a/templates/codelists/_about_tab.html
+++ b/templates/codelists/_about_tab.html
@@ -1,3 +1,4 @@
+{% load methodology_cloned_codelist_name_filter %}
 {% load markdown_filter %}
 
 <style>
@@ -29,7 +30,7 @@
   {% if codelist.methodology %}
     <h3 class="h4">Methodology</h3>
     <div class="about__markdown">
-      {{ codelist.methodology|markdown_filter|safe }}
+      {{ codelist.methodology|methodology_cloned_codelist_name_filter|markdown_filter|safe }}
     </div>
   {% endif %}
 


opencodelists/tests/templatetags/test_markdown_filter.py
```diff
@@ -0,0 +1,48 @@
...
+        ),
+        (
+            "Cloned from codelist : [CodelistToBeCloned](/codelist/user/katieb5/codelisttobecloned/2ed738d9/)\r\nSome additional methodology text",
+            "Cloned from codelist : [CodelistToBeCloned](/codelist/user/katieb5/codelisttobecloned/2ed738d9/)\r\nSome additional methodology text",
+        ),
+        (
+            "Methodology text without cloned codelist",
+            "Methodology text without cloned codelist",
+        ),
+    ],
```
StevenMaude, line 35: suggestion: We could add some other cases, like:

* those with underscores in the remaining text itself (you correctly check this in the URLs, but we also expect the subsequent text to be unchanged too)
* `""` — we expect a zero-length string back from a zero-length string
* `"Some other text. Cloned from codelist :…"` — we only expect the `Cloned from codelist :…` we've added to match at the start of text.

opencodelists/templatetags/markdown_filter.py
```diff
@@ -7,6 +7,31 @@
 register = template.Library()
 
 
+def escape_underscores(codelist_name):
+    """Helper function to escape underscore characters in cloned codelist names."""
+    return codelist_name.replace("_", r"\_")
```
StevenMaude: suggestion: It's good that you thought about making this a separate function, which was perhaps to make it easier to test.

In this case, since it's a single use of the `str.replace()` method, we might manage without testing it at all. The tests for it are really checking that Python's `str.replace()` is working as we expect. That's a core part of Python that we shouldn't expect to change.

Because of that, I think we could remove `escape_underscores()` entirely and write this directly inline in the `cloned_codelist_name_filter()` function instead. We're only using `escape_underscores()` in the filter, so there's no other duplication elsewhere.

opencodelists/templatetags/markdown_filter.py
```diff
@@ -7,6 +7,31 @@
...
+    if not methodology_text:
+        return ""
+
+    prefix = "Cloned from codelist : ["
+    prefix_start = methodology_text.find(prefix)
+    if prefix_start == -1:
+        return methodology_text
+
+    name_start = prefix_start + len(prefix)
+    name_end = methodology_text.find("]", name_start)
```
StevenMaude: thought: There's an edge case not covered here, which is that we don't find the `]`. That's unlikely because we always expect that if we have `"Cloned from codelist…"` that we have `"["` and `"]"`, but it's also good to not have potentially odd behaviour falling through and doing things we don't expect in unusual cases.

We could check this too, as you did for `prefix_start`.

Or we could rewrite this function to use a regex which checks this for us:

```python
# Reference: `validate_slug` in Django source.
CLONED_CODELIST_NAME_PATTERN = re.compile(
    r"^Cloned from codelist : \[(?P<name>[-a-zA-Z0-9_].*?)\]"
)


@register.filter
def cloned_codelist_name_filter(methodology_text):
    match = re.search(CLONED_CODELIST_NAME_PATTERN, methodology_text)
    if not match:
        return methodology_text

    escaped_name = match.group("name").replace("_", r"\_")
    return (
        methodology_text[: match.start("name")]
        + escaped_name
        + methodology_text[match.end("name") :]
```

Regexes can be more complicated than using Python's string methods, but here it allows us to be more declarative about what we do.

opencodelists/templatetags/markdown_filter.py
StevenMaude, line 1: suggestion: Maybe the filename should be more specific than `markdown_filter`, and again mention something around cloned codelists?

opencodelists/templatetags/markdown_filter.py
```diff
@@ -7,6 +7,31 @@
 register = template.Library()
 
 
+def escape_underscores(codelist_name):
+    """Helper function to escape underscore characters in cloned codelist names."""
+    return codelist_name.replace("_", r"\_")
+
+
+@register.filter
+def cloned_codelist_name_filter(methodology_text):
```
StevenMaude: suggestion:

```suggestion
def methodology_cloned_codelist_name_filter(methodology_text):
```

Maybe? This naming is quite tricky as you've said though.

opencodelists/templatetags/markdown_filter.py
```diff
@@ -7,6 +7,31 @@
...
+    prefix = "Cloned from codelist : ["
+    prefix_start = methodology_text.find(prefix)
+    if prefix_start == -1:
+        return methodology_text
+
+    name_start = prefix_start + len(prefix)
+    name_end = methodology_text.find("]", name_start)
+
+    name = methodology_text[name_start:name_end]
+    escaped_name = escape_underscores(name)
```
StevenMaude: thought: There's an unlikely edge case not covered here, which is that:

1. we don't find the `"]"`
2. `name_end` is `-1`
3. `-1` is a valid slice index
4. `name` is the text from `name_start` up to index `-1`, the last character
5. The text in `name` gets underscores escaped in that entire selection.

That's not a big problem.

It's also *unlikely* because we always expect that if we have `"Cloned from codelist…"`, we will also have both `"["` and `"]"`. 

But it's also good to prevent unusual cases falling through unguarded.

How would we address this?

Well, we could check that `find` doesn't give us `-1` for `"]"`, just as you did for `"["`, and then handle that if it happens…

…or, alternatively, we could rewrite this function to use a regular expression. That allows us to ensure we have a complete match in one statement:

```python
# Reference: see `validate_slug` definition in Django source.
CLONED_CODELIST_NAME_PATTERN = re.compile(
    r"^Cloned from codelist : \[(?P<name>[-a-zA-Z0-9_]+)\]"
)


@register.filter
def cloned_codelist_name_filter(methodology_text):
    """Template filter to escape methodology markdown formatting for cloned codelist names with underscore characters."""
    match = re.search(CLONED_CODELIST_NAME_PATTERN, methodology_text)
    if not match:
        return methodology_text

    escaped_name = match.group("name").replace("_", r"\_")
    return (
        methodology_text[: match.start("name")]
        + escaped_name
        + methodology_text[match.end("name") :]
    )
```

(This also inlines `escape_underscores()` as mentioned above.)

Regular expressions *can* be more unwieldly to use than Python's string methods. But here the regex allows us to use a more declarative and less imperative style, more direct in its intent.

StevenMaude (approved):
Thanks for this; I'll approve this because the functionality's there, and it's an improvement on what we have, as it already stands.

I've added a couple of comments about the implementation to consider.

StevenMaude:
thought: It's unfortunate that we don't have a direct record on the codelist model that it is cloned, and we instead modify the actual methodology text.

But this is a good fix for the formatting problem 😃

opencodelists/tests/templatetags/test_markdown_filter.py
KatieB5: Thanks Steve, this is a good shout; I'll add those cases to the test :)

opencodelists/templatetags/markdown_filter.py
KatieB5: Thanks Steve! I'm definitely happy to change it. My original rationale for doing it this way was, as you say, to so I could specifically isolate and test this bit where I'd written code to modify something. But also because I thought this followed the same/similar pattern used `markdown_filter`, and I thought maybe there's a chance we'd have a use for it later, outside of the filter. 

That said, you're right: we're testing Python's `str.replace()` and we don't need to. Also YAGNI (in reference to me thinking we might need it later).

Happy to refactor as you suggest, and then we can remove the extra test too :)

opencodelists/templatetags/markdown_filter.py
KatieB5: That's a good point; I hadn't thought to change the file name. 

Although now I'm thinking about it, perhaps `def cloned_codelist_name_filter()` can live in it's own file (with a more specific name e.g. cloned_codelist_name_filter.py), and we can leave the current file name as-is.

opencodelists/templatetags/markdown_filter.py
KatieB5: I like your suggestion - it's slightly more verbose, it's more specific (which I prefer). I'll update to use this.

opencodelists/templatetags/markdown_filter.py
KatieB5: Thanks so much for this Steve! I really appreciate the super-clear explanation of the edge case, and the approach to fixing it :smiley: I originally tried to go down a regex route for this PR (I got ChatGPT to build me specifically with the regex patterns and matching cases, but the solution it suggested was complicated and difficult to read so I binned it off and went down the python route and wrote something I could actually read and understand instead).

However, what you've written is a nice alternative to the python approach, so I'm happy to go with this instead. The only slight modification I made was removing the `^` anchor to get the "Some other text. Cloned from codelist : ..." test case to pass.

StevenMaude: > The only slight modification I made was removing the ^ anchor to get the "Some other text. Cloned from codelist : ..." test case to pass.

I originally took it that the text should only be found at the start of the methodology text, and that's why I included the `^`.

But I could be wrong if cloning a clone means that you get the text included more than once?

Issue #2985, opened 2026-02-24
Enhancement request:  Improve sorting in a user's list of codelists


Mary-Anya:
# Summary
Nice to have: improve sorting for users with long lists of codelists that they've created.

#2981 sorting alphabetically case insensitive, this ticket is intended to explore further improvements that could be made to sorting users' codelists.

# What would you like to achieve?
From [slack message](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1771320626021939): It wasn't clear how codelists created by the user were sorted. It appears that the codelists were sorted alphabetically but case sensitivity played a factor in the confusion. User expected to see all the codelists together regardless. 
Being able to sort by date would be helpful too.

# Who would benefit and how?
Frequent creators of codelists - less time spent searching own codelists to figure out which they were working on / more likely to delete codelists that are not needed anymore

Issue #2981, opened 2026-02-17
Enhancement request:  Sorting in a user's list of codelists


Mary-Anya:
# Summary
Nice to have: improve sorting for users with long lists of codelists that they've created.

# What would you like to achieve?
From [slack message](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1771320626021939):  It's not clear to users how their codelists are sorted. It appears that the sorting is case sensitive. User expected to see all the codelists together regardless of the metadata completion status.
"_Earlier I was looking for a codelist whose title I knew, but couldn't find it, b/c it appeared later in the list in the "no metadata" group._

_I don't think sorting by date by default would be that useful, but having that option would be good._"

# Who would benefit and how?
Frequent creators of codelists - less time spent searching own codelists to figure out which they were working on / more likely to delete codelists that are not needed anymore

This ticket should make sure that the sorting is case insensitive, so that all codelists are sorted together alphabetically regardless of upper/lower case. 
#2985 created to investigate further approaches to improve the codelist sorting at a later date.

tomodwyer:
I am going to split this in to four PRs to fix the existing sorting, allow users to sort their own codelists, and to update the templates to use Tailwind styling. The four PRs will be:

- [x] https://github.com/opensafely-core/opencodelists/pull/3003
- [x] https://github.com/opensafely-core/opencodelists/pull/3011
- [x] https://github.com/opensafely-core/opencodelists/pull/3017
- [x] https://github.com/opensafely-core/opencodelists/pull/3029

PR #3029, opened 2026-04-15
Allow users to sort the order of their codelists

tomodwyer:
...on their own profile page.

The following changes have been made:

- `codelist-table` script renamed to `codelists-filter` as that script is now responsible for only filtering the table
- Added SVG sort icons to the `this_user` template, the visibility of which icon is shown is controlled by the `aria-sort` direction on the table header cell
- Sorting of the columns is then performed by JavaScript (transpiled from TypeScript)
  - Columns can be sorted by `none`, `asc`, or `desc` and only one column can be sorted at a time
  - Columns are sorted by date or by string
  - If a row contains a whole `tbody` (as it has more than one version of a codelist) and the row needs to be sorted by date, it will be sorted by the oldest and newest dates in the tbody

https://github.com/user-attachments/assets/eaa9db7a-43f8-41bb-b025-1f86274e021a


opencodelists/tests/functional/test_this_user_sort.py
```diff
@@ -0,0 +1,207 @@
...
+from opencodelists.models import Organisation, User
+
+
+pytestmark = [
+    pytest.mark.functional,
+    pytest.mark.django_db(databases=["default", "snomedct_test_20200101"]),
+]
+
+
+def create_draft_codelist(*, owner, author, name, updated_at):
```
mikerkelly, line 21: Question: Can we use fixtures (either existing ones or adding these as fixtures) to get enough versions for this test in the database instead of these one-offs?

assets/src/scripts/codelists-sort.ts
```diff
@@ -0,0 +1,205 @@
...
+  header.ariaSort = direction;
+
+  // Set all other header sorts to "none"
+  table.querySelectorAll("th[aria-sort]").forEach((th) => {
+    if (th !== header) {
+      th.ariaSort = "none";
+    }
+  });
+
+  // Get the column index and the type of sort from the th dataset
```
mikerkelly: typo: the th dataset?

mikerkelly (approved):
Thanks, Thomas, this looks great and the code seems clean.

I think https://github.com/opensafely-core/opencodelists/issues/2981 was just to make the sort order insensitive. I think you have gone further and implemented https://github.com/opensafely-core/opencodelists/issues/2985 which was for additional improvements. That is fine as this looks good, but you may want to update the issues?

I repeat from a previous issue that if we do too much more on this in future we should consider using a dependency to provide sorted/filtered tables.

How did you decide which columns should be sortable? Coding system and status seem potentially useful. Would it be much effort to add those too?

Icons look wrong in Chrome but work in Firefox, any idea why?

Image

tomodwyer:
> Icons look wrong in Chrome but work in Firefox, any idea why?

I can't replicate this on Firefox on macOS - to confirm did you rebuild the assets, or are you running the asset dev server?

mikerkelly:
I ran `just assets-rebuild`. Then I run `just assets run` and `just run`. http://localhost:7000/users/USERNAME/  shows the icons correctly in Firefox but not in Chrome on Linux, locally for me.

assets/src/scripts/codelists-sort.ts
tomodwyer: That was unclear, I've improved it.

PR #3044, opened 2026-04-29
Copy backups to block storage

### LINKED ISSUE #2910: Use block storage for database backups
# Why are we doing this?

Database backups use significant space on the main volume of `dokku3`. To save space on the main drive, and to improve our storage capacity, we want to move the backups storage to DigitalOcean Block Storage

When the automated database backups are stored on a Block Storage mounted volume.

# What are we doing?

- [Outcome of the spike here describes the process](https://github.com/opensafely-core/opencodelists/issues/2225#issuecomment-3570273820)
- [Slack thread here showing further Team REX discussion](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1763990829381529)

---

mikerkelly:
Part fixes #2910.

I think it makes sense that the backup generation, sanitisation, and
compression will continue to be done locally as the local file system access is
likely faster than doing everything directly in block storage.

For now, copy the compressed backups that are created locally to block storage,
and repeat the symlink and retention steps. This allows us to see that the new
location works, without changing the current process for downloading the
backups. If it seems to work in production, we can change the `cp`s to `mv`s
and remove the older commands.

This adds one more dokku environment variable, `BLOCK_BACKUP_DIR`. I will add that to the
production configuration at the time of merging this.

Tested locally with:

```
DATABASE_DIR=. BLOCK_STORAGE_DIR=tmp_block SKIP_SENTRY=true ./deploy/bin/backup.sh
```

Then checking expected files and symlinks were in place.

### DIFF
--- a/deploy/bin/backup.sh
+++ b/deploy/bin/backup.sh
@@ -36,7 +36,9 @@ on_error() { # shellcheck disable=SC2329
     RESULT=${exit_code:-1}
 
     # Clean up a stale tmp on failure
-    [ -n "${SANITISED_BACKUP_TMP:-}" ] && rm -f "$SANITISED_BACKUP_TMP" 2>/dev/null || true
+    if [ -n "${SANITISED_BACKUP_TMP:-}" ]; then
+      rm -f "$SANITISED_BACKUP_TMP" 2>/dev/null || true
+    fi
 
     # Avoid exiting inside the handler on failures of Sentry call
     set +e
--- a/deploy/bin/backup.sh
+++ b/deploy/bin/backup.sh
@@ -25,8 +25,9 @@ if [ "$SKIP_SENTRY" != "true" ]; then
     SENTRY_CRON_URL=$(sentry_cron_url "$SENTRY_DSN" "$SENTRY_MONITOR_NAME")
 fi
 
-# DATABASE_DIR is configured via dokku (see DEPLOY.md)
+# DATABASE_DIR and BLOCK_STORAGE_DIR are configured via dokku (see DEPLOY.md)
 BACKUP_DIR="$DATABASE_DIR/backup/db"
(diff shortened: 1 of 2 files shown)


deploy/bin/backup.sh
```diff
@@ -161,11 +172,33 @@ else
...
+    exit 5
+fi
+
 # Keep only the last 14 days of raw backups
 find "$BACKUP_DIR" -name "*-db.sqlite3.zst" -type f -mtime +14 -delete
+find "$BLOCK_BACKUP_DIR" -name "*-db.sqlite3.zst" -type f -mtime +14 -delete
 
 # Keep only the most recent sanitised backup.
 find "$BACKUP_DIR" -name "*sanitised*-db.sqlite3.zst" ! -wholename "$SANITISED_BACKUP_FILEPATH.zst" -type f -delete
+find "$BLOCK_BACKUP_DIR" -name "*sanitised*-db.sqlite3.zst" ! -name "$SANITISED_BACKUP_FILENAME.zst" -type f -delete
```
lucyb, line 201: In the above command we use `-wholename`, but we're using `-name` here. I think that makes sense, but I wanted to check that was deliberate?

lucyb (approved):
This looks like a sensible approach. I like that you made some improvements to make it nicer/easier to run locally too.

deploy/bin/backup.sh
mikerkelly: Thanks for checking. Yes, using `-wholename` on the full local file path would not match as we are looking in a different file tree in this check. So I used `-name` on the file name instead.

mikerkelly: I did think it might make sense to add some more variables for the several files/paths that we interact with to make this all a bit more legible, but that can wait for when we tidy up this script.

PR #3050, opened 2026-05-05
Python 3.12 via uv rather than deadsnakes ppa

rw251:
Swapping in "uv" to install python 3.12. This version of python is not available via `apt install` in the base docker image, so previously we used the deadsnakes ppa. This is a test to see if uv can do it instead to move away from this ppa dependency.

- remove deadsnakes ppa
- install uv by COPYing latest image from ghcr.io
- use uv to install python3.12

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,19 +13,19 @@ FROM ghcr.io/opensafely-core/base-docker:22.04 as base-python
 # docker clean up that deletes that cache on every apt install
 RUN rm -f /etc/apt/apt.conf.d/docker-clean
 
-# Add deadsnakes PPA for installing new Python versions
-# ensure fully working base python3 installation
-# see: https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e
-# use space efficient utility from base image
 
-RUN --mount=type=cache,target=/var/cache/apt \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -69,7 +69,8 @@ RUN --mount=type=cache,target=/var/cache/apt \
     /root/docker-apt-install.sh /tmp/build-dependencies.txt
 
 # Install everything in venv for isolation from system python libraries
-RUN python3.12 -m venv /opt/venv
+# Use uv to create and manage the venv directly - seed pip to allow python -m pip usage
```
Jongmassey, line 72: Would it not make more sense to go all in on UV so we don't need pip?

rw251: At some point in the tech incident channel thread it was mentioned that this could be a temporary change until the deadsnakes ppa is back - so this is a working version with minimal changes. The other PR about bumping to 24.04 is another alternative.

rebkwok: If you go with this option, I'd suggest updating to the latest [repo template](https://github.com/opensafely-core/repo-template/blob/main/docker/Dockerfile) for standardised uv docker things

rw251:
This was a temporary potential fix that is no longer required.

PR #3051, opened 2026-05-05
Bump base image to 24.04 - remove deadsnakes

rw251:
Test to see if we can use the 24.04 base image, and then remove the dependency on deadsnakes PPA which is currently only used because python 3.12 is not in "apt" in 22.04

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,20 +7,11 @@
 # the latest base image, by design.
 #
 # hadolint ignore=DL3007
-FROM ghcr.io/opensafely-core/base-docker:22.04 as base-python
+FROM ghcr.io/opensafely-core/base-docker:24.04 as base-python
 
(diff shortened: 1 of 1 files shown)


Jongmassey:
I've built and run a docker container locally and this seems to work just fine

Jongmassey (approved):

tomodwyer (approved):

Jongmassey:
Site seems to be up fine and running 24.04

```sh
I have no name!@a0ae47838f3d:/app$ cat /etc/*ease
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu 24.04.4 LTS"
PRETTY_NAME="Ubuntu 24.04.4 LTS"
```

PR #3052, opened 2026-05-06
Fix empty table for user with no codelists

tomodwyer:
As currently it shows an empty table when the user has no codelists, we should hide it and just leave the message "Create your first codelist above.".

### DIFF
--- a/templates/opencodelists/this_user.html
+++ b/templates/opencodelists/this_user.html
@@ -37,158 +37,160 @@
       {% endif %}
     </div>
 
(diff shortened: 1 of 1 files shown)


KatieB5 (approved):
Thanks very much Thomas! I tested it by creating a user with no codelists, running the main branch locally, then switching to your branch and running a local site to see the difference. It looked good to me, and I think it's a nice improvement, so I'm happy to approve :)

**Before**:

Image

**After**:
Image

PR #3060, opened 2026-05-13
Move backups to block storage instead of copy

### LINKED ISSUE #2910: Use block storage for database backups

mikerkelly:
Part fixes https://github.com/opensafely-core/opencodelists/issues/2910.

Database backups use significant space on the main volume of dokku3. To save space on the main drive, and to improve our storage capacity, we want to move the backups storage to DigitalOcean Block Storage.

Now that we have seen the backups successfully copy across, we can change the `cp` command to a `mv`, as the backups don't need to exist in both places. This will free up space.

I think it makes sense that the backup generation, sanitisation, and compression will continue to be done locally as the local file system access is likely faster than doing everything directly in block storage.

Cleaning up the existing backup folder can be done manually to free the space once we see that this has worked.

Tested locally with:

```
DATABASE_DIR=. BLOCK_STORAGE_DIR=tmp_block SKIP_SENTRY=true ./deploy/bin/backup.sh
```

Then checking expected files and symlinks were in place in the "block storage" location and not in the "database storage" location.

### DIFF
--- a/deploy/bin/backup.sh
+++ b/deploy/bin/backup.sh
@@ -142,39 +142,15 @@ verify_file_min_size "${SANITISED_BACKUP_FILEPATH}.zst"
 # Only now remove the uncompressed originals
 rm "$BACKUP_FILEPATH" "$SANITISED_BACKUP_FILEPATH"
 
-# Copy the compressed backups to block storage to reduce local disk usage and
+# Move the compressed backups to block storage to reduce local disk usage and
 # separate them from the production system.
-# This is WIP in https://github.com/opensafely-core/opencodelists/issues/2910
-# For now, just copy the files so they exist in both places. Once we are sure
-# this is working, this could be a move rather than a copy, and we could adjust
-# the symlinks and retention below to refer to the new location.
 mkdir -p "$BLOCK_BACKUP_DIR"
-cp --preserve=mode,timestamps "$BACKUP_FILEPATH.zst" "$BLOCK_BACKUP_DIR"
-cp --preserve=mode,timestamps "$SANITISED_BACKUP_FILEPATH.zst" "$BLOCK_BACKUP_DIR"
+rsync -av --remove-source-files "$BACKUP_FILEPATH.zst" "$BLOCK_BACKUP_DIR"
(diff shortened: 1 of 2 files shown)


deploy/bin/backup.sh
```diff
@@ -142,40 +142,16 @@ verify_file_min_size "${SANITISED_BACKUP_FILEPATH}.zst"
...
 # separate them from the production system.
-# This is WIP in https://github.com/opensafely-core/opencodelists/issues/2910
-# For now, just copy the files so they exist in both places. Once we are sure
-# this is working, this could be a move rather than a copy, and we could adjust
-# the symlinks and retention below to refer to the new location.
 mkdir -p "$BLOCK_BACKUP_DIR"
-cp --preserve=mode,timestamps "$BACKUP_FILEPATH.zst" "$BLOCK_BACKUP_DIR"
-cp --preserve=mode,timestamps "$SANITISED_BACKUP_FILEPATH.zst" "$BLOCK_BACKUP_DIR"
+mv "$BACKUP_FILEPATH.zst" "$BLOCK_BACKUP_DIR"
+mv "$SANITISED_BACKUP_FILEPATH.zst" "$BLOCK_BACKUP_DIR"
```
StevenMaude: question: Since I think this involves a different filesystem, the `mv` is presumably a full copy operation, if I'm right?

If so, is it worthwhile to verify, either using `rsync` or hashing the files before and after, that the files have correctly copied, as a guard against faulty hardware?

StevenMaude (approved):
summary: I have one question about maybe making the backup more robust. Otherwise, the changes look fine.

deploy/bin/backup.sh
mikerkelly: Good point `rsync -av --remove-source-files` ?

Issue #2456, opened 2025-04-22
Consider switching to `uv` for updating dependencies


mikerkelly:
# Why are we doing this?

`uv` can be used to manage dependencies, instead of `pip-compile` with `uv pip compile`.

`uv` generates requirements files in largely the same format. The `requirements.txt` files it outputs have slightly different comment headers, and do not place what pip-tools deems "unsafe" packages, such as `setuptools`, in a separate section.

This also allows us to try out `uv` with a view to potentially using it for more of its features in future.[^1]

[^1]: `uv` also has its own newer lock format, though that would require switching to use `uv` installed inside the Dockerfile for deployment, and developers using `uv` for installing the Python packages. This is out of scope for this issue. It's also possible to use `uv` to manage Python installation, virtualenvs and package installation, but these are also out of scope for this issue.

`uv`:

* is considerably faster than `pip-compile` to generate the requirements files
* allows for packages newer than a particular date to be excluded from consideration for update with `--exclude-newer`; this would be particularly useful to help reduce the risk of pulling in previously trusted dependencies that have been compromised 
* is a near-direct drop-in for `pip-compile`; only the steps where requirements are generated need to change

pip-tools is still being developed but has not seen a release in over a year; at time of writing, [v7.4.1 is the latest](https://github.com/jazzband/pip-tools/releases/tag/7.4.1), released 2024-03-06.

All of that context makes it reasonable to consider whether `uv` would represent a real improvement for us.

See also https://github.com/opensafely-core/job-server/issues/5049 for the Job Server equivalent.

# What are we doing?

Considering whether to switch to `uv` for updating our requirements files.

If we switch, we may need to do some or all of the following:

* update the `justfile` recipes that generate requirements files to use `uv pip compile`
* consider making `uv` a developer dependency, if necessary (this may not be deemed necessary, but we do pin `pip-compile` at the moment)
* rebuild the dependencies with `uv pip compile`
* ensure any automated dependency update workflows use `uv pip compile` too (this should be automatic if we alter the `justfile` recipes)
* communicate this to developers working on this code 
* update any affected developer documentation to reflect the change

See [these notes](https://github.com/opensafely-core/actions-registry/blob/e9163534ef08afac990726549ef644ba934ee10e/uv-spike.md#chat-with-simon) on the actions-registry spike, and [this presentation](https://drive.google.com/file/d/1md04WsNMGLqOgJZaJ9D780CZd-uSeGy0/view?usp=drive_web) about --exclude-newer.

PR #3065, opened 2026-05-14
KatieB5/switch to uv for updating dependencies

KatieB5:
Addresses part of #2456 . The rest of the work (outlined below) will be addressed in subsequent PRs.

### Work done in this PR

- [x] Update `pyproject.toml` and generate `uv.lock` file
- [x] Update `justfile` recipes
- [x] Update Docker files
- [x] Update GHA workflows to use `uv`
- [x] Remove `pip-tools` dependency

### Work to be done in subsequent PR(#3075):
- [x] Implement dependency cooldown
- [x] Remove version constraints in `pyproject.toml` for dependencies we don't explicitly need to constrain after implementing dependency cooldown
- [x] Update documentation
- [x] Communicate the change to developers working on this codebase ([Slack message in tech channel](https://bennettoxford.slack.com/archives/C63UXGB8E/p1780050503394289)).


pyproject.toml
```diff
@@ -1,3 +1,48 @@
+[project]
+name = "opencodelists"
+version = "0.1.0"
+description = "OpenCodelists"
+readme = "README.md"
+authors = [{name = "OpenSAFELY", email = "tech@opensafely.org"}]
+license = {file = "LICENSE"}
```
StevenMaude: nit:

```suggestion
license = {file = "LICENCE"}
```

pyproject.toml
```diff
@@ -1,3 +1,48 @@
...
+    "furl>=2.1.4",
+    "gunicorn>=25.3.0",
+    "lxml>=6.1.0",
+    "markdown2>=2.5.5",
+    "nh3>=0.3.5",
+    "openpyxl>=3.1.5",
+    "opentelemetry-exporter-otlp>=1.41.1",
+    "opentelemetry-instrumentation-django>=0.62b1",
+    "opentelemetry-sdk>=1.41.1",
+    # pinned to mimmum version for security patch (https://github.com/opensafely-core/opencodelists/commit/99374a6aee474e6873f41f0fb2af766b9a8da61d)
```
StevenMaude, line 34: nit: I think we can remove this comment now; we've specified minimum versions for everything here. Previously in the requirements.prod.in, only a few packages had specified minimum versions.

pyproject.toml
```diff
@@ -1,3 +1,48 @@
...
+    "requests>=2.32.4",
+    "sentry-sdk>=2.58.0",
+    "slippers>=0.7.0",
+    "sqlean-py>=3.50.4.5",
+    "structlog>=25.5.0",
+    "tqdm>=4.67.3",
+    # Serves static files. Dependabot wasn't updating past 6.3.0 for
+    # unclear reasons, and we want fix #612 from 6.8.0.
```
StevenMaude: nit: similarly, for this comment, I think we can remove it.

pyproject.toml
```diff
@@ -1,3 +1,48 @@
...
+    "whitenoise[brotli]>=6.8.0",
```
StevenMaude: nit: This doesn't make much difference, but typically, I think `uv` goes for the latest minimums when you add the dependencies in, so this could be:

```suggestion
    "whitenoise[brotli]>=6.12.0",
```

which is the version which was actually in `requirements.prod.txt`.

.github/workflows/deploy.yml
```diff
@@ -37,6 +37,8 @@ jobs:
       - uses: opensafely-core/setup-action@v1
         with:
           install-just: true
+          install-uv: true
+          cache: uv
```
StevenMaude: question: Is `uv` and the cache here required at all? Is it used, or can we remove it from some or all steps?

If all these steps run in Docker in a container, then I don't believe specifying this cache or installing `uv` is necessary.

(The caching here probably refers to some `uv`-related directory on the host. Docker does have its own caching behaviour too.)

.github/workflows/main.yml
StevenMaude, line 1: question: Same question throughout this file as elsewhere:

> Is `uv` and the cache here required at all? Is it used, or can we remove it from some or all steps?

StevenMaude:
question: In the commit message for 1753c444905d5d9a06ba28a2a5a6bfac8dcdc2a7, it says:

> A .keep file was removed when testing the updated `just docker-` recipes

Is this still relevant? There's no associated change committed here. Is it to warn people of some behaviour: in which case, perhaps it's worth clarifying in the commit message?

StevenMaude (approved):
summary: Thanks for this one @KatieB5! (I look forward to developer installations being fast: I always forget how slow regular `pip` is, compared with working in job-server 😄)

This is good to merge, I think. It's worth reviewing whether we need everything in the workflows as I mentioned.

I tested:

* building dev and production Python environments
* updating all dependencies
* updating a package
* running the Docker build
* running the server inside Docker
* running the server outside of Docker

***

Finally, I'd note that switching to `uv` gives us the possibility of simplifying the `justfile`. For example, we could use `uv run …` instead of `$BIN/python …`, and I can think we can avoid specifying `VIRTUAL_ENV`.

That said: we don't do this in job-server yet. So it's fine to defer it, and I think there's also benefit in leaving the changes and associated pull requests minimal and as incremental improvements.

StevenMaude:
thought: You may need to rebuild the requirements based on the current installed versions again, so we don't inadvertently downgrade anything here.

pyproject.toml
KatieB5: Great spot! I will update this :)

pyproject.toml
KatieB5: I'd originally kept the comments about the specifically pinned versions because that's what we'd done in job-server ([commit](https://github.com/opensafely-core/job-server/commit/1399d0fcc955092d9c46a325138b4a65cf55bb71)). I can't find the Slack thread now, but I'm pretty sure we discussed it at the time.

I also kept all the existing pins in this Part 1 PR because I hadn't yet implemented the cooldown logic, and I didn't want to change the dependency set from what was already in production when I branched off main.

In the [Part 2 PR](https://github.com/opensafely-core/opencodelists/pull/3075/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711), I then removed the unnecessary pins, keeping only the ones we'd previously agreed to retain. That now matches what we currently do in job-server, including the accompanying comments.

I'm inclined to leave the comments in for now because I think it's helpful to have the context once the other pins are removed, unless you feel strongly about removing them?

pyproject.toml
KatieB5: I agree it makes sense to pin to 6.12.0 instead. If 6.12.0 was in `requirements.prod.txt` then I guess dependabot was working as it should; I think this might also mean the comment _and_ the pin are both no longer needed.

.github/workflows/deploy.yml
KatieB5: > Is uv and the cache here required at all?

I don't believe so, for the reasons you've mentioned: these steps run inside Docker containers rather than on the host runner, and as far as I understand it we don't use `uv` on the host runner itself.

I can remove the `uv` cache statements in the opencodelists workflows, and we could potentially also remove them from [job-server](https://github.com/opensafely-core/job-server/blob/376cdd0f96bbaf097714104dc55fb8b5856ddff0/.github/workflows/main.yml#L32) too.

> I don't believe .... installing uv is necessary.

Agreed, I don't think it's necessary either, for the same reasons above. We only need uv inside the Docker containers, not on the host runner.

I'll remove the `uv` install and cache configuration from our workflows.

.github/workflows/main.yml
KatieB5: See reply [here](https://github.com/opensafely-core/opencodelists/pull/3065#discussion_r3317201805) for full response.

I think we can remove the `uv` cache and installation on the host runner

pyproject.toml
KatieB5: Agree; based on your comment below, it seems dependabot is now updating past 6.8.0 so this is no longer an issue.

KatieB5:
Thanks very much for picking up this review @StevenMaude , and for the great feedback as usual! :) 

> Is this still relevant? There's no associated change committed here.

I replaced the `.keep` file that was removed, and removed the deletion when I rebased. The reasons why the .keep was removed as part of this work aren't directly related to the uv switch for dependency management ([Slack thread discussion])(https://bennettoxford.slack.com/archives/C069SADHP1Q/p1779124881685009?thread_ts=1779124590.975649&cid=C069SADHP1Q). I'm inclined to remove my comment from the commit message (it's no longer relevant and creates unhelpful noise), but perhaps continue our Slack discussion about why this happened, and preventing it from happening in future. 

For this:

> Finally, I'd note that switching to uv gives us the possibility of simplifying the justfile. For example, we could use uv run … instead of $BIN/python …, and I can think we can avoid specifying VIRTUAL_ENV.
> 
> That said: we don't do this in job-server yet. So it's fine to defer it, and I think there's also benefit in leaving the changes and associated pull requests minimal and as incremental improvements.

My preference would be to defer both as you suggest; it keeps this PR focused on dependency management and hopefully more manageable from a review perspective (update: have created #3088 to capture these).

PR #3075, opened 2026-05-20
Katie b5/implement uv dependency cooldown

KatieB5:
Addresses part of #2456 .

Follows work done in #3065.

### Work done in this PR:
- [x] Implement dependency cooldown
- [x] Remove version constraints in `pyproject.toml` for dependencies we don't explicitly need to constrain after implementing dependency cooldown
- [x] Update documentation

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -150,6 +150,58 @@ require them or build codelists.  Refer to the team manual [OpenCodelists
 playbook](https://bennett.wiki/tech-group/playbooks/opencodelists/) for how to
 do those tasks on our infrastructure.
 
+## Dependency management
+We use [Dependabot](https://github.com/opensafely-core/opencodelists/blob/main/.github/dependabot.yml) and an [automated update workflow](https://github.com/opensafely-core/opencodelists/blob/main/.github/workflows/update-python-dependencies.yml) to keep our dependencies up to date.
+
+Dependencies are managed with `uv`.
+
+### Overview
+
+See the [uv documentation](https://docs.astral.sh/uv/concepts/projects/dependencies) for details on usage.
+
+Changes to dependencies should be made via `uv` commands, or by modifying `pyproject.toml` directly, followed by
+[locking and syncing](https://docs.astral.sh/uv/concepts/projects/sync/) via `uv` or `just` commands like
+`just devenv` or `just update-dependencies`. You should not modify `uv.lock` manually.
+
+Commands for adding, removing or modifying constraints of dependencies will automatically respect the
+global timestamp cutoff specified in the `pyproject.toml`:
+```toml
+[tool.uv]
+exclude-newer = "YYYY-MM-DDTHH:MM:SSZ"
+```
+
+Note that `uv.lock` must be reproducible from `pyproject.toml`.
+
+Otherwise, the `just check` and `just check-lockfile` recipes will fail with an error saying that the timestamps must match. You might have modified one of `uv.lock` or `pyproject.toml` but not the other:
+  - If you modified `pyproject.toml`:
+      - update `uv.lock` via `uv lock` or `just update-dependencies`
+  - If you did not modify `pyproject.toml` but have changes in `uv.lock`:
+      1. revert the changes to `uv.lock`
+      1. modify `pyproject.toml` as you require
+      1. run `uv lock` to update the `uv.lock` file.
+
+The timestamp cutoff should usually be set to midnight UTC of a past date.
+In general, the date is expected to be between 7 and 14 days ago as a result of automated weekly dependency updates.
+
+If you require a package version that is newer than the cutoff allows, you can either manually bump the global cutoff date or add a package-specific timestamp cutoff. Both options are described below.
+
+### Manually bumping the cutoff date
+The cutoff timestamp can be modified to a more recent date either manually in the `pyproject.toml` or with `just bump-uv-cutoff <days-ago>`.
+For example, to set the cutoff to today's date and upgrade all dependencies, run:
+```
+just bump-uv-cutoff 0
+just update-dependencies
+```
+
+### Adding a package-specific timestamp cutoff
+It is possible to specify a package-specific timestamp cutoff in addition to the global cutoff.
+This should be done in the `pyproject.toml` to ensure reproducible installs; see the [uv documentation](https://docs.astral.sh/uv/reference/settings/#exclude-newer-package) for details.
+If set, the package-specific cutoff will take precedence over the global cutoff regardless of which one is more recent.
+
+You should not set a package-specific cutoff that is older than the global cutoff - use a version constraint instead.
+If there is good reason to set a package-specific cutoff that is more recent than the global cutoff, **care should be taken to ensure that the package-specific cutoff is manually removed once it is over 7 days old**, as otherwise future automated updates of that package will be indefinitely blocked.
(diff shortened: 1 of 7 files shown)


justfile
```diff
@@ -94,13 +94,57 @@ upgrade-package package: && devenv
     uv lock --upgrade-package {{ package }}
 
 
-# Upgrade all dev and prod dependencies to the latest versions per the pyproject.toml
+# Move the cutoff date in pyproject.toml to N days ago (default: 7) at midnight UTC
+bump-uv-cutoff days="7":
+    #!/usr/bin/env -S uvx --with tomlkit python3.13
+    # Note we specify the python version here and we don't care if it's different to
+    # the .python-version; we need 3.11+ for the datetime code used.
```
StevenMaude, line 101: suggestion:

Would this work to use the existing Python in `.python-version`?`

```suggestion
    #!/usr/bin/env -S uvx --with tomlkit
```

Advantages are that we don't need to:

* to bump the version, it's always in line with `.python-version`
* have an extra version of Python installed other than that for the project itself (_in reality, developers probably would have this other version installed, but it's also sensible to keep a cleaner developer environment where we can_)

DEVELOPERS.md
```diff
@@ -150,6 +150,51 @@ require them or build codelists.  Refer to the team manual [OpenCodelists
...
+Commands for adding, removing or modifying constraints of dependencies will automatically respect the
+global timestamp cutoff specified in the `pyproject.toml`:
+```toml
+[tool.uv]
+exclude-newer = "YYYY-MM-DDTHH:MM:SSZ"
+```
+Changes to dependencies should be made via `uv` commands, or by modifying `pyproject.toml` directly, followed by
+[locking and syncing](https://docs.astral.sh/uv/concepts/projects/sync/) via `uv` or `just` commands like
+`just devenv` or `just update-dependencies`. You should not modify `uv.lock` manually.
+
```
StevenMaude, line 172: suggestion: A slight reordering to put the information on timestamps closer to where it continues to be discussed.

````suggestion
See the [uv documentation](https://docs.astral.sh/uv/concepts/projects/dependencies) for details on usage.

Changes to dependencies should be made via `uv` commands, or by modifying `pyproject.toml` directly, followed by
[locking and syncing](https://docs.astral.sh/uv/concepts/projects/sync/) via `uv` or `just` commands like
`just devenv` or `just update-dependencies`. You should not modify `uv.lock` manually.

Commands for adding, removing or modifying constraints of dependencies will automatically respect the
global timestamp cutoff specified in the `pyproject.toml`:
```toml
[tool.uv]
exclude-newer = "YYYY-MM-DDTHH:MM:SSZ"
```
````

File: DEVELOPERS.md [hunk 3]
```diff
@@ -150,6 +150,51 @@ require them or build codelists.  Refer to the team manual [OpenCodelists
...
+Note that `uv.lock` must be reproducible from `pyproject.toml`. Otherwise, `just check` and `just check-lockfile` will fail.
+If either of these recipes fails with an error saying that the timestamps must match, you might have modified one file but not the other:
```
StevenMaude: suggestion: A slight edit to make the [nouns more obvious](https://developers.google.com/tech-writing/one/words#recognize_ambiguous_pronouns).

```suggestion
Note that `uv.lock` must be reproducible from `pyproject.toml`.

Otherwise, the `just check` and `just check-lockfile` recipes will fail with an error saying that the timestamps must match. You might have modified one of `uv.lock` or `pyproject.toml` but not the other:
```

DEVELOPERS.md
```diff
@@ -150,6 +150,51 @@ require them or build codelists.  Refer to the team manual [OpenCodelists
...
+  - If you modified `pyproject.toml`, you must update `uv.lock` via `uv lock` / `just update-dependencies` or similar.
+  - If you did not modify `pyproject.toml` but have changes in the `uv.lock` file, you should revert the changes to `uv.lock`, modify `pyproject.toml` as you require, then run `uv lock` to update the `uv.lock` file.
```
StevenMaude: suggestion: Here's a list-based formatting of the same instructions to perhaps make it easier for the reader to see the individual instructions, and navigate where they are if following them.

(⚠️ It's difficult writing suggestions this way without a preview; please ensure the formatting is right if you apply the suggestion!)

```suggestion
  - If you modified `pyproject.toml`:
      - update `uv.lock` via `uv lock` or `just update-dependencies`
  - If you did not modify `pyproject.toml` but have changes in `uv.lock`:
      1. revert the changes to `uv.lock`
      1. modify `pyproject.toml` as you require
      1. run `uv lock` to update the `uv.lock` file.
```

StevenMaude (approved):
summary: Looks good!

To note that I haven't tested anything here and have only read through the changes. It looks OK and is comparable to how job-server works, which is why I'm happy to only read through here.

Again, we'll need to rebase on the preceding PR, and may need to rebuild the dependencies in `pyproject.toml`, `uv.lock` based on the existing dependencies we're using, just prior to deploying this.

justfile
KatieB5: If I remove python3.13 from line 99, I get errors when I try to run the recipes that invoke this line. For example:

```
just bump-uv-cutoff error: 
Failed to parse: /run/user/1000/just/just-B6RmlM/bump-uv-cutoff 
Caused by: Expected path (/run/user/1000/just/just-B6RmlM/bump-uv-cutoff) to end in a supported file extension: .whl, .tar.gz, .zip, .tar.bz2, .tar.lz, .tar.lzma, .tar.xz, .tar.zst, .tar, .tbz, .tgz, .tlz, or .txz 
/run/user/1000/just/just-B6RmlM/bump-uv-cutoff 
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I asked ChatGPT why this happened because I have very little bash/scripting experience or knowledge. Here's what it said...

The old shebang was doing two jobs:
```
#!/usr/bin/env -S uvx --with tomlkit python3.13
```

`uvx --with tomlkit` made `tomlkit` available in a temporary tool environment. `uvx` is just `uv tool run`, and it runs tools in an isolated temporary environment. `python3.13` was the actual interpreter being launched. That matters because your just recipe is being written out to a temp file with no `.py` extension, so it still needs a real Python interpreter to execute it.

When you remove `python3.13`:
```
#!/usr/bin/env -S uvx --with tomlkit
```
you are left with only the tool runner. 

For the purposes of getting this PR merged, I think I'd like to leave this as is, and investigate alternative ways to pass in a python version here as part of another PR.

KatieB5:
Thanks very much for reviewing this @StevenMaude . I'm happy to accept all the documentation changes you've suggested.

> we'll need to rebase on the preceding PR, and may need to rebuild the dependencies in pyproject.toml, uv.lock based on the existing dependencies we're using, just prior to deploying this

I'll do the rebase and sort the merge conflicts. Re: rebuilding dependencies, following our [discussion in Slack](https://bennettoxford.slack.com/archives/C080XJJ0KDJ/p1779970124835099?thread_ts=1779204249.481769&cid=C080XJJ0KDJ) on this, we've decided to go with: not updating the `exclude-newer` timestamp in this PR, and not regenerating the `uv.lock`. This should keep dependencies aligned with what is currently on `main`. Then, once the `uv` switch PRs are merged, re-run the dependency update workflow to generate a fresh `uv`-based dependency update PR.

PR #3076, opened 2026-05-20
Move SSI codelists to OxHPRU

### LINKED ISSUE #3066: Move Surgical Site Infection Codelists
# Why are we doing this?

@LFISHER7  has [requested](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1778150903356459) that the surgical site infection codelists previously uploaded (see #2954 ) to the `ukhsa` organisation are moved to the new `oxhpru` one. They may be identified in the `ukhsa` organisation as being of the OPCS4 coding system and only having `UNDER_REVIEW` CodelistVersions.

Louis has confirmed that re-uploading these codelists into the `oxhpru` organisation is acceptable if moving them is difficult. He has also confirmed that in such a situation, leaving the existing codelists in-situ is acceptable. There is no requirement for redirection from old to new codelist locations.

The OPCS4 codelists [here](https://github.com/bennettoxford/surgical-site-infections/tree/main/data/codelists/opcs4) are all published within the `oxhpru` organisation on OpenCodelists

# What are we doing?

Either
* Move and publish the latest version of the requested Codelists from the `ukhsa` organisation to the `oxhpru` one
  * A management command or data migration feel like the most robust way to do this. New Handles would be created for the codelists and the old ones flagged as inactive.
* Re-upload the codelists to the `oxhpru` organisation

---

Jongmassey:
The surgical site OPCS4 codelists previously bulk uploaded to the `ukhsa` org were requested to be moved to the `OxHPRU` org and to be published.

This data migration performs both these actions with provision for it to be reversed if necessary

Fixes #3066


codelists/migrations/0064_move_and_publish_ukhsa_ssi_codelists.py
```diff
@@ -0,0 +1,78 @@
...
+def publish_codelist_versions(codelists):
+    for codelist in codelists:
+        for version in codelist.versions.filter(status=Status.UNDER_REVIEW):
+            version.status = Status.PUBLISHED
+            version.save()
+
+
+def unpublish_codelist_versions(codelists):
+    for codelist in codelists:
+        for version in codelist.versions.filter(status=Status.UNDER_REVIEW):
```
rw251: Should that be filtering to PUBLISHED codelists?

codelists/migrations/0064_move_and_publish_ukhsa_ssi_codelists.py
```diff
@@ -0,0 +1,78 @@
...
+    if reverse:
+        from_org = oxhpru
+        to_org = ukhsa
+    else:
+        from_org = ukhsa
+        to_org = oxhpru
+    codelists = [
+        codelist
+        for codelist in from_org.codelists.filter(coding_system_id="opcs4")
+        if not reverse
```
rw251: Doesn't this mean when reverse==True that codelists is empty?

rw251:
A couple of comments inline. Also, just to confirm, the set of UNDER_REVIEW opcs4 codelists under ukhsa is exactly the set of codelists to move?

codelists/migrations/0064_move_and_publish_ukhsa_ssi_codelists.py
Jongmassey: Ah yes, correct

codelists/migrations/0064_move_and_publish_ukhsa_ssi_codelists.py
Jongmassey: ah yes, need to switch the and for an or there

rw251 (approved):

PR #3077, opened 2026-05-20
Jongmassey/icd10 multi edition model

### LINKED ISSUE #3072: Design the new combined ICD10 coding system model
**Why are we doing this?**

We are creating a new release of ICD10 that is a combination of the nhs-modified 2016 edition (apcs data), and the international 2019 with COVID additions (ons data). This requires a change to the underlying coding system model for ICD10.

**How will we know when it's done?**

We will have a plan for a db model.

**What are we doing?**

- Consider how we will store different codes, descriptions etc from the different underlying editions
- Adding the dagger/asterisk info to the model
- Adding additional rubric info to the model
- This is _not_ the ticket for implementing the model, that will be a separate ticket

Jongmassey:
Fixes #3072


coding_systems/icd10/models.py
```diff
@@ -1,10 +1,110 @@
...
+    kind = get_char_choices_field(RubricKind)
+    text = models.TextField()
+    concept = models.ForeignKey(
+        Concept, on_delete=models.CASCADE, related_name="rubrics"
+    )
+
+
+class ModifierRubric(models.Model):
+    kind = get_char_choices_field(RubricKind)
+    text = models.TextField
```
rw251: Missing a closing `()`. Interesting that django must just ignore this, or at least not complain too noisily

coding_systems/icd10/models.py
```diff
@@ -1,10 +1,110 @@
...
     )
+    editions = models.ManyToManyField(Edition, through="ConceptEdition")
+
+
+class ConceptRubric(models.Model):
+    kind = get_char_choices_field(RubricKind)
+    text = models.TextField()
+    concept = models.ForeignKey(
+        Concept, on_delete=models.CASCADE, related_name="rubrics"
+    )
```
rw251: This might be a failing of my understanding of django models.. but does this relation here (and the `ModifierRubric` equivalent) allow you to link a rubric directly to a `Concept` without going via an `Edition`? `ConceptEdition` has the "edition-aware" `rubrics` field - so is this FK relation here also needed?

coding_systems/icd10/models.py
```diff
@@ -1,10 +1,110 @@
...
+    edition = models.ForeignKey(
+        Edition, on_delete=models.CASCADE, related_name="concept_editions"
+    )
+    kind = get_char_choices_field(ConceptKind)
+    usage = get_char_choices_field(ConceptUsage)
+    term = models.CharField(max_length=255, blank=True)
+    term_modifier = models.CharField(max_length=255, blank=True, null=True)
+    modifier_position = models.IntegerField(blank=True, null=True)
+    rubrics = models.ForeignKey(
+        ConceptRubric, on_delete=models.CASCADE, related_name="concepts"
```
rw251: I think this is a FK to a single ConceptRubric rather than many? If we're trying to associate multiple Rubrics with a ConceptEdition, then should we just change the FK on ConceptRubric (and ModifierRubric) to point to `ConceptEdition` rather than just `Concept`, and remove `rubrics` from here?

coding_systems/icd10/models.py
Jongmassey: You're right. Now I've thought about this some more it needs to go to `ConceptEdition`

Jongmassey:
NB it is necessary to run the migration associated with the transitional model commit whilst checked out to that commit, and the migration that migrates the existing data at its commit before finally running the final model migration.

This will probably need split into multiple PRs come deploy time to make all that work.

Jongmassey:
`coding_systems/icd10/tests/test_import_data.py::TestImportData::test_import_data` is expected to continue to fail until the work in the `eli/add-icd10-data` branch is completed and merged into this branch

Jongmassey:
This has now been updated such that the enum of Rubric kinds is complete wrt the kinds found in the WHO-provided 2016 and 2019 ClaML, and those scraped from the NHS ICD-10 Browser website.

Jongmassey:
Agreed in @opensafely-core/team-rsi  this morning to merge with failing test as we feel high certainty that model is complete

PR #3100, opened 2026-06-03
Update medication banner behaviour

### LINKED ISSUE #3059: Enhancement request:  Adding the warning banner for codelists 'Under Review'
# Summary
User has asked "The warning about medication codelists becoming outdated only appears on the published version. Given it's common to use "under review" codelists in projects, should the warning appear on all medication codelists?"

# What would you like to achieve?
153 "under review" codelists have been executed on the backend. Users have previously [shared](https://bennettoxford.slack.com/archives/C07AA91BZD4/p1769168890196019) that "A codelist is created and used as a placeholder, either with the intention to come back to it or while waiting to have it reviewed by someone else. But it gets forgotten once you start running your analyses and getting results"

It would be helpful to add the warning below so that users are reminded that the codelist may not be suitable for reuse. However, it would be confusing to users if they click 'submit for review' and immediate see the banner below.

Image

[Discussion of options so far](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1778151811270739) - it may need something that's dynamic depending on whether the codelist is out of date or not?

More broadly, it maybe worth reconsidering the expected user journey here, and the value of the clinical review to users.

# Who would benefit and how?
Users reusing codelist, either 'under review' or 'published' are given the same warning so they can decide whether any further updates are needed to the codelist before they execute it.

KatieB5:
Fixes #3059. Note: User research findings mentioned in the issue will be shared separately in the TeamREX Slack channel.

Following user feedback and [team discussion](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1779357784043099), the medication warning banner banner is now shown when a codelist version was built using an outdated dm+d or BNF coding system release, rather than being tied to publication status.

A coding system release is considered outdated when a newer release is available in OpenCodelists for the same coding system.

### DIFF
--- a/codelists/views/version.py
+++ b/codelists/views/version.py
@@ -2,10 +2,15 @@ from django.contrib import messages
 from django.shortcuts import render
 from django.utils.html import format_html
 
+from coding_systems.versioning.models import CodingSystemRelease
+
 from ..models import Status
 from ..presenters import present_search_results
 from ..tree_data import build_tree_data
 from .decorators import load_version
+
+
+MEDICATION_WARNING_BANNER_CODING_SYSTEMS = ["dmd", "bnf"]
 
 
 @load_version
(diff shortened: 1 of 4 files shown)


codelists/views/version.py
```diff
@@ -66,6 +71,16 @@ def version(request, clv):
     except AttributeError:
         latest_published_version_url = None
 
+    # Check whether this codelist was built from an outdated coding system release,
+    # so we know whether to show the medication warning banner.
+    if clv.coding_system_id in MEDICATION_WARNING_BANNER_CODING_SYSTEMS:
+        coding_system_release_outdated = CodingSystemRelease.objects.filter(
```
tomodwyer: In `coding_systems/base/coding_system_base.py` we use `CodingSystemRelease.objects.ready()` before filtering, to only filter on codelists that are ready to use - should we do that here?

nishtha-kalra: I think you are right. We should filter on codelists that are ready to use.

nishtha-kalra (approved):

PR #3102, opened 2026-06-04
Fix bnfdmd mapping release import

KatieB5:
New [BNF/dm+d (SNOMED) mapping data](https://www.nhsbsa.nhs.uk/prescription-data/understanding-our-data/bnf-snomed-mapping) is released by the NHS Business Services Authority (NHSBSA) every month. Every Wednesday we run a cron job that downloads and imports the latest mapping release.

The mapping data is downloaded as a ZIP file with an `.xlsx` workbook inside it. During the import process, `import_release()` extracts the ZIP, and then runs `import_data()` passing in the full path to the workbook file that the mapping data should be imported from.

Previously, `import_release()` assumed that the workbook had the same filename as the ZIP file (with `.xlsx` instead of `.zip`). However, for the May 2026 release, the workbook had a different filename from the ZIP file, meaning the path passed to `import_data()` was incorrect. This caused a `FileNotFoundError` and triggered a [Sentry alert](https://ebm-datalab.sentry.io/issues/7510271220/?alert_rule_id=1877862&alert_type=issue&notification_uuid=0dc684c5-96e8-40d5-9065-9ed963e578de&project=5248013&referrer=slack&statsPeriod=90d). 

### Work done:
Updated the `import_release()` to get the path to the workbook from the extracted ZIP contents, rather than constructing a path from the ZIP filename. `import_release()` now also checks that we find exactly one workbook (expected) before calling `import_data()`, and raises an error if not. Thank you for the [suggestions](https://bennettoxford.slack.com/archives/C069SADHP1Q/p1780578444868179?thread_ts=1780562790.214199&cid=C069SADHP1Q) @nishtha-kalra  :) 

### Testing:
Added tests covering different workbook filename formats, archive layouts, and workbook counts (zero, and multiple `.xlsx` files).

### DIFF
--- a/mappings/bnfdmd/import_data.py
+++ b/mappings/bnfdmd/import_data.py
@@ -1,4 +1,4 @@
-import os
+from pathlib import Path
 from tempfile import TemporaryDirectory
 from zipfile import ZipFile
 
@@ -50,6 +50,9 @@ def import_release(release_zipfile, release_name, valid_from):
         release_zip = ZipFile(str(release_zipfile))
         logger.info("Extracting", release_zip=release_zip.filename)
         release_zip.extractall(path=tempdir)
-        import_data(
-            os.path.join(tempdir, release_zipfile.name.replace(".zip", ".xlsx"))
-        )
+        xlsx_files = list(Path(tempdir).rglob("*.xlsx"))
+        if len(xlsx_files) != 1:
+            raise ValueError(
+                f"Expected exactly one .xlsx file in {release_zipfile}, found {xlsx_files}"
+            )
+        import_data(str(xlsx_files[0]))
--- a/mappings/bnfdmd/tests/test_import_data.py
+++ b/mappings/bnfdmd/tests/test_import_data.py
@@ -1,3 +1,8 @@
+from unittest.mock import patch
+from zipfile import ZipFile
+
+import pytest
+
 from mappings.bnfdmd.import_data import import_release
 from mappings.bnfdmd.models import Mapping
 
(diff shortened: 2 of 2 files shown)


mappings/bnfdmd/import_data.py
```diff
@@ -50,6 +50,5 @@ def import_release(release_zipfile, release_name, valid_from):
         release_zip = ZipFile(str(release_zipfile))
         logger.info("Extracting", release_zip=release_zip.filename)
         release_zip.extractall(path=tempdir)
-        import_data(
-            os.path.join(tempdir, release_zipfile.name.replace(".zip", ".xlsx"))
-        )
+        xlsx_files = list(Path(tempdir).rglob("*.xlsx"))
+        import_data(str(xlsx_files[0]))
```
nishtha-kalra, line 58: We could also add a condition here to check whether the file exists in the list before accessing its first element. Like:
```
if len(xlsx_files) > 0:
    import_data(str(xlsx_files[0]))
```

KatieB5: That's a good idea. We could potentially have a scenario where there isn't the 1 file we expect in the downloaded zip from NHSBSA. I'll update the code :)

Jongmassey: I'm failing to see how `len(xlsx_files)` could be == 0 at this line. Surely the check on line 54 would have raised a `ValueError` if that were the case?

Jongmassey (approved):
Looks like a good fix!

PR #3108, opened 2026-06-10
Eli/extend parser with rubrics

eli-miriam:
Extend the ClaML parser to handle rubrics, and to handle codes like `A52.7` where it has a dagger usage denoted in `Fragment` sections of its code but not in its top-level element.

Currently the data importer does not do anything with the newly added rubrics.


coding_systems/icd10/claml_parser.py
```diff
@@ -74,6 +74,25 @@
...
+    "footnote",
+    "text",
+    "coding_hint",
+    "definition",
+    "introduction",
+    "modifierlink",
+    "note",
+    "exclusion",
+    "inclusion",
+    "preferredlong",
```
rw251, line 88: Think the `l` is capitalised. But actually, maybe easiest to just remove `preferred` and `preferredLong` as the code skips if it one of these so will never match this set.

rw251: Or, better yet, when we rebase we can just use the `RubricKind` class from the model

Jongmassey:
Absorbed into #3106

PR #3109, opened 2026-06-11
Check for and fix incorrect django settings imports

Jongmassey:
Django project settings should be imported `from django.config` per [the django docs](https://docs.djangoproject.com/en/6.0/topics/settings/#using-settings-in-python-code), not from the project root.

At @iaindillingham 's suggestion, this PR introduces a `ruff` rule to check for this, and a commit that fixes the one instance that was found to contravene this rule.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,7 @@ extend-select = [
   "ISC",  # flake8-implicit-str-concat
   "UP",  # pyupgrade
   "W",  # pycodestyle warning
+  "TID251", # banned imports
 ]
 extend-ignore = [
   "E501",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -123,6 +124,9 @@ lines-after-imports = 2
 "codelists/views/__init__.py" = ["F401"]
 "opencodelists/views/__init__.py" = ["F401"]
 
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
+"opencodelists.settings".msg = "Import settings through django.config instead."
```
mikerkelly: Typo: This should read `django.conf`. Maybe "Use 'from django.conf import settings' instead", as a hint that you have to import the whole module, not individual settings.

mikerkelly (approved):

pyproject.toml
Jongmassey: good spot, thanks. I like your suggested hint text

PR #3129, opened 2026-06-25
Jongmassey/port icd10 scrape

Jongmassey:
Move icd browser scraper and claml converter code from bennettoxford/icd-browser-scraper into opencodelists for production scraping purposes.

The source project uses `lxml` for XML wrangling, but closer inspection shows the base python `xml` package will suffice and so this code is changed to use the latter as not to add another dependency.

Both the scraping and conversion scripts were written with the expectation of the browser scraper project structure and to be called as standalone scripts from the CLI. They are changed here such that they have main entry functions that can be easily called from other scripts as part of the ICD-10 import process.


Jongmassey:
Integrated into `icd-10-multi-editions` via #3139  and #3130

PR #3130, opened 2026-06-26
New combined ICD-10 edition parser and importer

rw251:
- Adds a new claml parser to pull out: codes, modifier codes, dagger/asterisk pairs, hierarchy and other rubrics
- Includes a new claml downloader to grab the 2016 and 2019 base WHO ICD-10 claml files
- Maintains a list of known differences (and what to do about them during the merge) between the 2016 base claml and the 2016 NHS-modified (scraped) claml
- Updates the import script to load the new ICD-10 multi-edition release
- changes to `coding_system.py` that
  - improve the speed of _code_to_chapter
  - fix to return the term modifier appended to the term for modifier codes
  - look for terms from all editions and take the latest - rather than find the latest edition and find the term (doesn't work if the term only exists in an earlier edition)


Jongmassey:
Going to do the git surgery here and final E2E testing before merging to `icd-10-multi-editions`

Issue #3058, opened 2026-05-12
Enhance sorting organisation codelists


mikerkelly:
# Why are we doing this?

The organisation codelist page (https://www.opencodelists.org/codelist/opensafely/, that uses a path in the index page) and the organisation under review codelist page (https://www.opencodelists.org/codelist/opensafely/under-review/, that uses `codelists/under_review_index.html`) both present a very basic list of codelist versions with no metadata and provide search only by name, and no sorting options. This makes them harder to use and find relevant codelists in different workflows compared to presenting in a table with metadata and better sorting options.

We recently improved several other pages that list codelists in https://github.com/opensafely-core/opencodelists/issues/2981, maybe we can do similar changes here.

These pages do get use by logged in users, in the last 60 days ([Honeycomb](https://ui.honeycomb.io/bennett-institute-for-applied-data-science/environments/production/datasets/opencodelists/result/vSedmZ8E7E2)) `
codelist/<slug:organisation_slug>/` was hit 334 times (5-6 times a day) by such users, and `
codelist/<slug:organisation_slug>/under-review/` 90 times (1-2 times a day). Though somewhat less than `users/<slug:username>/` that was improved in https://github.com/opensafely-core/opencodelists/issues/2981, which got 1704 hits (~30/day).

<details>
<summary>Screenshots - click to expand</summary>

### Before

<img width="1197" height="521" alt="Image" src="https://github.com/user-attachments/assets/5c4f55ea-f7d5-42ed-8a20-b21d70e16a1e" />

<img width="1197" height="521" alt="Image" src="https://github.com/user-attachments/assets/6f28c914-42e3-436a-a748-a4e41210589e" />

### After

Like on a different page in https://github.com/opensafely-core/opencodelists/pull/3011

<img width="2493" height="1787" alt="Image" src="https://github.com/user-attachments/assets/0bad250f-f2a5-4e6a-a44f-1b9fc57a1573" />

</details>

# How will we know when it's done?

...

# What are we doing?

...

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

PR #3134, opened 2026-06-30
Enable sorting of organisation codelists

tomodwyer:
Part of #3058 - this PR adds the functionality to sort codelists and stores the state in query params.

Published codelist tests cover:

- `sort=name&direction=asc`
- `sort=name&direction=desc`
- `sort=updated_at&direction=asc`
- `sort=updated_at&direction=desc`

Under-review tests cover:

- `sort=name&direction=asc`
- `sort=name&direction=desc`
- `sort=created_at&direction=asc`
- `sort=created_at&direction=desc`

# Screenshots

### DIFF
--- a/codelists/tests/views/test_organisations.py
+++ b/codelists/tests/views/test_organisations.py
@@ -1,6 +1,17 @@
+from datetime import datetime
+
+import pytest
+from django.utils import timezone
+
 from codelists.actions import publish_version
-from codelists.models import Handle, Status
+from codelists.models import Codelist, CodelistVersion, Handle, Status
 from opencodelists.list_utils import flatten
+
+
+def aware_datetime(year: int, month: int, day: int) -> datetime:
+    """Return a timezone-aware datetime for deterministic timestamp assertions."""
+
+    return timezone.make_aware(datetime(year, month, day))
 
 
 def test_published_index_only_returns_public_codelists_with_published_versions(
@@ -51,14 +62,76 @@ def test_paginate_published_codelists(client, organisation, create_codelists):
     assert len(page_obj.object_list) == min(50, len(published_for_organisation) - 50)
 
 
-def test_published_pagination_preserves_search_query(
-    client, organisation, create_codelists
+@pytest.mark.parametrize(
+    ("path", "status"),
+    [
+        ("", Status.PUBLISHED),
+        ("under-review/", Status.UNDER_REVIEW),
+    ],
+)
+def test_organisation_pagination_preserves_search_and_sort_query(
+    client, organisation, create_codelists, path, status
 ):
-    create_codelists(70, owner=organisation, status=Status.PUBLISHED)
+    create_codelists(70, owner=organisation, status=status)
 
-    rsp = client.get(f"/codelist/{organisation.slug}/", {"q": "Codelist"})
+    rsp = client.get(
+        f"/codelist/{organisation.slug}/{path}",
+        {"q": "Codelist", "sort": "name", "direction": "desc"},
+    )
 
-    assert 'href="?page=2&q=Codelist"' in rsp.content.decode()
+    assert rsp.context["pagination"]["next_url"] == (
+        "?page=2&q=Codelist&sort=name&direction=desc"
+    )
+    assert 'href="?page=2&amp;q=Codelist&amp;sort=name&amp;direction=desc"' in (
+        rsp.content.decode()
+    )
+
+
+@pytest.mark.parametrize(
+    ("sort", "direction", "expected_names"),
+    [
+        (
+            "name",
+            "asc",
+            ["SNOMED alpha", "SNOMED Bravo", "SNOMED Charlie"],
+        ),
+        (
+            "name",
+            "desc",
+            ["SNOMED Charlie", "SNOMED Bravo", "SNOMED alpha"],
+        ),
+        (
+            "updated_at",
+            "asc",
+            ["SNOMED Charlie", "SNOMED alpha", "SNOMED Bravo"],
+        ),
+        (
+            "updated_at",
+            "desc",
+            ["SNOMED Bravo", "SNOMED alpha", "SNOMED Charlie"],
+        ),
+    ],
+)
+def test_published_codelists_search_and_sort(
+    client, organisation, create_codelist, sort, direction, expected_names
+):
+    bravo = create_codelist("SNOMED Bravo")
+    alpha = create_codelist("SNOMED alpha")
+    charlie = create_codelist("SNOMED Charlie")
+    create_codelist("Ignored published")
+
+    Codelist.objects.filter(pk=bravo.pk).update(updated_at=aware_datetime(2020, 3, 1))
+    Codelist.objects.filter(pk=alpha.pk).update(updated_at=aware_datetime(2020, 2, 1))
+    Codelist.objects.filter(pk=charlie.pk).update(updated_at=aware_datetime(2020, 1, 1))
+
+    rsp = client.get(
+        f"/codelist/{organisation.slug}/",
+        {"q": "SNOMED", "sort": sort, "direction": direction},
+    )
+
+    assert [
+        codelist.name for codelist in rsp.context["page_obj"].object_list
+    ] == expected_names
 
 
 def test_published_index_does_not_duplicate_codelists_with_multiple_published_versions(
@@ -148,6 +221,59 @@ def test_paginate_under_review_versions(client, organisation, create_codelists):
     rsp = client.get(f"/codelist/{organisation.slug}/under-review/?page=2")
     page_obj = rsp.context["page_obj"]
     assert len(page_obj.object_list) == min(50, len(under_review_for_organisation) - 50)
+
+
+@pytest.mark.parametrize(
+    ("sort", "direction", "expected_names"),
+    [
+        (
+            "name",
+            "asc",
+            ["DM+D alpha", "DM+D Bravo", "DM+D Charlie"],
+        ),
+        (
+            "name",
+            "desc",
+            ["DM+D Charlie", "DM+D Bravo", "DM+D alpha"],
+        ),
+        (
+            "created_at",
+            "asc",
+            ["DM+D Charlie", "DM+D alpha", "DM+D Bravo"],
+        ),
+        (
+            "created_at",
+            "desc",
+            ["DM+D Bravo", "DM+D alpha", "DM+D Charlie"],
+        ),
+    ],
+)
+def test_under_review_versions_search_and_sort(
+    client, organisation, create_codelist, sort, direction, expected_names
+):
+    bravo = create_codelist("DM+D Bravo", status=Status.UNDER_REVIEW)
+    alpha = create_codelist("DM+D alpha", status=Status.UNDER_REVIEW)
+    charlie = create_codelist("DM+D Charlie", status=Status.UNDER_REVIEW)
+    create_codelist("Ignored review", status=Status.UNDER_REVIEW)
+
+    CodelistVersion.objects.filter(codelist=bravo).update(
+        created_at=aware_datetime(2020, 3, 1)
+    )
+    CodelistVersion.objects.filter(codelist=alpha).update(
+        created_at=aware_datetime(2020, 2, 1)
+    )
+    CodelistVersion.objects.filter(codelist=charlie).update(
+        created_at=aware_datetime(2020, 1, 1)
+    )
+
+    rsp = client.get(
+        f"/codelist/{organisation.slug}/under-review/",
+        {"q": "DM+D", "sort": sort, "direction": direction},
+    )
+
+    assert (
+        [version.codelist.name for version in rsp.context["page_obj"].object_list]
+    ) == expected_names
 
 
 def test_search_only_returns_codelists_with_under_review_versions(
--- a/codelists/views/organisations.py
+++ b/codelists/views/organisations.py
@@ -32,46 +32,48 @@ def organisation_published(
     request: HttpRequest, organisation_slug: str
 ) -> HttpResponse:
     organisation, handles, q = _get_organisation_handles(request, organisation_slug)
-    sort, direction = _get_sort(request, Status.PUBLISHED)
+    sort_by, sort_direction = _get_sort(request, Status.PUBLISHED)
     published_handles = _handles_with_status(
         handles, Status.PUBLISHED
     ).prefetch_related(_current_handle_prefetch("codelist__handles"))
     codelists = _sort_published_codelists(
-        [handle.codelist for handle in published_handles], sort, direction
+        [handle.codelist for handle in published_handles], sort_by, sort_direction
     )
 
     page_obj = Paginator(codelists, PAGE_SIZE).get_page(request.GET.get("page"))
 
     ctx = {
         "page_obj": page_obj,
-        "pagination": _pagination_context(page_obj, q, (sort, direction)),
+        "pagination": _pagination_context(page_obj, q, sort_by, sort_direction),
         "organisation": organisation,
         "q": q,
-        "sort": sort,
-        "direction": direction,
-        "sort_options": _sort_options_context(q, sort, direction, SORT_FOR_PUBLISHED),
+        "sort_by": sort_by,
+        "sort_direction": sort_direction,
+        "sort_options": _sort_options_context(
+            q, sort_by, sort_direction, SORT_FOR_PUBLISHED
+        ),
     }
     return render(request, "codelists/organisation_published.html", ctx)
 
 
 def organisation_review(request: HttpRequest, organisation_slug: str) -> HttpResponse:
     organisation, handles, q = _get_organisation_handles(request, organisation_slug)
-    sort, direction = _get_sort(request, Status.UNDER_REVIEW)
+    sort_by, sort_direction = _get_sort(request, Status.UNDER_REVIEW)
     versions = _sort_under_review_versions(
-        _under_review_versions(handles), sort, direction
+        _under_review_versions(handles), sort_by, sort_direction
     )
 
     page_obj = Paginator(versions, PAGE_SIZE).get_page(request.GET.get("page"))
 
     ctx = {
         "page_obj": page_obj,
-        "pagination": _pagination_context(page_obj, q, (sort, direction)),
+        "pagination": _pagination_context(page_obj, q, sort_by, sort_direction),
         "organisation": organisation,
         "q": q,
-        "sort": sort,
-        "direction": direction,
+        "sort_by": sort_by,
+        "sort_direction": sort_direction,
         "sort_options": _sort_options_context(
-            q, sort, direction, SORT_FOR_UNDER_REVIEW
+            q, sort_by, sort_direction, SORT_FOR_UNDER_REVIEW
         ),
     }
     return render(request, "codelists/organisation_review.html", ctx)
@@ -132,10 +134,12 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
     )
 
 
-def _get_sort(request: HttpRequest, status: Status) -> tuple[str, str]:
-    direction = request.GET.get("direction", SORT_DIRECTION_ASC)
-    if direction not in SORT_DIRECTIONS:
-        direction = SORT_DIRECTION_ASC
+def _get_sort(request, status):
+    query_sort_direction = request.GET.get("direction")
+    if query_sort_direction in SORT_DIRECTIONS:
+        sort_direction = query_sort_direction
+    else:
+        sort_direction = SORT_DIRECTION_ASC
 
     sort_options_by_status = {
         Status.PUBLISHED: SORT_FOR_PUBLISHED,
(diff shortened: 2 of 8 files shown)


codelists/views/organisations.py
```diff
@@ -15,32 +17,64 @@
...
+SORT_DIRECTION_ASC = "asc"
+SORT_DIRECTION_DESC = "desc"
+SORT_DIRECTIONS = {SORT_DIRECTION_ASC, SORT_DIRECTION_DESC}
+
+SORT_BY_CREATED_AT = "created_at"
+SORT_BY_NAME = "name"
+SORT_BY_UPDATED_AT = "updated_at"
+SORT_FOR_PUBLISHED = {SORT_BY_NAME, SORT_BY_UPDATED_AT}
+SORT_FOR_UNDER_REVIEW = {SORT_BY_NAME, SORT_BY_CREATED_AT}
+
```
mikerkelly, line 29: Suggestion, optional, terminology: I think `field` and `order` are more idiomatic and informative terms to qualify a sort than `sort` and `direction`. Those are the words used in Django QuerySet, so it probably makes sense to mirror that in our code.

https://docs.djangoproject.com/en/6.0/ref/models/querysets/#order-by

mikerkelly: Suggestion, optional: Python has `Enum` to represent a set of names bound to values. Since Python 3.11, we have `StrEnum`, which is slightly nicer when just working with string values, like we do here.

https://docs.python.org/3/library/enum.html#enum.StrEnum

mikerkelly: Suggestion, optional: We know that later we will want to select valid sort fields by status. So instead of defining `SORT_FOR_PUBLISHED` and `SORT_FOR_UNDER_REVIEW`, you could define a `dict` keyed on `Status`, mapping to the allowed values. Then the other code can change behaviour by keying into this dictionary based on status rather than having to define explicitly which to use. This can make code drastically more expressive and concise, which makes it easier to follow and validate.

mikerkelly: Putting those together could could look something like:

```python
from enum import StrEnum, auto


class SortOrder(StrEnum):
    ASC = auto()  # automatically "asc" ...
    DESC = auto()


class SortField(StrEnum):
    CREATED_AT = auto()
    NAME = auto()
    UPDATED_AT = auto()


ALLOWED_SORT_FIELDS = {
    Status.PUBLISHED: {SortField.NAME, SortField.UPDATED_AT},
    Status.UNDER_REVIEW: {SortField.NAME, SortField.CREATED_AT},
}
```

For example, that could simplify this:

```python 
sort_options_by_status = {
    Status.PUBLISHED: SORT_FOR_PUBLISHED,
    Status.UNDER_REVIEW: SORT_FOR_UNDER_REVIEW,
}
sort_options = sort_options_by_status.get(status, {SORT_BY_NAME})

query_sort_by = request.GET.get("sort")
if query_sort_by in sort_options:
    sort_by = query_sort_by
else:
    sort_by = SORT_BY_NAME
```

to this (combining with ternary operator):

```python
query_sort = request.GET.get("sort")
sort_by = query_sort if query_sort in ALLOWED_SORT_FIELDS[status] else SortField.NAME
```

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+        Status.PUBLISHED: SORT_FOR_PUBLISHED,
+        Status.UNDER_REVIEW: SORT_FOR_UNDER_REVIEW,
+    }
+    sort_options = sort_options_by_status.get(status, {SORT_BY_NAME})
+
+    query_sort_by = request.GET.get("sort")
+    if query_sort_by in sort_options:
+        sort_by = query_sort_by
+    else:
+        sort_by = SORT_BY_NAME
```
mikerkelly, line 154: Earlier I suggest that this can be simplified like:

```python
query_sort = request.GET.get("sort")
sort_by = query_sort if query_sort in ALLOWED_SORT_FIELDS[status] else SortField.NAME
```

codelists/views/organisations.py
```diff
@@ -15,32 +17,64 @@
...
+SORT_BY_NAME = "name"
+SORT_BY_UPDATED_AT = "updated_at"
+SORT_FOR_PUBLISHED = {SORT_BY_NAME, SORT_BY_UPDATED_AT}
+SORT_FOR_UNDER_REVIEW = {SORT_BY_NAME, SORT_BY_CREATED_AT}
+
 
 def organisation_published(
     request: HttpRequest, organisation_slug: str
 ) -> HttpResponse:
     organisation, handles, q = _get_organisation_handles(request, organisation_slug)
```
mikerkelly, line 34: Separation of concerns: I think that we should do the extraction of all of the query parameters first, in the view or in a function dedicated to that. Currently the view is passing the request to many of the utility functions, and it is not obvious why until you hunt down which specific parameters they later pluck out of the object.

It seems surprising that a function called `_get_organisation_handles` is returning something called `q` and also the `organisation`, as well as the handles. And that not all organisation handles are returned. Maybe these should be reorganised?

mikerkelly: Naming: Please let's not use variable names that are just initial letters of what we really mean. Otherwise the reader has to work that out for themselves. Obvious things are not obvious, and there are often more than one relevant concepts starting with a letter. So `q` should be `query`, probably, or maybe `search_term` would be even better. Here and throughout. Possibly also in the URL query parameters.

I know that this is how the index code does it. Probably it should change there, too. Actually, that module provides a great example of why this is problematic, as `search_handles` includes both a parameter called `q` and many references to Django `Q` objects, which are a totally different thing (and also have a bad name).

mikerkelly: So something like:

```python
def organisation_published(request: HttpRequest, organisation_slug: str) -> HttpResponse:
    status = Status.PUBLISHED
    search_term, page, sort_order, sort_field = _parse_query_params(request)
    organisation = get_object_or_404(Organisation, slug=organisation_slug)
    handles = (
        Handle.objects.filter(organisation=organisation)
        .filter(is_current=True, codelist__is_private=False)
        .filter(
            Exists(
                CodelistVersion.objects.filter(
                    codelist=OuterRef("codelist"), status=Status.PUBLISHED  # change this in the other view 
                )
            )
        )
        .select_related("codelist")
        .order_by(Lower("name"))  # can branch or have a look to add the right parameter here
    )

    # compose with `search_handles` if search term set.    

    paginator =  Paginator(handles, PAGE_SIZE).get_page(request.GET.get("page"))
```

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+            else None
+        ),
+        "next_url": url_for(page_obj.next_page_number())
+        if page_obj.has_next()
+        else None,
+        "first_url": url_for(1),
+        "last_url": url_for(paginator.num_pages),
+        "page_links": [
+            {"number": page_number, "url": url_for(page_number)}
+            for page_number in paginator.page_range
```
mikerkelly, line 210: I don't ever see first, last, or page number links no matter how many search results I have or which page I am on. Could you please check that this is working for you and tell me what parameters you were using to see that?

codelists/views/organisations.py
```diff
@@ -15,32 +17,64 @@
...
+        "page_obj": page_obj,
+        "pagination": _pagination_context(page_obj, q, sort_by, sort_direction),
         "organisation": organisation,
         "q": q,
+        "sort_by": sort_by,
+        "sort_direction": sort_direction,
+        "sort_options": _sort_options_context(
+            q, sort_by, sort_direction, SORT_FOR_PUBLISHED
+        ),
     }
```
mikerkelly, line 55: This is almost identical between the two view functions. Maybe this can be a shared function?

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
         queryset=Handle.objects.filter(is_current=True).select_related("organisation"),
     )
+
+
+def _get_sort(request, status):
+    query_sort_direction = request.GET.get("direction")
+    if query_sort_direction in SORT_DIRECTIONS:
+        sort_direction = query_sort_direction
+    else:
+        sort_direction = SORT_DIRECTION_ASC
```
mikerkelly, line 142: Combined with the StrEnum I suggest above, maybe this could be:

```python
class SortOrder(StrEnum):
    ASC = auto() 
    DESC = auto()

....

query_order = request.GET.get("order")
order = query_order if query_order in SortOrder else SortOrder.ASC
```

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+    sort_key_funcs = {
+        SORT_BY_NAME: lambda codelist: codelist.name.casefold(),
+        SORT_BY_UPDATED_AT: lambda codelist: codelist.updated_at,
+    }
+
+    return sorted(
+        codelists,
+        key=sort_key_funcs[sort_by],
+        reverse=sort_direction == SORT_DIRECTION_DESC,
+    )
```
mikerkelly, line 169: `sorted` returns a new list, a list was already passed in. `sort` should be called on the list to sort it in place, else we need to allocate all that memory again, which is expensive.

... all that would be avoided if we sort with QuerySets `order_by`. I think that would be a nice-to-have for this PR but not essential.

codelists/views/organisations.py
mikerkelly: Why do we have `sort_options`  the context as a sub-dict but also have each sort option as a top-level entry in the context dict?

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+            for page_number in paginator.page_range
+        ],
+    }
+
+
+def _sort_options_context(q, current_sort_by, current_sort_direction, sort_options):
+    return {
+        sort_by: _sort_context(q, current_sort_by, current_sort_direction, sort_by)
+        for sort_by in sort_options
+    }
```
mikerkelly, line 219: Question: what is the purpose of this iteration? Don't we only ever allow one search term?

templates/_partials/organisation_table.html
```diff
@@ -8,14 +8,14 @@
           <thead class="border-b-slate-300 bg-slate-100 text-left text-sm font-semibold text-slate-900">
             <tr>
               {% if under_review_codelists %}
-                <th class="px-3 py-3.5 pl-4 whitespace-nowrap sm:pl-6">Name</th>
+                {% include "_partials/organisation_table_sort_header.html" with column_sort=sort_options.name label="Name" th_class="px-3 py-3.5 pl-4 whitespace-nowrap sm:pl-6" only %}
```
mikerkelly, line 11: Could this also be called `sort_options` in  "_partials/organisation_table_sort_header.html"? Otherwise it is quite hard to start from the view function and trace through to understand where it is finally used.

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+
+
+def _sort_context(q, current_sort_by, current_sort_direction, sort_by):
+    sort_direction = SORT_DIRECTION_ASC
+    if current_sort_by == sort_by and current_sort_direction == SORT_DIRECTION_ASC:
+        sort_direction = SORT_DIRECTION_DESC
```
mikerkelly, line 225: Question: why are we inverting `sort_direction` in this case? Could this be commented on in the code please?

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+
+    aria = "none"
+    icon = "none"
+    if current_sort_by == sort_by:
+        icon = current_sort_direction
+        aria = (
+            "ascending"
+            if current_sort_direction == SORT_DIRECTION_ASC
+            else "descending"
+        )
```
mikerkelly, line 235: Separation of concerns: deciding which exact HTML attributes to pass to an element feels like a distinctly template-y concern. Is the current location the best place for this logic?

If so I think that a brief explanation of what this is for would be useful.

codelists/views/organisations.py
```diff
@@ -15,32 +17,64 @@
 
 PAGE_SIZE = 50
```
mikerkelly, line 18: 50 feels like a lot of items for one page, I have to scroll down a couple of times just to see them all. Could this be more like 15 or 20?

codelists/tests/views/test_organisations.py
```diff
@@ -1,8 +1,19 @@
...
+import pytest
+from django.utils import timezone
+
 from codelists.actions import publish_version
-from codelists.models import Handle, Status
+from codelists.models import Codelist, CodelistVersion, Handle, Status
 from opencodelists.list_utils import flatten
 
 
+def aware_datetime(year: int, month: int, day: int) -> datetime:
```
mikerkelly, line 11: I think it will be simpler to use the `freezer` pytest extension fixture to move time before creating each codelist.

If you want to do it declaratively without `freezer` then the required datetime should be a parameter to `create_codelist` so that functionality can be shared between tests, not special a helper just for this module.

codelists/tests/views/test_organisations.py
```diff
@@ -51,14 +62,76 @@ def test_paginate_published_codelists(client, organisation, create_codelists):
...
+            ["SNOMED Bravo", "SNOMED alpha", "SNOMED Charlie"],
+        ),
+    ],
+)
+def test_published_codelists_search_and_sort(
+    client, organisation, create_codelist, sort, direction, expected_names
+):
+    bravo = create_codelist("SNOMED Bravo")
+    alpha = create_codelist("SNOMED alpha")
+    charlie = create_codelist("SNOMED Charlie")
```
mikerkelly, line 120: Maybe worth a brief comment explanation that the creation order is BaC but the case-insenstive lexicographic order is aBC. That makes it really obvious why these are called what they are.

You can also add `id` parameters to pytest test parametrisation arguments to hint as to the purpose of each parametrisation.

codelists/tests/views/test_organisations.py
```diff
@@ -51,14 +62,76 @@ def test_paginate_published_codelists(client, organisation, create_codelists):
...
+    ],
+)
+def test_organisation_pagination_preserves_search_and_sort_query(
+    client, organisation, create_codelists, path, status
 ):
-    create_codelists(70, owner=organisation, status=Status.PUBLISHED)
+    create_codelists(70, owner=organisation, status=status)
 
-    rsp = client.get(f"/codelist/{organisation.slug}/", {"q": "Codelist"})
+    rsp = client.get(
```
mikerkelly, line 77: Question: do we really need to use the test `client` fixture in these tests? Would it be sufficient to use the `rf` request factory fixture instead and call the view code directly?

I ask because using the `client` makes a full real HTTP request to the test server, which is much slower than `rf`.

Using `client` makes this more like what in Job Server we call an integration test, of high level behaviour, with separate `rf` tests for the more numerous and detailed unit tests.

codelists/views/organisations.py
mikerkelly: We can pass the handles queryset directly to the paginator, we don't need to convert it into a list. The templates can access ``{{handle.codelist}}`` fine.

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+    }
+
+    return sorted(
+        versions,
+        key=sort_key_funcs[sort_by],
+        reverse=sort_direction == SORT_DIRECTION_DESC,
+    )
+
+
+def _pagination_context(page_obj, q, sort_by, sort_direction):
```
mikerkelly, line 185: I think it would be quite a bit cleaner to extend `Paginator` to either overload its methods or add new ones, if we want to do this logic in the view Python code. Don't have time to expand now, sorry.

codelists/views/organisations.py
```diff
@@ -98,3 +132,116 @@ def _current_handle_prefetch(lookup: str) -> Prefetch:
...
+
+    return sorted(
+        codelists,
+        key=sort_key_funcs[sort_by],
+        reverse=sort_direction == SORT_DIRECTION_DESC,
+    )
+
+
+def _sort_under_review_versions(versions, sort_by, sort_direction):
+    sort_key_funcs = {
```
mikerkelly, line 173: If we use `order_by` to do QuerySet sorting as suggested elsewhere, this dict could be at the module level and take a `(SortField, SortOrder)` tuple as key, and map to the argument needed to `order_by`. Then function is not needed, we can do something like 


```python
ordering_params = {
    (SortField.NAME, SortOrder.ASC): "name__iexact"
    (SortField.NAME, SortOrder.DESC): "-name__iexact"
    ...
}
...
.order_by(ordering_params[(sort_field, sort_order)])

mikerkelly (commented):
Thanks, Thomas. This seems to mostly do the right thing on local testing and the UI looks good. I like the URL parameter design which is clear and functional.

I think that the overall idea of separating out the organisation list views from the index but with some shared code is a very good one, same for the partial templates. Much of the functionality is along the right lines, but I think we can make some simplifications which will make this much easier to develop and maintain. I can also see that some of the issues I am having with this come from the index code that this was split off from and are longstanding.

I have been finding this module hard to understand and review and so I am 
going to make some fairly extensive review comments. Given all that, I am
very happy to assist however I can, whether that is futher review, discussion,
pairing, or doing some or all of the work I am asking for. Let me know what you
think, and feel free to push back. I definitely think we can leave many of these 
thoughts for the future if at all.  I still have not looked in detail at the
templates or testing.

Probably I have left too many comments, but did not want to lose any thoughts. Sorry if there is too much text and it is unstructured and repetitive.

Some general comments below, more inline.

Correctness of template use of pagination
=========================================

The `Pagination` object seems to have been set up right but the context has and template tries to use information about First and Last buttons and page numbers. I never see these, even on an organisation with many pages so I think something may be wrong in the logic in one of those but have not had time to look into that. Could you investigate, or let me know where I can see it working if you have.

I tried going to http://localhost:7000/codelist/opensafely/ which has 12 pages
that I went through and never see the links.

Code reuse and organisation
===========================

It seems like the index and organisation codelist pages do most of the sorting logic in Python whereas https://github.com/opensafely-core/opencodelists/pull/3029 was written in JavaScript. Is there a reason for the difference? We now have two separate implementations of sorting a list of codelists for different pages. I think I like the Python implementation for this but could we share code for listing, sorting, and filtering codelists in future PR(s)?

Within views/organisations.py, the module is structured as two separate view functions sharing a lot of private utility functions to handle the differences between the views for published and under-review codelists. This does the job, but it introduces a lot of indirection that I am finding makes it quite hard to follow and reason about the differences.

I think for this PR we should worry less about good query performance and DRY code and more about making the QuerySet the view makes easy to understand and maintain. We can do some refactoring to add custom QuerySets and share code if we keep developing this.

I would like us to try and reduce the number of private helpers down so that there is less indirection and the top-level flow in the view is clearer.  Each view is extracting and validating some URL params, making a QuerySet, and constructing a context including a Paginator.  Probably those should be the top level helpers, if any. Inline review comments expand on this theme.

All of the QuerySet filtering and sorting for Codelist, CodelistVersion, and Handle get quite complicated here and in `index`. Probably this should not be in the views at all but somewhere in the models as a model queryset or manager, or in a separate queries module that the view can then call out to with the necessary parameters. I don't necessarily think that we should do anything about that in this PR.

I think it will be much easier for this PR to reason about the QuerySet if we construct one QuerySet with the generally in-scope handles (right org, current, not private, published version exists) then compose that with `search_handles`. And the part that does the composition should ideally all be done in one place so it can be understood, then detail looked at if required, not split across the `organisation_review`, `_get_organisation_handles`, `_handles_with_status`, and `_under_review_versions` functions. I think I am finding it too hard to understand and reason about the way it is currently organised. 

I have a comment near the start of `organisation_published` show what that could look like.

For the future, if we want to be more DRY and have shared queryset functions like "saved" `filter`s and `order_by`s, I think an idiomatic way to do this is to extend the QuerySet class with some custom methods that also return QuerySets, and use that as the model manager. That approach is explained in the documentation [here](https://docs.djangoproject.com/en/6.0/topics/db/managers/#creating-a-manager-with-queryset-methods) and a nice example of what that looks like in [this Julia Evans post](https://jvns.ca/blog/2026/07/21/more-nice-django-things/#i-m-enjoying-query-builders).

Sorting
=======

This PR sorts a QuerySet of Django model instances into an ordered list for display. Django QuerySets have a composable `order_by` method that should be used for ordering queryest as required. We should only do the sort ourselves in Python as a last resort, if there is a good reason not to do in with `order_by`. Using QuerySet methods lets us use the nice API the framework provides, write less code ourselves, and will do the work in the database when required, which will probably be orders of magnitude more performant.

https://docs.djangoproject.com/en/6.0/ref/models/querysets/#django.db.models.query.QuerySet.order_by

On performance: It took me 13s to open http://localhost:7000/codelist/opensafely/under-review/ on first visit, of which only 4ms was database queries. After that visits (such as changing sort order) took about 75-150ms, of which almost all was CPU time. http://localhost:7000/codelist/opensafely/ was similar. I am unsure if the initial visit was my local database warming up or populating caches or just a blip. ~100ms delay is something we could ignore as not materially affecting UX if we wanted, but we can do better that that and should use QuerySet for consistency reasons anyway.

Back on https://github.com/opensafely-core/opencodelists/commit/dc7645163294e5b7cc2a2893e3cf078cef5cdc4d the page took about 50-60ms to load, about 3ms on database queries. An extra 50-100ms is not a dealbreaker as performance is not very important to us, but hints how expensive unpacking into Python lists and sorting in Python can be, compared to doing it in the database. And we should be careful about compounding such performance hits.

PR #3139, opened 2026-07-06
Refactor ICD-10 load process to match standard pattern

Jongmassey:
Needs thorough end to end checks before deploy. 

We may wish to consider a retention policy for downloaded files, especially the un-zipped xmls that are never loaded because they are found to be the same as previously loaded ones


Jongmassey:
- [ ] Update README
- [ ] E2E/integration tests

rw251:
The pattern for the other importers seems to be that the data downloader throws an error if the release already exists. This is then handled by `import_latest_data.py` management command, which prints a message and exits _before_ calling the import function. For out current set up here, if there are any changes to any input, then the import function will throw an error - but if there are no changes I think it will just proceed to create a new release which would be identical to the previous one. I wonder if we can take advantage of the fact that in order for a new release (based on actual changed data) to work, you'd need to update the known_diffs.py file. i.e. maybe we put the `release_name` hardcoded somewhere and the importer only moves to import if the `release_name` has changed - and instructions to only update that once you've modified the `known_diffs.py` file to fix the errors?

Jongmassey:
>The pattern for the other importers seems to be that the data downloader throws an error if the release already exists. This is then handled by import_latest_data.py management command, which prints a message and exits before calling the import function.

Happy to move the call to release builder to be in the downloader

>but if there are no changes I think it will just proceed to create a new release which would be identical to the previous one

We definitely don't want this behaviour. I'll take a quick look at making the downloader throw an error if there are no changes so it matches the other coding systems.

rw251:
> making the downloader throw an error

I thought about that, but I'm not sure if that means running all the parsing stuff from the downloader - unless it's just comparing the claml files for any changes

Jongmassey:
Yep, you're right. 

If the other coding systems' Downloaders throw errors when the release is the same, then I think we can make the icd10 importer throw a similar error when no changes are found (i.e. the release is the same).

The outer shell script looks for the string "already exists for the latest release" in the error message and posts a "no new release" message to sentry and sets the cron status to success. All other errors lead to a non-zero return and the error message being posted to slack (via sentry). So long as we copy the pattern for the "no new release" error message then all should be fine.

Jongmassey:
I reckon we just put the combined file hash in the release name, then the downloader will know if the newly assembled file is the same as previous.

Jongmassey:
e.g.

```python
>>> import zipfile
>>> import hashlib
>>> from pathlib import Path
>>> tmp = Path("/tmp")
>>> (tmp / "foo.txt").write_text("12345")
5
>>> (tmp / "bar.txt").write_text("abcdef")
6
>>> with zipfile.ZipFile((tmp / "combined.zip"), "w", zipfile.ZIP_DEFLATED) as zf:
...     zf.write((tmp/"foo.txt"))
...     zf.write((tmp/"bar.txt"))
... 
>>> md5_1 = hashlib.md5((tmp / "combined.zip").read_bytes()).hexdigest()
>>> md5_1
'66382ced9619599d7795f4293a2e8204'
>>> (tmp / "bar.txt").write_text("abgdef")
6
>>> with zipfile.ZipFile((tmp / "combined.zip"), "w", zipfile.ZIP_DEFLATED) as zf:
...     zf.write((tmp/"foo.txt"))
...     zf.write((tmp/"bar.txt"))
... 
>>> md5_2 = hashlib.md5((tmp / "combined.zip").read_bytes()).hexdigest()
>>> md5_2
'ea9e7f6b7299643cb093b46b70ea6618'
>>> with zipfile.ZipFile((tmp / "combined.zip"), "w", zipfile.ZIP_DEFLATED) as zf:
...     zf.write((tmp/"foo.txt"))
...     zf.write((tmp/"bar.txt"))
... 
>>> md5_3 = hashlib.md5((tmp / "combined.zip").read_bytes()).hexdigest()
>>> md5_3
'ea9e7f6b7299643cb093b46b70ea6618'
```

Jongmassey:
Scratch that, the changing scraped filename (due to timestamp, included for easier debugging) means we can't rely on the zip file content's has as its file manifest will change.

We can, however, use the hashes of the contents of each of the three xml files and this seems to work well (I've added a commit to this effect).

Need to double-check that the scraped ClaML generation is idempotent

rw251:
Local testing through up a couple of things which I've addressed in the last 4 commits. Two other things we may want to change:
- We may want to think about the names - it's a bit unwieldy in the UI at present e.g.
image

Maybe just the first 8 chars of the hash? So it becomes:
image
- We should probably add *.xml and *.zip to the .gitignore for `coding_systems/icd10/data/`

coding_systems/icd10/migrations/0003_migrate_existing_data_to_transitional_model.py
Jongmassey, line 1: Looks good. We can squash this with the original migration for clarity

coding_systems/icd10/scraped_to_claml.py
```diff
@@ -37,6 +37,24 @@ def tbal_id():
...
+    global SECTIONAL_MODIFIER_ID_COUNT
+    global classes_to_write
+    global modifiers_to_apply
+
+    D_ID_COUNT = count(1)
+    TBAL_ID_COUNT = count()
+    SECTIONAL_MODIFIER_ID_COUNT = count(1)
+    classes_to_write = []
+    modifiers_to_apply = defaultdict(list)
+
```
Jongmassey: I don't understand why this is necessary

Jongmassey:
> Maybe just the first 8 chars of the hash?

Agree

coding_systems/icd10/scraped_to_claml.py
rw251: To ensure the ids in the claml are always the same even with repeated calls - but I reckon you knew that and probably meant why is it necessary given the management command will only call `convert_chapters_to_claml()` once. I suppose it's just future proofing so that even if you called the method multiple times in a single session (e.g. maybe in tests?) that the output is always the same.

PR #3151, opened 2026-07-17
Add the ICD-10 warning banners

### LINKED ISSUE #3123: Displaying a warning banner for ICD10 codes with definitions that change meaningfully between 2016 and 2019
There are some codes whose meaning changed between 2016 (apcs data) and 2019 (ons deaths data). Users with codelists containing these codes should be presented with a warning banner so they can decide what to do about it.

Considerations:
- different message if building a codelist / viewing my own codelist / viewing someone else's codelist
- need to get the balance right between a detailed explanation (it's not straightforward) and something that isn't very intrusive

### LINKED ISSUE #3124: Display a warning banner if a codelist has a "moved" ICD10 code but not its partner
There are some concepts that are represented with different codes in the 2016 vs 2019 editions. I.e. they have moved. For example "Post COVID condition" is `U074` in 2016, but `U08` and `U089` in 2019. If a codelist has one but not all of these codes it is likely a mistake as you would miss events. We should display a banner warning users of this.

Considerations:
- different message if building a codelist / viewing my own codelist / viewing someone else's codelist
- need to get the balance right between a detailed explanation (it's not straightforward) and something that isn't very intrusive

rw251:
Fixes #3123. Fixes #3124

### DIFF
--- a/builder/tests/test_views.py
+++ b/builder/tests/test_views.py
@@ -10,6 +10,10 @@ from codelists.tests.views.assertions import (
     assert_post_unauthenticated,
     assert_post_unauthorised,
 )
+from coding_systems.icd10.known_diffs import (
+    COMBINED_2016_VS_2019_DIFFERENCES,
+    ReleaseTermDifference,
+)
 from opencodelists.tests.assertions import assert_difference
 
 
@@ -164,6 +168,60 @@ def test_new_search_for_code(client, draft):
     assert rsp.redirect_chain[-1][0] == draft.get_builder_search_url(
         last_search_id, "code:128133004"
     )
+
+
+def test_new_search_flags_clinically_inequivalent_icd10_codes(
+    client, icd10_data, monkeypatch, organisation, organisation_user
+):
+    monkeypatch.setitem(
+        COMBINED_2016_VS_2019_DIFFERENCES,
+        "M770",
+        ReleaseTermDifference(
+            combined_2016="Old term",
+            who_2019="New term",
+            clinically_equivalent=False,
+        ),
+    )
+    codelist = create_codelist_from_scratch(
+        owner=organisation,
+        author=organisation_user,
+        name="ICD10 draft",
+        coding_system_id="icd10",
+        coding_system_database_alias="icd10_test_20200101",
+    )
+    draft = codelist.versions.get()
+    client.force_login(organisation_user)
+
+    rsp = client.post(
+        draft.get_builder_new_search_url(),
+        {"search": "M770", "search-type": "code"},
+        follow=True,
+    )
+
+    assert rsp.status_code == 200
+    assert rsp.context["icd10_term_differences"] == {
+        "M770": {"combined_2016": "Old term", "who_2019": "New term"}
+    }
+
+
+def test_builder_known_difference_flags_are_icd10_only(
+    client, draft_with_complete_searches, monkeypatch
+):
+    monkeypatch.setitem(
+        COMBINED_2016_VS_2019_DIFFERENCES,
+        "202855006",
+        ReleaseTermDifference(
+            combined_2016="Old term",
+            who_2019="New term",
+            clinically_equivalent=False,
+        ),
+    )
+    client.force_login(draft_with_complete_searches.author)
+
+    rsp = client.get(draft_with_complete_searches.get_builder_draft_url())
+
+    assert rsp.status_code == 200
+    assert rsp.context["icd10_term_differences"] == []
 
 
 def test_new_search_no_results(client, draft):
--- a/coding_systems/icd10/known_diffs.py
+++ b/coding_systems/icd10/known_diffs.py
@@ -1002,7 +1002,10 @@ class RubricChange:
         final list of values after applying the changes.
         """
         rubrics = {
-            rubric_type: list(values) for rubric_type, values in self.who_2016.items()
+            # Using list() to create a shallow copy of the list to avoid mutating
+            # the original WHO rubrics
+            rubric_type: list(values)
+            for rubric_type, values in self.who_2016.items()
         }
 
         for rubric_type, replacements in self.replace.items():
--- a/coding_systems/icd10/known_diffs.py
+++ b/coding_systems/icd10/known_diffs.py
@@ -977,3 +977,496 @@ def moved_codes(codes):
         if any(code in normalised_codes for code in possible_codes):
             moved_codes.append(moved_code_set)
     return moved_codes
+
+
+@dataclass
+class RubricChange:
+    """
+    Represents a change to the rubrics of a code between releases. This can
+    include additions, removals, and replacements of rubric text. The `who`
+    field represents the original rubrics from the WHO release, while the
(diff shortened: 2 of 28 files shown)


coding_systems/icd10/known_diffs.py
```diff
@@ -882,3 +882,98 @@ def get_2016_2019_description_difference(
...
+                "who_2019": difference.who_2019,
+            }
+    return differences
+
+
+# Moved code sets. We flag codelists that contain only part of one of these sets,
+# because the same clinical concept is represented by different codes across releases.
+# These moved codes were derived from this report (https://github.com/bennettoxford/icd-browser-scraper/blob/main/claml/2016-vs-2019.md)
+# which shows all the codes that appear in one set and not the other.
+MOVED_CODE_SETS = (
```
Jongmassey, line 909: Convention is to put consts at the top of the file so they're easy to spot

coding_systems/icd10/known_diffs.py
```diff
@@ -882,3 +882,98 @@ def get_2016_2019_description_difference(
...
+    {
+        "title": "COVID-19 vaccines causing adverse effects in therapeutic use",
+        "nhs2016": ["U077"],
+        "who2019": ["U12", "U129"],
+        "comment": "This is U077 in 2016, but U12/U129 in 2019.",
+    },
+)
+
+
+def moved_codes(codes):
```
Jongmassey: why type annotations elsewhere but not here?

I'm ambivalent about partial introduction throughout a codebase, but think should probably be consistent within the same file

coding_systems/icd10/tests/test_known_diffs.py
```diff
@@ -135,3 +137,44 @@ def test_expand_who_2016_place_of_occurrence_fails_when_expected_override_missin
 
     with pytest.raises(AssertionError, match="expected codes missing"):
         expand_who_2016_place_of_occurrence(records, modifiers)
+
+
+def test_clinically_different_codes():
```
Jongmassey, line 217: A possibly overly philosophical questions is what this test is actually testing (see also moved codes test below)?

This is testing both that the logic of the test is correct and that the content of the `COMBINED_2016_VS_2019_DIFFERENCES` constant is as expected.

The purist in me would want to monkeypatch the const and have this test test the logic of the function in isolation.

coding_systems/icd10/known_diffs.py
```diff
@@ -977,3 +977,496 @@ def moved_codes(codes):
...
+@dataclass
+class RubricChange:
+    """
+    Represents a change to the rubrics of a code between releases. This can
+    include additions, removals, and replacements of rubric text. The `who`
+    field represents the original rubrics from the WHO release, while the
+    `remove`, `add`, and `replace` fields represent the changes to be applied.
+    """
+
+    who: dict[str, list[str]] = field(default_factory=dict)
```
Jongmassey: which WHO? 2016 or 2019?

coding_systems/icd10/known_diffs.py
```diff
@@ -977,3 +977,496 @@ def moved_codes(codes):
...
+    remove: dict[str, list[str]] = field(default_factory=dict)
+    add: dict[str, list[str]] = field(default_factory=dict)
+    replace: dict[str, dict[str, str]] = field(default_factory=dict)
+    comment: str = ""
+
+    @property
+    def resolved_use(self) -> dict[str, list[str]]:
```
Jongmassey: what does "use" mean in this context?

coding_systems/icd10/known_diffs.py
```diff
@@ -977,3 +977,496 @@ def moved_codes(codes):
...
+        """
+        Resolve the rubric changes by applying removals, additions, and replacements
+        to the original WHO rubrics. Returns a dictionary of rubric types to their
+        final list of values after applying the changes.
+        """
+        rubrics = {
+            rubric_type: list(values) for rubric_type, values in self.who.items()
```
Jongmassey: the type annotation for `who` says the values should already be a list?

coding_systems/icd10/known_diffs.py
Jongmassey, line 1: I meant to mention this in the earlier iteration of this file, but I feel like having the lookups for the known diffs and the code all in one massive file feels a bit unwieldy. Even more so now it's c.1.5kloc

Would it be easier to maintain if the lists of known changes were serialised out to, for example, a set of json files?

builder/tests/test_views.py
```diff
@@ -166,6 +170,60 @@ def test_new_search_for_code(client, draft):
...
+    )
+
+    assert rsp.status_code == 200
+    assert rsp.context["icd10_term_differences"] == {
+        "M770": {"combined_2016": "Old term", "who_2019": "New term"}
+    }
+
+
+def test_builder_known_difference_flags_are_icd10_only(
+    client, draft_with_complete_searches, monkeypatch
```
Jongmassey, line 208: side note: I'd love to refactor all of these test fixture names to make it clear what coding system they are!

Jongmassey (approved):
AFAICT it works fine, I can't review the minutiae of the typescript very well but I've got some nits and some architectural thoughts

coding_systems/icd10/known_diffs.py
rw251: I lean towards the partial introduction - especially here when it's a brand new file that I can apply it in consistently. But good spot that it's not complete. I'll fix

coding_systems/icd10/known_diffs.py
rw251: The cast is actually just to shallow copy "who" so that we don't mutate the original list. I'm happy to add a comment to that effect if you think it would be helpful in future.

coding_systems/icd10/known_diffs.py
rw251: The known_diffs refactor mentioned later that we've also talked about will resolve this in the feature branch

coding_systems/icd10/known_diffs.py
rw251: Changed to `who_2016` and updated the comment

coding_systems/icd10/known_diffs.py
rw251: Changed to `resolved_rubrics` as that's probably clearer

coding_systems/icd10/known_diffs.py
rw251: As per discussion we'll refactor this in the feature branch

coding_systems/icd10/tests/test_known_diffs.py
rw251: It's a good point. But I'll wait until making the known_diffs refactor before updating the tests in case that changes things.

coding_systems/icd10/known_diffs.py
rw251: Comment to that effect added

builder/tests/test_views.py
rw251: I'll leave that on your todo list for now!

PR #3153, opened 2026-07-17
Add extra rubrics and term differences to more info modal

### LINKED ISSUE #3126: Update UI for ICD-10 codes with different descriptions in 2016 vs 2019
The combined 2016/2019 ICD-10 release contains some codes whose descriptions have changed. If this change is significant then we will display a warning banner (#3123). But even if not a significant change, we may want to show the user the different descriptions. This ticket covers:
- [x] Deciding what to display in the main tree view (i.e. just 2019 or just 2016 definition)
- [x] Deciding what to display in the "more info" modal pop up
- [x] Implementing the changes

### LINKED ISSUE #3128: Update UI to display the extra rubric information for ICD-10 codes
In the new model for ICD-10 we now capture additional information for each code - called "rubrics". Rubrics fall into different types e.g. "inclusion" for equivalent terms/conditions included under the code, "exclusion" for things that aren't here and coded elsewhere, "coding-hint" for hints to help coders code etc.

Now that we store this information we should plan how we will present this to users.

- [x] Decide which rubric types to display | **Decision: All of them except the superfluous `modifier_link`**
- [x] Decide how to display each type | **Decision: (1) Within a box explaining what they are, (2) inc/exclusion as bullet lists (3) others as paragraphs**
- [x] Decide which rubric types to treat as synonyms during search | **Decision: Just the "inclusions" as they are the only "safe" synonyms**
- [x] Update the UI

rw251:
Fixes #3126 . Fixes #3128 .

**Extra rubrics**

- An example with a note and inclusion/exclusion rubrics

image

- If a concept has modifier rubrics and concept rubrics they are visually distinct:

image

**Term differences**

- Typos are shown in a green box with both definitions

image

- Clinically different definitions in a red box (NB the main thing users will see is a warning banner at the top of the page, this info in the more info modal is on top of that)

image

### DIFF
--- a/assets/src/scripts/components/Codelist/MoreInfoModal.tsx
+++ b/assets/src/scripts/components/Codelist/MoreInfoModal.tsx
@@ -2,7 +2,14 @@ import React, { useEffect, useState } from "react";
 import { Button, Modal } from "react-bootstrap";
 import type Hierarchy from "../../_hierarchy";
 import { getCookie, readValueFromPage } from "../../_utils";
-import type { Code, PageData, Status, Term } from "../../types";
+import type {
+  Code,
+  DaggerAsteriskInfo,
+  DaggerAsteriskUsage,
+  PageData,
+  Status,
+  Term,
+} from "../../types";
 
 interface CreateModalTextProps {
   allCodes: PageData["allCodes"];
@@ -341,9 +348,21 @@ interface MoreInfoModalProps {
   code: Code;
   codeToStatus: PageData["codeToStatus"];
   codeToTerm: PageData["codeToTerm"];
+  daggerAsteriskInfo?: DaggerAsteriskInfo;
   hierarchy: Hierarchy;
   status: Status;
   term: Term;
+}
+
+const daggerAsteriskGuidanceUrl =
+  "https://icd.who.int/browse10/Content/statichtml/ICD10Volume2_en_2019.pdf#page=29";
+
+function usageLabel(usage: DaggerAsteriskUsage) {
+  return usage === "dagger" ? "Dagger" : "Asterisk";
+}
+
+function usageGlyph(usage: DaggerAsteriskUsage) {
+  return usage === "dagger" ? "†" : "*";
 }
 
 function MoreInfoModal({
@@ -351,6 +370,7 @@ function MoreInfoModal({
   code,
   codeToStatus,
   codeToTerm,
+  daggerAsteriskInfo,
   hierarchy,
   status,
   term,
@@ -430,6 +450,17 @@ function MoreInfoModal({
 
   return (
     <>
+      {daggerAsteriskInfo && (
+        <button
+          className="builder__dagger-asterisk-link"
+          onClick={handleShow}
+          type="button"
+          aria-label={`Show ICD-10 ${daggerAsteriskInfo.usage} information for ${code}`}
+          title={`ICD-10 ${daggerAsteriskInfo.usage} code`}
+        >
+          <span>{usageGlyph(daggerAsteriskInfo.usage)}</span>
+        </button>
+      )}
       <Button
         className="builder__more-info-btn plausible-event-name=More+info+click"
         onClick={handleShow}
@@ -479,7 +510,7 @@ function MoreInfoModal({
               termDifferences={termDifferences}
             />
           )}
-          {hasRubrics(rubrics) && (
+          {(hasRubrics(rubrics) || daggerAsteriskInfo) && (
             <section className="border border-info rounded overflow-hidden mb-4">
               {/* WHO header */}
               <div className="bg-info text-white px-4 py-3">
@@ -496,6 +527,42 @@ function MoreInfoModal({
               <div className="bg-white pt-3 px-4">
                 {rubrics && hasRubrics(rubrics) && (
                   <RubricBlock rubrics={rubrics} term={term} />
+                )}
+                {daggerAsteriskInfo && (
+                  <section className="builder__additional-info-card builder__additional-info-card--usage">
+                    <h3 className="h6 font-weight-bold mb-3 d-flex align-items-center">
+                      Usage: {usageLabel(daggerAsteriskInfo.usage)}
+                      <span className="builder__dagger-asterisk-info__marker ml-2">
+                        {usageGlyph(daggerAsteriskInfo.usage)}
+                      </span>
+                    </h3>
+                    <p>
+                      Dagger codes identify the underlying condition; asterisk
+                      codes identify the manifestation.
+                    </p>
+                    <p className="mb-0 small">
+                      {daggerAsteriskInfo.url && (
+                        <>
+                          <a
+                            href={daggerAsteriskInfo.url}
+                            rel="noreferrer"
+                            target="_blank"
+                          >
+                            View {code} in the WHO ICD-10 browser
+                          </a>
+                          {" · "}
+                        </>
+                      )}
+                      <a
+                        href={daggerAsteriskGuidanceUrl}
+                        rel="noreferrer"
+                        target="_blank"
+                      >
+                        Read the full dagger/asterisk guidance (section 3.1.3 -
+                        p20)
+                      </a>
+                    </p>
+                  </section>
                 )}
               </div>
             </section>
--- a/assets/src/scripts/components/Codelist/MoreInfoModal.tsx
+++ b/assets/src/scripts/components/Codelist/MoreInfoModal.tsx
@@ -13,6 +13,187 @@ interface CreateModalTextProps {
   status: Status;
 }
 
+type RubricValues = Record<string, string[]>;
+
+interface CodeRubrics {
+  concept_rubrics?: RubricValues;
+  modifier_rubrics?: Record<string, RubricValues>;
+}
+
+const rubricDisplayNames: Record<string, string> = {
+  exclusion: "Excludes:",
+  inclusion: "Includes:",
+  "coding-hint": "Coding hint:",
+};
+
+const rubricOrder = [
+  "definition",
+  "text",
+  "note",
+  "inclusion",
+  "exclusion",
+  "coding-hint",
+];
+const italicRubricKinds = ["footnote", "text", "note", "definition"];
+const inlineHeaderRubricKinds = ["coding-hint"];
+const groupedConceptRubricKinds = ["inclusion", "exclusion"];
+
+function emptyRubrics(): CodeRubrics {
+  return {
+    concept_rubrics: {},
+    modifier_rubrics: {},
+  };
+}
+
+function formatRubricKind(kind: string) {
+  return rubricDisplayNames[kind] || kind.replaceAll("_", "-");
+}
+
+function sortAndFilterRubricEntries(rubrics: RubricValues) {
+  return Object.entries(rubrics)
+    .filter(([kind]) => kind !== "modifierlink")
+    .sort(([left], [right]) => {
+      const leftIndex = rubricOrder.indexOf(left);
+      const rightIndex = rubricOrder.indexOf(right);
+
+      if (leftIndex !== -1 || rightIndex !== -1) {
+        return (
+          (leftIndex === -1 ? rubricOrder.length : leftIndex) -
+          (rightIndex === -1 ? rubricOrder.length : rightIndex)
+        );
+      }
+
+      return left.localeCompare(right);
+    });
+}
+
+function hasRubricValues(rubricValues?: RubricValues) {
+  return Object.values(rubricValues || {}).some((values) => values.length > 0);
+}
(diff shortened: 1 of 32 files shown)


coding_systems/base/coding_system_base.py
```diff
@@ -105,6 +105,9 @@ def lookup_synonyms(self, codes):  # pragma: no cover
     def lookup_references(self, codes):  # pragma: no cover
         return {}
 
+    def lookup_additional_rubrics(self, codes):  # pragma: no cover
```
Jongmassey: In what sense are they additional?

This feels like a slight abstraction leak since only ICD10 has them.

Given the only use of reference, synonym, and now rubric lookups is in the "more info modal" view:

```py
[...]
synonyms = {"synonyms": cs.lookup_synonyms(codes)}
references = {"references": cs.lookup_references(codes)}
additional_rubrics = cs.lookup_additional_rubrics(codes)
return JsonResponse(synonyms | references | additional_rubrics)
[...]
```

might it be cleaner to do a pre-change refactor ("make the change easy, then make the change") to have a single `lookup_more_info(self, codes)`?

e.g. in `coding_system_base`

```py
def lookup_more_info(self,codes): # pragma; no cover
    return {}
```

then each coding system just adds the bits it has to the dict, e.g. dmd would change to

```py
def _lookup_synonyms(self, codes):
    descriptions = (
        AMP.objects.using(self.database_alias)
        .filter(id__in=codes)
        .values("id", "nm")
    )

    result = collections.defaultdict(list)
    for d in descriptions:
        result[d["id"]].append(d["nm"])
    return dict(result)

def _lookup_references(self, codes):
    # OpenPrescribing's dm+d browser supports VTMs, VMPs, AMPs, VMPPs, and AMPPs
    # We need to know the type of the code to correctly form the URL for this.
    codes = set(codes)
    codes_and_types = []
    for model_cls in [AMP, VMP, AMPP, VMPP, VTM]:
        matched = (
            model_cls.objects.using(self.database_alias)
            .filter(id__in=codes)
            .values_list("id")
        )
        codes_and_types += [
            (code[0], model_cls.__name__.lower()) for code in matched
        ]

    return {
        code: [
            (
                "OpenPrescribing dm+d browser",
                f"https://openprescribing.net/dmd/{codetype}/{code}/",
            )
        ]
        for code, codetype in codes_and_types
    }
def lookup_more_info(self, codes):
    return {
        "synonyms" : self._lookup_synonyms(codes),
        "references" : self._lookup_references(codes)
    }
```

and ICD-10 would look like 

```py
def _lookup_additional_rubrics(self, codes):
    [...]

def lookup_more_info(self, codes):
    return { "rubrics": self._lookup_additional_rubrics(codes) }
```


I introduced "lookup references" when I added references to dm+d, I think my thinking at the time was that every coding system would ultimately have per-code(s) references lookups but that was probably unrealistic. 

I don't think adding coding-system-specific methods to the base class is good practice in this case.

coding_systems/icd10/coding_system.py
```diff
@@ -114,6 +121,91 @@ def code_to_term(self, codes):
         unknown = set(codes) - set(lookup)
         return {**lookup, **{code: "Unknown" for code in unknown}}
 
+    def lookup_additional_rubrics(self, codes):
```
Jongmassey: I honestly found this a bit hard to follow. I've not really used `setdefault` much which probably doesn't help.

I spent a bit of time trying to rewrite this just to see if there was an obvious way that was clearer and I'm not sure I've entirely succeeded! Here's my attempt (which passes tests):

```py
concept_rubrics_to_return = defaultdict(lambda: defaultdict(list))
for code, kind, text in concept_rubrics:
    concept_rubrics_to_return[code][kind].append(text)

parent_concept_rubrics = defaultdict(list)
for parent_code, kind, text in modifier_concept_rubrics:
    parent_concept_rubrics[parent_code].append((kind, text))

for code, parent_code in modifier_parent_codes.items():
    for kind, text in parent_concept_rubrics[parent_code]:
        concept_rubrics_to_return[code][kind].append(text)

modifier_rubrics_to_return = defaultdict(
    lambda: defaultdict(lambda: defaultdict(list))
)
for code, term_modifier, kind, text in modifier_rubrics:
    term_modifier = term_modifier or "Modifier"
    modifier_rubrics_to_return[code][term_modifier][kind].append(text)

rubrics = {
    code: {
        "concept_rubrics": concept_rubrics_to_return.get(code, {}),
        "modifier_rubrics": modifier_rubrics_to_return.get(code, {}),
    }
    for code in (
        concept_rubrics_to_return.keys() | modifier_rubrics_to_return.keys()
    )
}
```


In the version under review there's a lot of function calls and some of them feel a bit unnecessary. For example why

```py
def empty_rubrics():
    return {"concept_rubrics": {}, "modifier_rubrics": {}}

def code_rubrics_for(collection, code):
    return collection.setdefault(code, empty_rubrics())
```
and not just 
```py
def code_rubrics_for(collection, code):
    empty_rubrics = {"concept_rubrics": {}, "modifier_rubrics": {}}
    return collection.setdefault(code, empty_rubrics)
```

Maybe it's just keeping track of how the state is modified or names that seem a bit unclear to me ("code_rubrics_for" in particular).

I think in this case at a minimum I'd like some comments and/or a fuller commit message to clarify the thinking here.

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
```diff
@@ -13,6 +13,187 @@ interface CreateModalTextProps {
...
+}
+
+function hasRubricValues(rubricValues?: RubricValues) {
+  return Object.values(rubricValues || {}).some((values) => values.length > 0);
+}
+
+function hasRubrics(rubrics: CodeRubrics | null) {
+  if (!rubrics) {
+    return false;
+  }
```
Jongmassey, line 90: TS n00b question - how does this ever return `true`?

coding_systems/icd10/known_diffs.py
```diff
@@ -902,6 +902,25 @@ def clinically_different_codes(codes: list[str]) -> dict[str, dict[str, str]]:
     return differences
 
 
+def different_codes(codes: list[str]) -> dict[str, dict[str, str | bool]]:
```
Jongmassey: nit: `different_codes` is a bit of an odd name for a function that 
>return[s] a dict containing the codes that have different descriptions

Jongmassey: perhaps `description_differences` ?

coding_systems/icd10/coding_system.py
```diff
@@ -124,7 +125,7 @@ def code_to_term(self, codes):
     def lookup_additional_rubrics(self, codes):
         codes = list(codes)
         if not codes:
-            return {"rubrics": {}}
+            return {"rubrics": {}, "term_differences": {}}
```
Jongmassey: why are term differences in "additional_rubrics"?

Jongmassey (changes requested):
There are quite a few things in here that don't entirely make sense to me. I'd really like to see some more explanation or them simplified.

Probably best to discuss in a call as maybe I've not made myself entirely clear in some of my feedback!

coding_systems/base/coding_system_base.py
rw251: This is a good shout - especially when you see the other PR that adds `lookup_dagger_asterisk_relations()`! Given the interrelated nature of the various UI pull requests, I propose that this refactor is added to the other one (known_diffs submodule) as something done in the feature branch once the individual UI changes are merged into it.

coding_systems/icd10/coding_system.py
rw251: Sorry - some of this stems from the fact that originally my code also pulled all rubrics from all ancestors for a code, so some of these helper methods made more sense. I rebased and pulled out all the ancestor code, but didn't fully review. I'm not sure I find your approach clearer than mine - but agree it would benefit from fewer helpers, and a couple of comments.

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
rw251: In JS, all variables can be coalesced to true/false. People talk about something being truthy or falsey. Things like `null`, `0`, `""` are all falsey, while `1`, `"string"` are truthy. An object or array, even if empty is truthy, so this pattern (`if (!rubrics)`) means that if rubrics isn't set (it's `null`) then we return false - otherwise we move to the next section.

coding_systems/icd10/known_diffs.py
rw251: I was reading it as a thing that returned codes with differences, hence different codes. But I see that I wasn't too convinced by that logic because the thing it's assigned to is `edition_description_differences`. So yes I agree `description_differences` or even `codes_with_description_differences` is better.

coding_systems/icd10/coding_system.py
rw251: I knew that we'd refactor this to `lookup_more_info()` at some point after we discussed it a while ago, so this is some sort of half way house to that. Apologies I probably could have saved you some time if I'd made that clear in the PR.

coding_systems/base/coding_system_base.py
Jongmassey: Yeah sure. Saves having to do it twice/reconciling the two

coding_systems/icd10/coding_system.py
Jongmassey: >I'm not sure I find your approach clearer than mine 

I'm not sure either!

I'll re-review once you've reduced the call depth and added some comments

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
Jongmassey: Thanks!

coding_systems/icd10/coding_system.py
Jongmassey: no worries, I'll resolve for now

coding_systems/icd10/coding_system.py
```diff
@@ -155,26 +138,29 @@ def add_modifier_rubric(collection, code, term_modifier, kind, text):
...
+        # Then we get the concept rubrics for any modifier codes. These are the
+        # rubrics for the parent code that the modifier code modifies.
+        inherited_concept_rubrics = (
             ConceptRubric.objects.using(self.database_alias)
             .filter(
-                concept_edition__concept_id__in=modifier_parent_codes.values(),
+                # The rubric belongs to the requested modifier code's parent.
                 concept_edition__edition_id=self.latest_edition.id,
+                concept_edition__concept__children__code__in=codes,
+                concept_edition__concept__children__concept_editions__edition_id=(
```
Jongmassey, line 153: I'm not sure we can filter to latest_edition here (or above?), as there are differences between concepts (and conceptrubrics?) between editions

```py
>>> e16_rubrics = ConceptRubric.objects.using(dba).filter(concept_edition__edition=e16)
>>> e19_rubrics = ConceptRubric.objects.using(dba).filter(concept_edition__edition=e19)
>>> e16_rubric_concepts= {cr.concept_edition.concept for cr in e16_rubrics}
>>> e19_rubric_concepts= {cr.concept_edition.concept for cr in e19_rubrics}
>>> e16_rubric_concepts.difference(e19_rubric_concepts)
{<Concept: Concept object (U076)>, <Concept: Concept object (U06)>, <Concept: Concept object (U077)>, <Concept: Concept object (P768)>, <Concept: Concept object (M139)>, <Concept: Concept object (B59)>, <Concept: Concept object (U073)>, <Concept: Concept object (K589)>, <Concept: Concept object (U075)>, <Concept: Concept object (U074)>}
```

Jongmassey: I think we need to prioritise rather than filter, as you did elsewhere

Jongmassey: example U06

```py
>>> ConceptRubric.objects.using(dba).get(concept_edition__concept__code="U06").text
'Codes U00-U49 are to be used by WHO for the provisional assignment of new diseases of uncertain etiology. In emergency situations codes are not always accessible in electronic systems. The specification of category U06 in the way it is done here will make sure this category and the subcategories are available in every electronic system at 
any time and that they can be used upon instruction by WHO, immediately.'
>>> ConceptRubric.objects.using(dba).get(concept_edition__concept__code="U06").concept_edition.edition
<Edition: Edition object (2016)>
```

or have I misunderstood?

rw251: ah yes - a very good point. I'll have a think, but yes I think prioritising NHS 2016 and falling back to WHO 2019 is the way.

Jongmassey (approved):

Jongmassey (approved):
lovely stuff

PR #3154, opened 2026-07-17
ICD-10 search changes

rw251:
- Extend search to use inclusion rubrics
- Default to the NHS 2016 definition

### DIFF
--- a/coding_systems/icd10/coding_system.py
+++ b/coding_systems/icd10/coding_system.py
@@ -1,10 +1,18 @@
 from collections import defaultdict
 from functools import lru_cache
 
+from django.db.models import Q
+
 from opencodelists.db_utils import query
 
 from ..base.coding_system_base import BuilderCompatibleCodingSystem
-from .models import Concept, ConceptEdition, ConceptKind, Edition
+from .models import (
+    Concept,
+    ConceptEdition,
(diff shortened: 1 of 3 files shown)


coding_systems/icd10/coding_system.py
Jongmassey: I didn't catch this in the earlier commit, but this feels like a bit of an abuse of `setdefault`, especially when it's being called multiple times per concept_id. There's a non-trivial performance penalty to that if my reading is correct.

How about replacing it with a simple dict comprehension:

```py
return {
    concept_id: f"{term} : {term_modifier}" if term_modifier else term
    for concept_id, term, term_modifier in concepts
}
```

(you'd need to invert the `preferred_edition` definition or sort order as `set_default` effectively take the first and the dict comprehension the last)

Jongmassey (approved):
Going to approve as it's a small set of changes that fundamentally work!

I do wonder about the appropriateness of the `set_default` call - I'm not sure we should be leaving performance on the table in this method that might get called a fair bit.

Jongmassey:
@rw251 prioritisation rewrite done - tweaked a test along the way to be more realistic.

Should squash this refactor into the previous commit before merging to feature branch

PR #3156, opened 2026-07-17
Add dagger/asterisk info to more info modal

rw251:
Appears in the tree view as:

image

With a tooltip on hover:

image

Clicking the dagger/asterisk just opens the more info modal, which looks like this:

image

The code links to the WHO ICD 2019 browser (all dagger asterisk codes in 2016 are also in 2019 so this is safe). The guidance link is to this: https://icd.who.int/browse10/Content/statichtml/ICD10Volume2_en_2019.pdf#page=29

### DIFF
--- a/assets/src/scripts/components/Codelist/MoreInfoModal.tsx
+++ b/assets/src/scripts/components/Codelist/MoreInfoModal.tsx
@@ -2,7 +2,14 @@ import React, { useEffect, useState } from "react";
 import { Button, Modal } from "react-bootstrap";
 import type Hierarchy from "../../_hierarchy";
 import { getCookie, readValueFromPage } from "../../_utils";
-import type { Code, PageData, Status, Term } from "../../types";
+import type {
+  Code,
+  DaggerAsteriskInfo,
+  DaggerAsteriskUsage,
+  PageData,
+  Status,
+  Term,
+} from "../../types";
 
 interface CreateModalTextProps {
   allCodes: PageData["allCodes"];
@@ -341,9 +348,21 @@ interface MoreInfoModalProps {
   code: Code;
   codeToStatus: PageData["codeToStatus"];
   codeToTerm: PageData["codeToTerm"];
+  daggerAsteriskInfo?: DaggerAsteriskInfo;
   hierarchy: Hierarchy;
   status: Status;
   term: Term;
+}
+
+const daggerAsteriskGuidanceUrl =
+  "https://icd.who.int/browse10/Content/statichtml/ICD10Volume2_en_2019.pdf#page=29";
+
+function usageLabel(usage: DaggerAsteriskUsage) {
+  return usage === "dagger" ? "Dagger" : "Asterisk";
+}
+
+function usageGlyph(usage: DaggerAsteriskUsage) {
+  return usage === "dagger" ? "†" : "*";
 }
 
 function MoreInfoModal({
@@ -351,6 +370,7 @@ function MoreInfoModal({
   code,
   codeToStatus,
   codeToTerm,
+  daggerAsteriskInfo,
   hierarchy,
   status,
   term,
@@ -430,6 +450,17 @@ function MoreInfoModal({
 
   return (
     <>
+      {daggerAsteriskInfo && (
+        <button
+          className="builder__dagger-asterisk-link"
+          onClick={handleShow}
+          type="button"
+          aria-label={`Show ICD-10 ${daggerAsteriskInfo.usage} information for ${code}`}
+          title={`ICD-10 ${daggerAsteriskInfo.usage} code`}
+        >
+          <span>{usageGlyph(daggerAsteriskInfo.usage)}</span>
+        </button>
+      )}
       <Button
         className="builder__more-info-btn plausible-event-name=More+info+click"
         onClick={handleShow}
@@ -479,7 +510,7 @@ function MoreInfoModal({
               termDifferences={termDifferences}
             />
           )}
-          {hasRubrics(rubrics) && (
+          {(hasRubrics(rubrics) || daggerAsteriskInfo) && (
             <section className="border border-info rounded overflow-hidden mb-4">
               {/* WHO header */}
               <div className="bg-info text-white px-4 py-3">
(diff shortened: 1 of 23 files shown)


coding_systems/base/coding_system_base.py
```diff
@@ -108,6 +108,9 @@ def lookup_references(self, codes):  # pragma: no cover
     def lookup_additional_rubrics(self, codes):  # pragma: no cover
         return {}
 
+    def lookup_dagger_asterisk_usages(self, codes):  # pragma: no cover
```
Jongmassey, line 111: For completeness, my thoughts here are the same as #3153 :)

coding_systems/icd10/coding_system.py
```diff
@@ -210,6 +211,39 @@ def add_modifier_rubric(collection, code, term_modifier, kind, text):
             "term_differences": edition_description_differences,
         }
 
+    def _icd10_browser_code(self, code):
```
Jongmassey: nit: If this is only used in `_icd10_browser_url`  can it be inlined there? Could reduce some clutter

coding_systems/icd10/coding_system.py
```diff
@@ -210,6 +211,39 @@ def add_modifier_rubric(collection, code, term_modifier, kind, text):
...
+        if len(code) > 3:
+            return f"{code[:3]}.{code[3:]}"
+        return code
+
+    def _icd10_browser_url(self, code):
```
Jongmassey: nit: If this is only used in `lookup_dagger_asterisk_usages`  can it be inlined there? Could reduce some clutter

coding_systems/icd10/coding_system.py
```diff
@@ -210,6 +211,39 @@ def add_modifier_rubric(collection, code, term_modifier, kind, text):
...
+        return f"https://icd.who.int/browse10/2019/en#/{self._icd10_browser_code(code)}"
```
Jongmassey: We prioritise NHS 2016 elsewhere but 2019 here. I don't think that's a major problem but are we ever likely to encounter a difference?

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
```diff
@@ -341,16 +347,29 @@ interface MoreInfoModalProps {
   code: Code;
   codeToStatus: PageData["codeToStatus"];
   codeToTerm: PageData["codeToTerm"];
+  daggerAsteriskUsage?: PageData["codeToDaggerAsteriskUsage"][string];
```
Jongmassey: n00b question, why `[string]` here and not the others?

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
```diff
@@ -341,16 +347,29 @@ interface MoreInfoModalProps {
...
   hierarchy: Hierarchy;
   status: Status;
   term: Term;
 }
 
+const daggerAsteriskGuidanceUrl =
+  "https://icd.who.int/browse10/Content/statichtml/ICD10Volume2_en_2019.pdf#page=29";
```
Jongmassey, line 358: If you've got, for example, the Adobe Acrobat extension installed the `#page=29` [doesn't work](https://community.adobe.com/questions-9/chrome-adobe-acrobat-extension-not-opening-a-doc-at-a-specific-page-with-page-pagenumber-in-url-1273762). I'm not sure that's anything we can do anything about. Just thought I'd mention. Adobe are very annoying.

Jongmassey (approved):
Like this a lot. Only nits/silly questions

coding_systems/icd10/coding_system.py
rw251: Agreed

coding_systems/icd10/coding_system.py
rw251: Do you mean inline the function (i.e. have a nested function), or just inline the code? I guess either way it's a short enough snippet, and easily understandable that it probably doesn't need the level of modularity I've introduced. Happy to change.

coding_systems/icd10/coding_system.py
rw251: I think it's unlikely. I'll be honest, I was partly driven by the fact the WHO urls are easier to construct - the NHS ones need you to know the block the code comes from. But also the 2019 dagger/asterisks are an extension of the 2016 ones, so every 2016 one appears in 2019 but no vice versa, so I think it's fine. I'll add a comment to this effect.

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
rw251: Oh that's interesting - does it still load the doc, but just not jump to the right page? If so then we could change the link text to say something like "Read the full guidance (p29)"

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
rw251: That's a good question. The short answer is that the other two are passed as dicts all the way down - because the status field of more info needs to traverse the hierarchy to output messages like "included because you included <diagnosis X>". But for the dagger/asterisk, we pass the dict down as far as the `Row`, at which point we can just lookup the code in the dict and pass the DaggerAsteriskUsage type to more info rather than the whole dict. So the code is saying pass the type of the value in that dict with str keys. But on reflection I should just actually put the type rather than this slightly convoluted, and non-intuitive way.

coding_systems/icd10/coding_system.py
rw251: Now inlined

coding_systems/icd10/coding_system.py
rw251: Also inlined

coding_systems/icd10/coding_system.py
rw251: Added comment to this effect in the code

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
rw251: Changed to use the actual type, rather than a lookup of the type. Realised that codeToDaggerAsteriskUsage should actually be codeToDaggerAsteriskInfo based on how I'd named the types, so had to change that in soooo many places. But done now, and much better for it.

assets/src/scripts/components/Codelist/MoreInfoModal.tsx
rw251: Added the section and page number to the link text. Flip flopped between p20 (what the page says at the bottom), or p29 (the actual page if including title pages etc). Went for former. Don't think it matters given I've also put the section number.

coding_systems/icd10/coding_system.py
Jongmassey: >I was partly driven by the fact the WHO urls are easier to construct

isn't that so!

Jongmassey (approved):
Really like the new commits.

Probably best to do any necessary squashing now to reduce the effort/likelihood of screwups doing it in the feature branch?

PR #3157, opened 2026-07-17
Allow "create new version" (in builder) for bnf-dmd codelists

### LINKED ISSUE #3031: Enhancement request: Convert BNF codelists to new-style dm+d codelists
# Summary
In investigating problems with codelist hierarchies,  I [found](https://bennettoxford.slack.com/archives/C096EMY1YDP/p1775809133071829) that none of the dm+d codelists that were converted from Psuedo-BNF were affected by the hierarchy non-round-tripping problem. This was slightly unexpected and revealed that the conversion process was happening  in a way that was slightly surprising to me.

[`convert_bnf_codelist_version_to_dmd`](https://github.com/opensafely-core/opencodelists/blob/d187b7036d41a0a7a65455d51bb7ac0ebab9af15/codelists/actions.py#L730) in `codelists/action.py` does this conversion. Broadly speaking, it grabs the name and slug from the source BNF codelist, appends "-dmd" to them, creates a methodology statement with a markdown link to the source codelist, calls the [`dmd_csv_data_for_download`](https://github.com/opensafely-core/opencodelists/blob/main/codelists/models.py#L956) method on the source codelist to get a csv of the converted dmd codes, then creates an _old-style_ new codelist with all of these.

Since the dm+d coding system is now a fully hierarchy-supporting Builder compatible coding system - I think we should consider changing this to create _new-style_ codelists.

# What would you like to achieve?
* Users cannot currently edit converted dm+d codelists in the builder, only convert or upload new versions
* The BNF to dm+d mapping file runs several months behind the main dm+d feed and so is often missing newer codes
* Creating a new style dm+d codelist would then allow a user to create a new version of the converted codelist which would open in the builder and thus show the missing codes and let them include them

# Who would benefit and how?

Users would be able to better maintain and adapt dm+d codelists initially created by converting them from BNF

Jongmassey:
Fixes #3031 

This adds the "Create New Version" button alongside the existing "Upload New Version" button for BNF to dmd converted codelists.

I chose to leave the Upload New Version button in place because users have previously uploaded new versions of converted codelists. We do not know if they do this only because they could not edit them in the builder. Nonetheless, because of the lack of support for additional columns (in this case the source BNF code), I believe we should retain this functionality in case of a user needing to update the codelist with this column intact. In time we may wish to review this decision.

image

Original converted version (old-style) with the BNF column
image

New draft new-style version in the builder, where the user is redirected to after clicking the "create new version" button
image

### DIFF
--- a/codelists/views/version.py
+++ b/codelists/views/version.py
@@ -85,6 +85,13 @@ def version(request, clv):
     else:
         coding_system_release_outdated = False
 
+    is_converted_from_bnf = (
+        clv.codelist.coding_system_id == "dmd"
+        and clv.codelist.slug.endswith("-dmd")
+        and clv.codelist.methodology is not None
+        and "[pseudo-BNF system codelist]" in clv.codelist.methodology
+    )
+
     ctx = {
         "clv": clv,
         "codelist": clv.codelist,
@@ -100,5 +107,6 @@ def version(request, clv):
         "latest_published_version_url": latest_published_version_url,
         "count_codes_included": len(rows),
         "coding_system_release_outdated": coding_system_release_outdated,
+        "is_converted_from_bnf": is_converted_from_bnf,
     }
     return render(request, "codelists/version.html", ctx)
--- a/codelists/actions.py
+++ b/codelists/actions.py
@@ -596,7 +596,7 @@ def delete_version(*, version):
 
 
 @transaction.atomic
-def convert_codelist_to_new_style(*, codelist):
+def convert_codelist_to_new_style(*, codelist, status=Status.UNDER_REVIEW):
     """Convert codelist to new style.
     Create a new version with the same codes as the latest version.
     """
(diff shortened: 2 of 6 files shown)


codelists/views/version.py
```diff
@@ -85,6 +85,13 @@ def version(request, clv):
     else:
         coding_system_release_outdated = False
 
+    is_converted_from_bnf = (
```
Jongmassey, line 88: This arguably could be implemented as a model property, but would only be used here

Jongmassey: Also, "is this codelist a BNF to dm+d converted one?" is something I've had to derive multiple times when doing bits of analysis, it'd be nice if it were a property.

Nonetheless, this is still flaky (can be broken if a user updates the metadata to remove the elements we rely on),and we should consider adding an actual flag to the model so we're not reliant on this in future.

Issue #3147, opened 2026-07-13
Don't show API auth tokens in User ListView in Django Admin area


lucyb:
# Why are we doing this?

The list of users in the Django Admin area should only show the information that's relevant at that point. Sensitive information, such as auth tokens should only be visible when clicking on a users name from the list view.

Timebox: 1-2 hours

# How will we know when it's done?

Sensitive information, such as auth tokens, are not visible in the User ListView

---

[Defining delivery tasks guidance](https://bennett.wiki/tech-group/delivery-process/task-definition/)

PR #3161, opened 2026-07-20
Update the Django Admin token interface

tomodwyer:
This PR creates a custom admin screen for the auth tokens. That means we can make the following changes:

## List view

- Show tokens as a list of username and created at date
- Format the created at date to show in UTC
- Do not show the auth token here

## Token view

- Show the user, key, and formatted created at date
- Do not allow admins to make any changes to the key, only deletion (for rotation)

## Add token

- Allow admins to select the user
- Hide the key and created at fields from being manually set

## Screenshots

Screenshot of the token interface in Django Admin

### DIFF
--- a/opencodelists/admin.py
+++ b/opencodelists/admin.py
@@ -1,4 +1,11 @@
+from datetime import UTC
+
 from django.contrib import admin
+from django.http import HttpRequest
+from django.utils import timezone
+from django.utils.formats import date_format
+from rest_framework.authtoken.admin import TokenAdmin as DRFTokenAdmin
+from rest_framework.authtoken.models import TokenProxy
 
 from .models import Membership, Organisation, User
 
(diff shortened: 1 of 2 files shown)


mikerkelly:
Sorry, Thomas, I don't think that this issue was quite ready for review. I cannot tell from the issue, this PR, or your commit messages why you have done what you have done or why you have done it that way.

Requirements
============

The only requirement I can find in the parent issue (assuming it is #3147) is "Sensitive information, such as auth tokens, are not visible in the User ListView".

Some reference to that requirement should be in the main commit message for this change. For all but the really obvious changes, please include some discussion of why you have made the changes that you have, in the way that you have. This helps you get your thoughts in order and  aids the reviewer and anyone looking at the commit in future to easily und the motivation and rationale for the change. Please read and more-or-less follow https://cbea.ms/git-commit/ when making git commit messages. Some of that rationale can often be copied to the PR description, too.

You also seem to be handling additional requirements: displaying the timestamps in UTC, and not allowing the user to edit the user a token is assigned to.  Could you explain where these have come from, please? It's okay to add things you think will be desireable, but they do need explanation.

The timestamps already seem to have been in UTC, also.

How to change the behaviour
===========================

I am not familiar with DRF-specific admin stuff but it seems a direct outgrowth of the main Django admin. This prefers to define and customize admin views as much as possible with declarative class attributes, using custom function only where really needed. I think we should try to work with the grain of the framework as far as possible. I don't think that is what your current solution is doing.

If I look at the DRF documentation I find this on customising the token admin: https://www.django-rest-framework.org/api-guide/authentication/#with-django-admin

> It is also possible to create `Tokens` manually through the admin interface. In case you are using a large user base, we recommend that you monkey patch the `TokenAdmin` class to customize it to your needs

```python
from rest_framework.authtoken.admin import TokenAdmin

TokenAdmin.raw_id_fields = ['user']
```

Following this approach, I tried the following:

```
from rest_framework.authtoken.admin import TokenAdmin

TokenAdmin.list_display = ['user', 'created']
TokenAdmin.fields = ['user', 'created' ]
TokenAdmin.readonly_fields = ['created']
```

That seems to immediately do most of what you want. It is a lot less code to maintain and completely standard Django admin stuff, so not sure it even needs automated tests.

List
image

Edit 
image


Create 
image

Post create 

image

If you want to disable reassigning a token, get_readonly_fields can be more custom:

```
def get_readonly_fields(self, request, obj=None):
    if obj is None:  # creating
        return ['created']
    return ['user', 'created']  # editing
TokenAdmin.get_readonly_fields = get_readonly_fields
```

That probably would deserve some tests.

PR #3163, opened 2026-07-21
Update Auth Token admin views

### LINKED ISSUE #3147: Don't show API auth tokens in User ListView in Django Admin area

tomodwyer:
Removes API token values from the Django admin to prevent accidental exposure. The list view now displays only the associated user and creation date, while the detail view allows only the user to be changed.

Closes #3147

### DIFF
--- a/opencodelists/admin.py
+++ b/opencodelists/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from rest_framework.authtoken.admin import TokenAdmin
 
 from .models import Membership, Organisation, User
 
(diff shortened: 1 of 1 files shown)


mikerkelly (approved):
Looks good to me.

PR #3179, opened 2026-07-31
Refactor known_diffs.py into module

Jongmassey:
These commits refactor the large, all-encompassing known_diffs.py into a module.

Each sort of difference (as defined by having its own const dict) is moved into its own file and has its helper functions with it.

Where const dicts were accessed outside of the module these are moved to a helper function for consistency.

The other major change is to tests which test functions that make use of the known difference data in some way, via a pre-existing helper function. These have been changed to monkeypatch the helper function rather than the underlying data. There were existing unit tests for the helper functions which patched the underlying data within test_known_diffs.py. 
This clarifies the responsibilities of these tests - the known diff helper function unit tests test that the helper function behaves correctly, the unit tests for the things that use the output of these helper functions (e.g. views) test that these call the helper function correctly and consume their outputs correctly - i.e. that the code within the function under test is correct, not testing its dependencies which are tested elsewhere.

Overall this improves readability, clarifies responsibilities, and hides internal implementation details from the rest of the codebase.

There remains some coupling between release_builder.py  (and its tests) and the `RubricDifference` dataclass. This was too tricky to unpick at this moment.


Jongmassey:
Absorbed into #3194

PR #3184, opened 2026-08-03
Add "Coding System Releases" link to footer

### LINKED ISSUE #3141: Link to the Coding System Releases page
# Why are we doing this?

OpenCodelists contains a useful [Coding System Releases](https://www.opencodelists.org/coding-systems/latest-releases) page, but it is not linked in any navigation. We should link to it so that users can see when coding systems were last updated

...

# What are we doing?

...

---

tomodwyer:
Including restyling the page to Tailwind styles, and improving the footer layout for the Tailwind base template.

Closes https://github.com/opensafely-core/opencodelists/issues/3141

# Screenshots


mikerkelly (approved):
Thanks, Thomas. These all seem like clear improvements.

PR #3194, opened 2026-08-07
UI changes for ICD-10 multi-edition releases

rw251:
The feature branch for all the UI changes required to support the new ICD-10 multi-edition release.

This includes:
- Warning banners for the two situations where we want people to fix their codelists:
  - A codelist that contains a code from one edition, but not the equivalent code in the other edition
  - A codelist that contains a code whose definition has changed meaningfully between editions
- Extra ICD-10 information in the "more info" pop up including:
  - Extra "rubrics" that give more information about the code
  - Dagger/asterisk information (a specific ICD-10 relationship)
  - Definition differences i.e. for codes with a changed description, we explain why and display both
- Extending the search so that we make use of "inclusion rubrics" as synonyms

This branch should be merged at the same time that the new ICD-10 coding release is minted.


userdocs/icd10-editions.md
```diff
@@ -0,0 +1,142 @@
...
+- clinically significant differences.
+
+The former are not a concern, e.g.
+["Cow milk"](https://icd.who.int/browse10/2019/en#/P71.0) vs
+["Cow's milk"](https://classbrowser.nhs.uk/#/book/ICD-10-5TH-Edition/vol1/block-p70-p74.htm+P71.0).
+However, the latter can lead to a codelist including events incorrectly. For
+example, the code `X59.0` will match fractures in ONS deaths because the 2019 edition defines it as
+["Exposure to unspecified factor _**causing fracture**_"](https://icd.who.int/browse10/2019/en#/X59.0)
+but that same code in admissions data is defined in the 2016 edition as
+["Exposure to unspecified factor (occurrence at home)"](https://classbrowser.nhs.uk/#/book/ICD-10-5TH-Edition/vol1/block-x50-x59.htm+X59.0).
```
jspickering: this link seems to be broken

userdocs/icd10-editions.md
```diff
@@ -0,0 +1,142 @@
...
+There are ~20 codes that have moved between the two editions. For example, the
+concept of "Pneumocystosis" is coded as
+[B59 in the 2016 edition](https://classbrowser.nhs.uk/#/book/ICD-10-5TH-Edition/vol1/block-b50-b64.htm+B59)
+, but
+[B48.5 in the 2019 edition](https://icd.who.int/browse10/2019/en#/B48.5).
+Fortunately, there are no instances where a code has been reused for a different
+concept in the other edition. This means that it is safe to include the code
+from both editions in a codelist, irrespective of which edition is used in the
+dataset being queried. A codelist with both `B59` and `B48.5` will match events
+in ONS deaths with `B48.5` and events in admissions data with `B59`.
```
jspickering, line 44: I’m possibly being pedantic, but should we refer to the datasets by their proper name each time rather than the short hands such as “deaths data” or “admissions data”? To aid search, visual scanning of the page etc. It might not be necessary though

userdocs/icd10-editions.md
jspickering, line 1: Is it worth adding information on this page about when the combined edition was added, which ICD-10 edition codelists were built with before that date, and how to update those to now use the combined edition? I know it’s going in other forms of communication, but there might be users in the future seeking to use older ICD-10 codelists who may want to know this information, potentially.

Will the below statements stand the test of time? I wonder if it’s worth avoiding this sort of specificity so that it doesn’t get out of date if we forget to check it in future:
- “There are ~20 codes that have moved…”
- “There are no instances where a code has been…”
- “…there are 5 other codes…”

I wonder if all a user needs to know is that these instances are few, so presumably for most users it will never be relevant, but if it does become relevant then they will see further information in the UI itself anyway?

userdocs/icd10-editions.md
Jongmassey: Huh. It should be https://classbrowser.nhs.uk/#/book/ICD-10-5TH-Edition/vol1/block-x58-x59.htm+X59

Is this a sign of the NHS ICD-10 browser being updated?!

Jongmassey:
```py
>>> c=Concept.objects.using(dba).get(code="X59").parent_id
'X58-X59'
```

The combined copy of ICD-10 I've got loaded locally has parent block being `X58-X59` (as per the current link) rather than `X50-X59` (per the old broken link). So if it has changed it was between whenever this doc was written and when I last pulled data. I've got an old scraped copy in the original browser scraper that has it as `X58-X59` as well so not quite sure what that broken URL has come from to be honest.

rw251: That is weird because I can't imagine I typed that url by hand. All the WHO ones are also split into `X50-X57` and `X58-X59` so I think most likely is the issue is with me. In any case, I'll correct it

userdocs/icd10-editions.md
rw251: Re the small numbers that might date. I'm almost certain that this won't change, or if it does only by 1 or 2, and I think there is a positive to having the actual numbers. Saying "20 codes" is nice and specific - but definitely a small number - but any other way of describing it "a few", "hardly any", "small number" etc is open to interpretation and might sound like we're trying to downplay the impact. I'm also happy with the small risk that this dates so will leave it as is.

Re the first comment on the extra information about the dates, versions, and instructions. This is a very good idea. I'll have a think of how to best represent this in a brief way.

userdocs/icd10-editions.md
rw251: Easy enough to change the deaths one so that every mention is "ONS deaths". For admissions I think it's fine to call it "HES/APCS admissions data" at the top, then just "admissions" after that - the full title is a bit clunky, and I think as long as we remain consistent it's ok.

userdocs/icd10-editions.md
rw251: I'll add something saying that originally OCL only supported WHO 2019 and this changed to the combined edition in Sep 2026. But after a bit more thought I'm not sure instructions on updating old codelists fits well in the docs. Anyone using a codelist with warnings, will already be guided to sections with instructions.

userdocs/icd10-editions.md
rw251: Link fixed

Jongmassey (approved):
Looks good to me! 🚀

PR #3205, opened 2026-08-12
Add cwd and envFile to OCL debug launcher

rw251:
The "Launch OpenCodelists" debug script worked locally for me, but not for all users as the `.env` file failed to be located due to environment differences. Adding the `cwd` and `envFile` properties explicitly, fixes this.


KatieB5 (approved):
Excellent, thanks for this Richard!

Issue #2532, opened 2025-05-13
Fix the broken state possible when deleting search terms


StevenMaude:
# Why are we doing this?

The builder can get into a broken state for the user with a blank page shown and no further way to interact with the builder.

This persists even if you navigate away from the blank page.

As a user, I can't see an obvious (or any?) way of recovering my lost work, or even deleting the broken codelist from my account. 

# How to reproduce

(This should work with the test fixtures too as this is where I observed it first.)

## Video

https://github.com/user-attachments/assets/042d09fd-14f4-4512-b5d6-ad64ff8bc7bd

## Instructions

1. Create a new SNOMED CT codelist.
1. Search for the term "arthritis".
1. Select all under "Disorder > Arthritis (`3723001`)" by clicking `+` next to it.
1. Search for the term "elbow".
1. Delete the "arthritis" search term.
1. Click on "no search term" under searches.
1. Deselect "Arthritis (`3723001`)" by clicking `+` next to it.
1. Delete the search "elbow".
1. Get a blank screen with a notification saying that the search has been deleted.
1. Return to "My codelists".
1. Select the draft codelist you created.
1. Still see the blank screen.

## More details

It's not SNOMED CT-specific behaviour.

For example, you can reproduce with ICD10:

1. Search for "whooping".
1. Select "Whooping cough (`A37`)" by clicking `+` next to it.
1. Search for "Bordetella".
1. Delete the "whooping" search.
1. Click "no search term".
1. Deselect "Whooping cough (`A37`)" by clicking `+` next to it.

# How will we know when it's done?

When the codelist builder gets into an at-least interactive state in this case. What that state might be probably needs defining.

# What are we doing?

Hopefully addressing the bug or the behaviour that allows getting into this state.

tomodwyer:
[This if statement is the issue](https://github.com/opensafely-core/opencodelists/blob/4280bbeee09a4844e9129135a7c7890cd792ebfb/assets/src/scripts/_hierarchy.ts#L303-L316)

adding a question mark ([optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)) here will make it "work", as in it will not crash the app, but it does not solve the underlying issue.

```js
codeToStatus[d]?.includes(codeToStatus[code])
```

tomodwyer:
I am going to split this in to two PRs, as the issue that was originally believed to only be in the React app, actually has a Python component.

[...]

PR #3209, opened 2026-08-19
Allow re-inclusion of previously-deleted orphan codes

### LINKED ISSUE #3206: Bug report: all codes removed from codelists created via CSV upload
# Summary
When updating a codelist created via CSV upload, if I remove the top-level parent code then include a descendant code and save the draft, the returned draft codelist has no codes.

### Include relevant details of specific URLs accessed, actions taken, and data entered.

- I downloaded a CSV for a codelist: https://www.opencodelists.org/codelist/bristol/anticoagulants-bnf/1099a4d8/
- I went to: https://www.opencodelists.org/users/<my-username>/new-codelist/
- I filled out the create a codelist form to create a Pseudo BNF codelist:
 Image
 Image

- I got the following builder draft page:
Image

- Note that this codelist has no previous searches, and the top-level parent code is explicitly included
- If I then modify the draft codelist to explicitly include a descendent code, and **after** doing that I explicitly exclude the top-level parent code:
Image

- And click "Save draft",  I get a codelist draft as follows:
Image

- And if I then click "Save for review", i get:
Image

- However, if I follow the same process and explicitly exclude the top-level parent code **before** I explicitly include a descendent code, I get the following codelist with no codes:
Image

### Can you reliably trigger the bug? What steps would you follow?
- Yes, follow the steps above

### What did you observe instead of what you expected?
- I expected to see the same codelist created after I clicked "Saved draft" regardless of the order in which I explicitly included and excluded parent and descendant codes.

### Does the behaviour happen only sometimes or every time?
- Every time

### Did the bug start happening recently?
- The code that I suspect may be contributing to this issue (`def update_code_statuses(*, draft, updates)` in [builder/actions.py](https://github.com/opensafely-core/opencodelists/blob/main/builder/actions.py)) was introduced about a year ago, so this behaviour may have been occurring for some time.

#### How does this bug affect you? 
- It removes all excluded codes from my codelist, if those codes do not have an explicitly-included parent code; this is not ideal.

Jongmassey:
Fixes #3206 

Orphan codes are codes without related searches. This may be because the search whence they came has been deleted or because the codelist was originally created by upload of a csv file.

Orphan codes that are not either themselves included or descendants of an included code are deleted to avoid a broken state (see #2639 ). However, the UI is not immediately synchronised with the db state after this deletion and so they persist in the UI.
A user may then request inclusion of these deleted-in-the-db-but-persisting-in-the-ui codes which would then show them as included in the UI but they would still be missing from the db. Upon saving the codelist, these codes and their descendants would disappear from view, in some cases leaving completely empty codelists (see #3206 ).

This PR ensures that any request for inclusion of codes is honoured in the db such that what the user sees in the builder view is what they will get when they click save.

### DIFF
--- a/builder/actions.py
+++ b/builder/actions.py
@@ -44,7 +44,7 @@ def get_codes_to_keep(codeset_version, potential_codes):
     1. Explicitly included in the codelist
     2. Descendants of included codes
     """
-    # Get all explicitly included codes on the codelist
+    # Get all explicitly and implicitly included codes on the codelist
     all_included_codes = codeset_version.codeset.codes()
 
     # Build a list of codes_to_keep from this search, consisting of any included codes or their descendants
@@ -158,8 +158,22 @@ def update_code_statuses(*, draft, updates):
     new_codeset = draft.codeset.update(updates)
 
     status_to_new_code = defaultdict(list)
+    previous_codes = draft.codeset.all_codes()
     for code, status in new_codeset.code_to_status.items():
         status_to_new_code[status].append(code)
+        # check for re-inclusion of previously-deleted orphan codes
+        if code not in previous_codes and status == "+":
+            codes_to_reinclude = (
+                {code} | draft.hierarchy.descendants(code)
+            ) - previous_codes
+            CodeObj.objects.bulk_create(
+                CodeObj(
+                    version=draft,
+                    code=code_to_reinclude,
+                    status=status if code_to_reinclude == code else "(+)",
+                )
+                for code_to_reinclude in codes_to_reinclude
+            )
 
     for status, codes in status_to_new_code.items():
         draft.code_objs.filter(code__in=codes).update(status=status)
--- a/builder/tests/test_actions.py
+++ b/builder/tests/test_actions.py
@@ -527,3 +527,71 @@ def test_cannot_save_with_incomplete_code_status(
 
     draft.refresh_from_db()
     assert draft.is_draft
+
+
+def test_re_inclusion_of_previously_deleted_orphan(draft_with_some_searches):
+    draft = draft_with_some_searches
+
+    # Double check that codes and statuses are as expected
+    assert dict(draft.code_objs.values_list("code", "status")) == {
+        # Part of arthritis search
+        "439656005": "+",  # Arthritis of elbow
+        "202855006": "(+)",  # Lateral epicondylitis
+        "3723001": "-",  # Arthritis
+        # Orphans
+        "128133004": "+",  # Disorder of elbow
+        "429554009": "(+)",  # Arthropathy of elbow
+        "35185008": "(+)",  # Enthesopathy of elbow region
+        "73583000": "(+)",  # Epicondylitis
+        "239964003": "(+)",  # Soft tissue lesion of elbow region
+        "156659008": "+",  # (Epicondylitis &/or ...
+    }
+
+    # Deselect orphaned disorder of elbow
+    actions.update_code_statuses(draft=draft, updates=[("128133004", "?")])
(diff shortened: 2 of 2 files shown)


builder/actions.py
```diff
@@ -158,8 +158,22 @@ def update_code_statuses(*, draft, updates):
     new_codeset = draft.codeset.update(updates)
 
     status_to_new_code = defaultdict(list)
+    previous_codes = draft.codeset.all_codes()
     for code, status in new_codeset.code_to_status.items():
         status_to_new_code[status].append(code)
+        # check for re-inclusion of previously-deleted orphan codes
+        if code not in previous_codes and status == "+":
```
KatieB5, line 165: Should we also be retaining codes with a status of "?" ?

KatieB5:
Thanks very much for this Jon! I've pulled down your branch and run it locally, and I'm happy that it fixes the original bug I flagged. I just have one question around whether we should also retain codes with a status of "?". 

I think maybe we should: When I go to make a draft codelist in the builder, if I have any 'unresolved'/'?' codes without a related search and I click "Save draft", these codes then disappear from my codelist. I think it's plausible that a user might do this if they're undecided about a particular code perhaps, and hit "Save draft" (not knowing this button doesn't actually do anything other than redirect back to this page) whilst they think more about their decision. They may be surprised, or not anticipate this behaviour (code removal), and because in this particular case there were no searches originally done to build the codelist, there'd be a small amount of work needed to add the unresolved code back in for consideration.

#### Example with an uploaded csv-based BNF codelist

1. I save my draft after including/excluding some codes and I'm left with this:
Screenshot from 2026-08-25 11-44-38

2. I create an unresolved code (Eliquis):
Screenshot from 2026-08-25 11-44-58


3. I click "Save draft", and the unresolved code is gone:
Screenshot from 2026-08-25 11-45-11

Jongmassey:
>I think maybe we should: When I go to make a draft codelist in the builder, if I have any 'unresolved'/'?' codes without a related search and I click "Save draft", these codes then disappear from my codelist.

This is a very pertinent question and is basically what I asked in the "Should we be deleting "orphan" codes after they or one of their parents are excluded/set to "unresolved?"" bit of my comment [here](https://github.com/opensafely-core/opencodelists/issues/3206#issuecomment-5393579507).

I could see it both ways (hence the YES/NO bits in aforementioned comment!).
If I'm building a codelist using searches then there are certainly circumstances in which I would reasonably expect them to be removed. For example, if I've added a search term, included some but not all terms, deleted the search to get rid of all the ones I didn't want, changed my mind about one of the ones I included and then set it to unresolved - I'd expect it to be removed as if it had been unresolved before I deleted the search. 
If I've uploaded a codelist then almost certainly not.
If I've started with an uploaded codelist then added some searches to it to modify it then... it depends?

Unfortunately there's no good way of robustly telling if a codelist in a new-style (i.e. builder-enabled) coding system originated as an uploaded codelist. It's indistinguishable from a codelist that was created using searches then had all those searches deleted. Equally, a codelist with some codes associated with search results and some not could equally be created with searches and then some of those deleted, or started from an upload then had those searches added.
Therefore, we can't easily have diverging behaviour for uploaded and search-based codelists.

It is very likely that #2639  was the wrong way of fixing #2532 and a more comprehensive solution would revisit the original bug. From my initial look at that it feels like a pretty big bit of work which would require some product involvement to properly answer the question both you and I have landed on.

This PR is only a partial fix, but felt like the correct tactical approach which does address the specific issue you and I both encountered.

KatieB5 (approved):

builder/actions.py
Jongmassey: It's debatable either way. One thing is for sure is that we absolute definitely should be re-including things that are included.

PR #3214, opened 2026-08-26
Add code comments to explain `Save draft` button behaviour for draft codelists

KatieB5:
The behaviour of the `Save draft` button isn't immediately obvious. Despite its name, it doesn't save the draft codelist; changes are saved as code statuses are updated.

This PR adds comments to clarify the current behaviour and make it easier to understand in future.

### DIFF
--- a/builder/views.py
+++ b/builder/views.py
@@ -64,6 +64,11 @@ def _handle_post(request, draft):
         )
         return redirect(draft)
     elif action == "save-draft":
+        # Note: the "Save draft" button does not save a draft
+        # codelist, we only redirect the user back to the draft codelist.
+        # We do this to avoid users wondering how to save their work.
+        # Updates to draft codelists are saved to the backend when a
+        # user changes the status of a code in the builder, via update_code_statuses().
         messages.add_message(request, messages.INFO, "Your changes have been saved")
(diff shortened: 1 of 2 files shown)


builder/views.py
```diff
@@ -64,6 +64,10 @@ def _handle_post(request, draft):
         )
         return redirect(draft)
     elif action == "save-draft":
+        # Note: the "Save draft" button does not save a draft
+        # codelist, we only redirect the user back to the draft codelist.
+        # Updates to draft codelists are saved to the backend when a
+        # user changes the status of a code in the builder, via update_code_statuses().
```
mikerkelly, line 71: It it worth adding one line here and in the commit message about the rationale? "We do this to avoid users wondering how to save their work" or similar? It may not be obvious why we have a """do nothing""" button.

mikerkelly (approved):
Thanks Katie, this is really helpful to clarify the behaviour. I like that you have put a short comment in the Typescript referencing a longer one in the view code.

One tiny suggestion if you like it.

builder/views.py
KatieB5: Thanks Mike! Yes, I think it's a good idea to include the 'why', too. I'll update to add this in.

PR #3223, opened 2026-09-02
Add report generation/emailing to ICD-10 feature branch

Jongmassey:


Jongmassey:
@rw251 Do you want to do any further fettling/squishing etc. of commits here before merged into the feature branch?

scripts/icd10_reports/email.py
```diff
@@ -0,0 +1,120 @@
...
+        response = requests.post(
+            _mailgun_url("messages"),
+            auth=("api", settings.ANYMAIL["MAILGUN_API_KEY"]),
+            files=files,
+            data=data,
+        )
+        # 429 codes are either rate exceeded (2 types), or too large
+        # https://documentation.mailgun.com/en/latest/api-sending.html#rate-limits
+        # with message "<why> limit exceeded, try again" where why=[bytes|request|recipient]
+        if response.status_code == "429":
```
rw251, line 60: `response.status_code` is an integer so I think this conditional is never true - but also I think the mailgun response has changed (https://documentation.mailgun.com/docs/mailgun/api-reference/api-overview#api-response-codes) and not sure it returns a "why" message. Might be easiest to just remove this

scripts/icd10_reports/email.py
```diff
@@ -0,0 +1,120 @@
...
+    }
+    for email, name, pdf_filename in recipients:
+        if dry_run:
+            print(
+                f"Would send email to {name} at {email} with attachment {pdf_filename}"
+            )
+            continue
+        try:
+            if only_seeds and email.lower() not in seed_emails:
+                continue
```
rw251: nit - but currently in dry run it prints "would send.." for everyone even if a seeds list is given. We could either move this `if only_seeds` higher up, or add a print like "Actually not in seeds so wouldn't send"

scripts/generate_icd10_warning_pdfs.py
```diff
@@ -0,0 +1,121 @@
...
+
+        user_count = sum(owner.kind == "user" for owner in reports)
+        organisation_count = sum(owner.kind == "organisation" for owner in reports)
+        print(f"Found {len(affected)} affected codelists.")
+        print(
+            f"Wrote {user_count} user PDFs, {organisation_count} organisation PDFs, "
+            f"recipients.csv, summary.md, issues.json, and example.pdf."
+        )
+    if args.send_emails:
+        send_emails(
```
rw251: I ran this locally and it didn't print out the dry run messages - I think the generator isn't being consumed?

Jongmassey: Correct, total brain fart on my part!

scripts/icd10_reports/email.py
Jongmassey: Oh, it has. I'll change accordingly thanks

PR #3224, opened 2026-09-04
Add standalone ICD-10 project report generator

rw251:
As part of the release of the multi-edition ICD-10 coding system we notified people who have projects that have used a codelist that we know to be affected by the change. E.g. a codelist with a code that has moved between editions, or where its definition has changed meaningfully. The code in this commit is to support a one-off script for generating PDF reports for affected project owners. This is kept as a record of what was done, but is unlikely to be needed again.


Jongmassey (approved):
Could do with a bit of info in the commit message and the PR itself but otherwise good to go

